Issue #2, opened 2023-03-07
Immediate allocation is missing


byako:
Implement example of immediate allocation

asm582:
it would be great if we could add Immediate mode to the example driver, thanks

byako:
Unfortunately, Immediate allocation mode is going away in 1.31, it's already gone from k/k master. Closing.

PR #4, opened 2023-04-10
Immediate allocation

byako:
Implements #2

### DIFF
--- a/api/example.com/resource/gpu/nas/v1alpha1/client/client.go
+++ b/api/example.com/resource/gpu/nas/v1alpha1/client/client.go
@@ -99,3 +99,18 @@ func (c *Client) Get() error {
 	*c.nas = *crd
 	return nil
 }
+
+func (c *Client) ListNames() ([]string, error) {
+	naslist, err := c.client.NodeAllocationStates(c.nas.Namespace).List(
+		context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	nasnames := []string{}
+	for _, nas := range naslist.Items {
+		nasnames = append(nasnames, nas.Name)
+	}
+
+	return nasnames, nil
+}
--- a/api/example.com/resource/gpu/nas/v1alpha1/client/client.go
+++ b/api/example.com/resource/gpu/nas/v1alpha1/client/client.go
@@ -100,17 +100,20 @@ func (c *Client) Get() error {
 	return nil
 }
 
-func (c *Client) ListNames() ([]string, error) {
-	naslist, err := c.client.NodeAllocationStates(c.nas.Namespace).List(
-		context.TODO(), metav1.ListOptions{})
+func (c *Client) Switch(newcrd *nascrd.NodeAllocationState) {
+	*c.nas = *newcrd
+}
+
+func (c *Client) GetAll() (map[string]*nascrd.NodeAllocationState, error) {
+	naslist, err := c.client.NodeAllocationStates(c.nas.Namespace).List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
-	nasnames := []string{}
+	nasmap := map[string]*nascrd.NodeAllocationState{}
 	for _, nas := range naslist.Items {
-		nasnames = append(nasnames, nas.Name)
+		nasmap[nas.Name] = &nas
 	}
 
-	return nasnames, nil
+	return nasmap, nil
 }
--- a/cmd/dra-example-controller/driver.go
+++ b/cmd/dra-example-controller/driver.go
@@ -93,10 +93,24 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 }
 
 func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
+
 	if selectedNode == "" {
-		return nil, fmt.Errorf("TODO: immediate allocations not yet supported")
+		allocationResult, err := d.allocateImmediateClaim(claim, claimParameters, class, classParameters)
+		if err != nil {
+			return nil, fmt.Errorf("error performing immediate allocation: %v", err)
+		}
+		return allocationResult, nil
 	}
 
+	allocationResult, err := d.allocateDelayedClaim(claim, claimParameters, class, classParameters, selectedNode)
+	if err != nil {
+		return nil, fmt.Errorf("error performing delayed allocation: %v", err)
+	}
+
+	return allocationResult, nil
+}
+
+func (d driver) allocateDelayedClaim(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
 	d.lock.Get(selectedNode).Lock()
 	defer d.lock.Get(selectedNode).Unlock()
 
@@ -128,7 +142,7 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
 	classParams := classParameters.(*gpucrd.DeviceClassParametersSpec)
 	switch claimParams := claimParameters.(type) {
 	case *gpucrd.GpuClaimParametersSpec:
-		onSuccess, err = d.gpu.Allocate(crd, claim, claimParams, class, classParams, selectedNode)
+		onSuccess, err = d.gpu.AllocateDelayed(crd, claim, claimParams, class, classParams, selectedNode)
 	default:
 		err = fmt.Errorf("unknown ResourceClaim.ParametersRef.Kind: %v", claim.Spec.ParametersRef.Kind)
 	}
@@ -144,6 +158,70 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
 	onSuccess()
 
 	return buildAllocationResult(selectedNode, true), nil
+}
+
+func (d driver) allocateImmediateClaim(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}) (*resourcev1.AllocationResult, error) {
+	crdconfig := &nascrd.NodeAllocationStateConfig{
+		Namespace: d.namespace,
+	}
+	crd := nascrd.NewNodeAllocationState(crdconfig)
+
+	client := nasclient.New(crd, d.clientset.NasV1alpha1())
+
+	nasnames, err := client.ListNames()
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving list of NodeAllocationState objects: %v", err)
+	}
+
+	for _, nodename := range nasnames {
+		allocationResult, err := d.allocateImmediateClaimOnNode(claim, claimParameters, class, classParameters, nodename)
+		if err == nil {
+			return allocationResult, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no suitable node found")
+}
+
+func (d driver) allocateImmediateClaimOnNode(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, nodename string) (*resourcev1.AllocationResult, error) {
+	d.lock.Get(nodename).Lock()
+	defer d.lock.Get(nodename).Unlock()
+
+	crdconfig := &nascrd.NodeAllocationStateConfig{
+		Namespace: d.namespace,
+		Name:      nodename,
+	}
+
+	crd := nascrd.NewNodeAllocationState(crdconfig)
+
+	client := nasclient.New(crd, d.clientset.NasV1alpha1())
+	err := client.Get()
+	if err != nil {
+		return nil, fmt.Errorf("retrieving NodeAllocationState CRD for node %v: %v", nodename, err)
+	}
+
+	if crd.Spec.AllocatedClaims == nil {
+		crd.Spec.AllocatedClaims = make(map[string]nascrd.AllocatedDevices)
+	}
+
+	classParams := classParameters.(*gpucrd.DeviceClassParametersSpec)
+	switch claimParams := claimParameters.(type) {
+	case *gpucrd.GpuClaimParametersSpec:
+		err = d.gpu.AllocateImmediately(crd, claim, claimParams, class, classParams)
+		if err != nil {
+			return nil, fmt.Errorf("could not find suitable devices: %v", err)
+		}
+	default:
+		return nil, fmt.Errorf("unknown ResourceClaim.ParametersRef.Kind: %v", claim.Spec.ParametersRef.Kind)
+	}
+
+	// sync ResourceClaimAllocations from allocations
+	err = client.Update(&crd.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("updating NodeAllocationState CRD: %v", err)
+	}
+
+	return buildAllocationResult(nodename, true), nil
 }
 
 func (d driver) Deallocate(ctx context.Context, claim *resourcev1.ResourceClaim) error {
--- a/cmd/dra-example-controller/driver.go
+++ b/cmd/dra-example-controller/driver.go
@@ -93,7 +93,6 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 }
 
 func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
-
 	if selectedNode == "" {
 		allocationResult, err := d.allocateImmediateClaim(claim, claimParameters, class, classParameters)
 		if err != nil {
@@ -167,15 +166,21 @@ func (d driver) allocateImmediateClaim(claim *resourcev1.ResourceClaim, claimPar
 	crd := nascrd.NewNodeAllocationState(crdconfig)
 
 	client := nasclient.New(crd, d.clientset.NasV1alpha1())
-
-	nasnames, err := client.ListNames()
+	nasmap, err := client.GetAll()
 	if err != nil {
 		return nil, fmt.Errorf("error retrieving list of NodeAllocationState objects: %v", err)
 	}
 
-	for _, nodename := range nasnames {
-		allocationResult, err := d.allocateImmediateClaimOnNode(claim, claimParameters, class, classParameters, nodename)
+	for nodename, nas := range nasmap {
+		allocationResult, err := d.allocateImmediateClaimOnNode(claim, claimParameters, class, classParameters, nodename, nas)
 		if err == nil {
+			// sync ResourceClaimAllocations from allocations
+			client.Switch(nas)
+			err := client.Update(&nas.Spec)
+			if err != nil {
+				return nil, fmt.Errorf("updating NodeAllocationState CRD: %v", err)
+			}
+
 			return allocationResult, nil
 		}
 	}
@@ -183,22 +188,9 @@ func (d driver) allocateImmediateClaim(claim *resourcev1.ResourceClaim, claimPar
 	return nil, fmt.Errorf("no suitable node found")
 }
 
-func (d driver) allocateImmediateClaimOnNode(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, nodename string) (*resourcev1.AllocationResult, error) {
+func (d driver) allocateImmediateClaimOnNode(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, nodename string, crd *nascrd.NodeAllocationState) (*resourcev1.AllocationResult, error) {
 	d.lock.Get(nodename).Lock()
 	defer d.lock.Get(nodename).Unlock()
-
-	crdconfig := &nascrd.NodeAllocationStateConfig{
-		Namespace: d.namespace,
-		Name:      nodename,
-	}
-
-	crd := nascrd.NewNodeAllocationState(crdconfig)
-
-	client := nasclient.New(crd, d.clientset.NasV1alpha1())
-	err := client.Get()
-	if err != nil {
-		return nil, fmt.Errorf("retrieving NodeAllocationState CRD for node %v: %v", nodename, err)
-	}
 
 	if crd.Spec.AllocatedClaims == nil {
 		crd.Spec.AllocatedClaims = make(map[string]nascrd.AllocatedDevices)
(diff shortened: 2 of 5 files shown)


api/example.com/resource/gpu/nas/v1alpha1/client/client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) Get() error {
 	*c.nas = *crd
 	return nil
 }
+
+func (c *Client) ListNames() ([]string, error) {
+	nasnames := []string{}
+	nases, err := c.client.NodeAllocationStates(c.nas.Namespace).List(
+		context.TODO(), metav1.ListOptions{})
```
klueska, line 106: Either leave this all on one line or put all arguments on a new line. Also, `nases` is a weird name, let's just call it `list` or `naslist`, e.g.:
```
	naslist, err := c.client.NodeAllocationStates(c.nas.Namespace).List(
		context.TODO(),
		metav1.ListOptions{})
```

api/example.com/resource/gpu/nas/v1alpha1/client/client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) Get() error {
...
+	if err != nil {
+		return nasnames, err
```
klueska, line 108:
```suggestion
		return nil, err
```

api/example.com/resource/gpu/nas/v1alpha1/client/client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) Get() error {
...
+	}
+	for _, nas := range nases.Items {
```
klueska, line 110:
```suggestion
	for _, nas := range naslist.Items {
```

cmd/dra-example-controller/driver.go
```diff
@@ -70,7 +70,11 @@ func (d driver) GetClassParameters(ctx context.Context, class *resourcev1.Resour
 	return &dc.Spec, nil
 }
 
-func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.ResourceClaim, class *resourcev1.ResourceClass, classParameters interface{}) (interface{}, error) {
+func (d driver) GetClaimParameters(
+	ctx context.Context,
+	claim *resourcev1.ResourceClaim,
+	class *resourcev1.ResourceClass,
+	classParameters interface{}) (interface{}, error) {
```
klueska, line 77: I'm not a fan of wrapping method signatures like this. In any case, this is not relevant to this PR, so it shouldn't be included here. We can have a separate discussion about whether we want to do this or not and then open a PR to address it across the code base.

cmd/dra-example-controller/driver.go
```diff
@@ -81,22 +85,45 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 	case gpucrd.GpuClaimParametersKind:
 		gc, err := d.clientset.GpuV1alpha1().GpuClaimParameters(claim.Namespace).Get(ctx, claim.Spec.ParametersRef.Name, metav1.GetOptions{})
 		if err != nil {
-			return nil, fmt.Errorf("error getting GpuClaimParameters called '%v' in namespace '%v': %v", claim.Spec.ParametersRef.Name, claim.Namespace, err)
+			return nil, fmt.Errorf(
+				"error getting GpuClaimParameters called '%v' in namespace '%v': %v",
+				claim.Spec.ParametersRef.Name, claim.Namespace, err)
```
klueska, line 90: Likewise here

cmd/dra-example-controller/driver.go
```diff
@@ -81,22 +85,45 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
...
 		}
 		err = d.gpu.ValidateClaimParameters(&gc.Spec)
 		if err != nil {
-			return nil, fmt.Errorf("error validating GpuClaimParameters called '%v' in namespace '%v': %v", claim.Spec.ParametersRef.Name, claim.Namespace, err)
+			return nil, fmt.Errorf(
+				"error validating GpuClaimParameters called '%v' in namespace '%v': %v",
+				claim.Spec.ParametersRef.Name, claim.Namespace, err)
```
klueska, line 96: And here

cmd/dra-example-controller/driver.go
```diff
@@ -81,22 +85,45 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
...
 }
 
-func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
+func (d driver) Allocate(
+	ctx context.Context,
+	claim *resourcev1.ResourceClaim,
+	claimParameters interface{},
+	class *resourcev1.ResourceClass,
+	classParameters interface{},
+	selectedNode string) (*resourcev1.AllocationResult, error) {
```
klueska, line 109: and here....

cmd/dra-example-controller/driver.go
```diff
@@ -81,22 +85,45 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
...
+
+	return buildAllocationResult(selectedNode, true), nil
+}
+
+func (d driver) allocateDelayedClaim(
+	claim *resourcev1.ResourceClaim,
+	claimParameters interface{},
+	class *resourcev1.ResourceClass,
+	classParameters interface{},
+	selectedNode string) error {
```
klueska, line 126: Let's undo all of these, and then I'll do another pass on the PR

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -12,7 +12,7 @@ allowDefaultNamespace: false
 imagePullSecrets: []
 image:
   repository: registry.example.com/dra-example-driver
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
klueska, line 15: This doesn't seem relevant to this PR -- or is it somehow needed for immediate allocation to work properly?

cmd/dra-example-controller/gpu.go
```diff
@@ -59,6 +59,35 @@ func (g *gpudriver) Allocate(crd *nascrd.NodeAllocationState, claim *resourcev1.
...
 
+func (g *gpudriver) AllocateImmediately(crd *nascrd.NodeAllocationState, ca *controller.ClaimAllocation) error {
+	claimUID := string(ca.Claim.UID)
+
+	cas := []*controller.ClaimAllocation{ca}
+	allocated := g.allocate(crd, cas)
+
+	claimParams := ca.ClaimParameters.(*gpucrd.GpuClaimParametersSpec)
+	if claimParams.Count != len(allocated[claimUID]) {
+		return fmt.Errorf("Not enough resources")
```
klueska, line 70: returned errors should be lower-case

cmd/dra-example-controller/driver.go
```diff
@@ -133,17 +160,91 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
...
+
+	crdconfig := &nascrd.NodeAllocationStateConfig{
+		Namespace: d.namespace,
+		Name:      nodename,
+	}
+
+	crd := nascrd.NewNodeAllocationState(crdconfig)
+
+	client := nasclient.New(crd, d.clientset.NasV1alpha1())
+
```
klueska, line 224: nit: no new line

api/example.com/resource/gpu/nas/v1alpha1/client/client.go
byako: naslist, ok. Happened to move back to one-line with no newlines for arguments before I saw your comments, got convinced by @bart0sh.

cmd/dra-example-controller/driver.go
```diff
@@ -94,9 +94,17 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 
 func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
 	if selectedNode == "" {
-		return nil, fmt.Errorf("TODO: immediate allocations not yet supported")
+		return d.allocateImmediateClaim(claim, claimParameters, class, classParameters)
 	}
 
+	if err := d.allocateDelayedClaim(claim, claimParameters, class, classParameters, selectedNode); err != nil {
+		return nil, err
+	}
```
klueska, line 102: I try and avoid this style of:
```
if err := blah(); err != nil 
```
I can go into a long explanation of why, but the short answer is that it keeps it consistent with how all other calls that return an error along with other variables need to handle their errors. Since this is how it's been done in this codebase elsewhere, let's keep it that way for consistency.

cmd/dra-example-controller/driver.go
```diff
@@ -94,9 +94,17 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 
 func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
 	if selectedNode == "" {
-		return nil, fmt.Errorf("TODO: immediate allocations not yet supported")
+		return d.allocateImmediateClaim(claim, claimParameters, class, classParameters)
```
klueska, line 97: We should wrap the error return here with a:
```
fmt.Errorf("error performing immediate allocation: %v", err)
```

cmd/dra-example-controller/driver.go
```diff
@@ -94,9 +94,17 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
...
 	}
 
+	if err := d.allocateDelayedClaim(claim, claimParameters, class, classParameters, selectedNode); err != nil {
+		return nil, err
```
klueska, line 101: We should wrap the error return here with a:
```
fmt.Errorf("error performing delayed allocation: %v", err)
```

deployments/helm/dra-example-driver/values.yaml
byako: It is needed for iterative development, helps pulling fresh build instead of reusing old one, why wouldn't you have it by default?
Strictly speaking it is not related to the immediate allocation functionality, I'll cut it out into separate PR.

cmd/dra-example-controller/driver.go
byako: Preceding or trailing?

cmd/dra-example-controller/driver.go
```diff
@@ -94,9 +94,17 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
...
+	}
+
+	return buildAllocationResult(selectedNode, true), nil
+}
+
+func (d driver) allocateDelayedClaim(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) error {
```
klueska, line 107: It's inconsistent for this to only return an error, but `allocateImmediateClaim()` to return an AllocationResult and an error. They should probably both return an AllocationResult and an error. This should also make the body of the wrapping call simpler and more logical to follow.

cmd/dra-example-controller/driver.go
byako: Trailing, never mind.

cmd/dra-example-controller/driver.go
```diff
@@ -133,17 +141,86 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
...
 	}
 
 	onSuccess()
 
-	return buildAllocationResult(selectedNode, true), nil
+	return nil
+}
+
+func (d driver) allocateImmediateClaim(claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}) (*resourcev1.AllocationResult, error) {
+
```
klueska, line 158: erroneous newline

cmd/dra-example-controller/driver.go
```diff
@@ -133,17 +141,86 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
...
+	crdconfig := &nascrd.NodeAllocationStateConfig{
+		Namespace: d.namespace,
+	}
+	crd := nascrd.NewNodeAllocationState(crdconfig)
+
+	client := nasclient.New(crd, d.clientset.NasV1alpha1())
```
klueska, line 164: This can be pulled up into `Allocate()` with the client passed directly to the `allocateDelayedClaim()` and `allocateImmediateClaim()` calls.

cmd/dra-example-controller/gpu.go
byako: Done.

deployments/helm/dra-example-driver/values.yaml
byako: Done.

cmd/dra-example-controller/driver.go

cmd/dra-example-controller/driver.go

cmd/dra-example-controller/driver.go

demo/gpu-test-immediate-pod.yaml
```diff
@@ -0,0 +1,22 @@
+# Pod for gpu-test-immediate-claim.yaml
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: gpu-test-immediate
+  name: pod1
```
klueska, line 7: use pod0 for consistency with other tests

cmd/dra-example-controller/driver.go

demo/gpu-test-immediate-pod.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    app: pod
+spec:
+  containers:
+  - name: ctr
```
klueska, line 12: use ctr0 for consistency with other tests

cmd/dra-example-controller/gpu.go
```diff
@@ -59,6 +59,35 @@ func (g *gpudriver) Allocate(crd *nascrd.NodeAllocationState, claim *resourcev1.
 	return onSuccess, nil
 }
 
+func (g *gpudriver) AllocateImmediately(crd *nascrd.NodeAllocationState, ca *controller.ClaimAllocation) error {
```
klueska, line 62: If we call this one `AllocateImmediately` can we call the other one `AllocateDelayed` for consistency? Also, is there a reason to not keep the function signatures identical (possibly with the only difference being the removal of `selectedNode`?

cmd/dra-example-controller/gpu.go
```diff
@@ -59,6 +59,35 @@ func (g *gpudriver) Allocate(crd *nascrd.NodeAllocationState, claim *resourcev1.
...
+			UUID: gpu,
+		}
+		devices = append(devices, device)
+	}
+
+	requestedDevices := nascrd.RequestedDevices{
+		Gpu: &nascrd.RequestedGpus{
+			Devices: devices,
+		},
+	}
```
klueska, line 85: Can we factor this out to a function so that it can be reused in unsuitable nodes and here?

 Ideally this function and unsuitable nodes would be almost line-for-line identical (and very short) except for putting the request into `PendingClaimRequests`.

cmd/dra-example-controller/driver.go
```diff
@@ -133,17 +141,85 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
...
+		return nil, fmt.Errorf("could not find suitable devices: %v", err)
+	}
+
+	// sync ResourceClaimAllocations from allocations
+	err = client.Update(&crd.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("updating NodeAllocationState CRD: %v", err)
+	}
+
+	d.gpu.PendingClaimRequests.Remove(string(ca.Claim.UID))
```
klueska, line 220: Why is this necessary? When / how is this claim ending up in here?

cmd/dra-example-controller/driver.go
```diff
@@ -133,17 +141,85 @@ func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, c
...
+	for _, nodename := range nasnames {
+		if allocationResult, err := d.allocateImmediateClaimOnNode(ca, nodename); err == nil {
+			return allocationResult, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no suitable node found")
+}
+
+func (d driver) allocateImmediateClaimOnNode(ca *controller.ClaimAllocation, nodename string) (*resourcev1.AllocationResult, error) {
```
klueska, line 188: Something doesn't feel right with all the redundancies in this function (especially around creating a new client). It feels like some other abstraction / refactoring would be better, but I don't have a suggestion off the top of my head without digging into the code myself and playing with it.

demo/gpu-test-immediate-pod.yaml
byako: And pod0, yes, thanks.

cmd/dra-example-controller/gpu.go
byako: Sure, but can we at least wrap claim, claimParams, class and classParams into ClaimAllocation right in the beginning of Allocate and pass it into allocateDelayedClaim and allocateImmedateClaim, so we don't have 160 chars long parameters line in func definition?

byako: No it has to be broken down again before being passed into gpu module, never mind, I'm just underslept, sorry.

api/example.com/resource/gpu/nas/v1alpha1/client/client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) Get() error {
 	*c.nas = *crd
 	return nil
 }
+
+func (c *Client) ListNames() ([]string, error) {
+	nasnames := []string{}
```
bart0sh, line 104:
```suggestion
```

api/example.com/resource/gpu/nas/v1alpha1/client/client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) Get() error {
...
+	naslist, err := c.client.NodeAllocationStates(c.nas.Namespace).List(
+		context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
```
bart0sh, line 109: nit:
```suggestion
	}
	nasnames := make([]string, 0, len(naslist.Items))
```

cmd/dra-example-controller/gpu.go
byako: Done.

cmd/dra-example-controller/gpu.go
byako: Done.

cmd/dra-example-controller/driver.go
byako: That's true. Both of them now return allocation result.

cmd/dra-example-controller/driver.go
byako: Done.

cmd/dra-example-controller/driver.go
byako: Done.

cmd/dra-example-controller/driver.go
byako: Sure, alright.

cmd/dra-example-controller/gpu.go
```diff
@@ -150,3 +160,22 @@ func (g *gpudriver) allocate(crd *nascrd.NodeAllocationState, pod *corev1.Pod, g
 
 	return allocated
 }
+
+func (g *gpudriver) buildRequestedGpuDevices(allocatedDevices []string) nascrd.RequestedDevices {
+
+	var devices []nascrd.RequestedGpu
```
bart0sh, line 166: nit: you can use make with capacity here

cmd/dra-example-controller/driver.go
```diff
@@ -93,10 +93,24 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
...
+		allocationResult, err := d.allocateImmediateClaim(claim, claimParameters, class, classParameters)
+		if err != nil {
+			return nil, fmt.Errorf("error performing immediate allocation: %v", err)
+		}
+		return allocationResult, nil
 	}
 
+	allocationResult, err := d.allocateDelayedClaim(claim, claimParameters, class, classParameters, selectedNode)
+	if err != nil {
+		return nil, fmt.Errorf("error performing delayed allocation: %v", err)
```
bart0sh, line 107: Can we just trust the code of local functions and simply return errors as they are? This also applies to the `allocateImmediateClaim ` call above.

klueska: It's not about "trusting" it. It's about consistency and being able to trace the errors through the function calls when you see them appear in the logs.

klueska: I specifically requested @byako to add these wrappers

api/example.com/resource/gpu/nas/v1alpha1/client/client.go
```diff
@@ -99,3 +99,22 @@ func (c *Client) Get() error {
...
 	return nil
 }
+
+func (c *Client) Switch(newcrd *nascrd.NodeAllocationState) {
+	*c.nas = *newcrd
+}
+
+func (c *Client) GetAll() (map[string]*nascrd.NodeAllocationState, error) {
+	naslist, err := c.client.NodeAllocationStates(c.nas.Namespace).List(
+		context.TODO(), metav1.ListOptions{})
```
bart0sh, line 109: Does this line differ somehow? Why do we wrap?

byako: Missed this when unwrapping. Wiill fix with and rebase to renaming change

byako:
Accomodated for renamed CRD fields.

asm582:
are we planning to support `GpuClaimParameters` in immediate mode?

byako:
There's been quite some changes since this PR was submitted, I think it'd be easier to close this one and submit this code rebased and slightly refactored for Kubernetes 1.28 API as a new PR.

byako:
I can pick this up later this week if nobody else will do so by then.

klueska:
I don't think we need to close the PR -- just rebase this one.

klueska:
> For this PR I set the GPU count as 8:
> 
> ```
> apiVersion: gpu.resource.example.com/v1alpha1
> kind: GpuClaimParameters
> metadata:
>   namespace: gpu-test4
>   name: multiple-gpus
> spec:
>   count: 8
> ```
> 
> We see that only 4 claims are allocated:
> 
> ```
>   Allocated Claims:
>   b21d90f4-d444-427e-84ad-196575e2f05f:
>     Gpu:
>       Devices:
>         Uuid:  GPU-657bd2e7-f5c2-a7f2-fbaa-0d1cdc32f81b
>         Uuid:  GPU-ee3e4b55-fcda-44b8-0605-64b7a9967744
>         Uuid:  GPU-9ede7e32-5825-a11b-fa3d-bab6d47e0243
>         Uuid:  GPU-18db0e85-99e9-c746-8531-ffeb86328b39
> ```
> 
> not sure what I am missing here?

Are you sure you didn't accidentally run with an old `GpuClaimParameters` object with count set to `4`? It won't necessarily be updated using ´kubectl apply` if the object already existed with a different value set for `count`.

PR #6, opened 2023-04-13
Rename ClaimRequests->AllocatedClaims & ClaimAllocations->PreparedClaims

klueska:
After discussions with Alexey and putting slides together to present this work, it was decided this is the better naming.


byako (approved):
LGTM

cmd/dra-example-controller/gpu.go
```diff
@@ -85,22 +85,22 @@ func (g *gpudriver) UnsuitableNode(crd *nascrd.NodeAllocationState, pod *corev1.
...
-			device := nascrd.RequestedGpu{
+			device := nascrd.AllocatedGpu{
 				UUID: gpu,
 			}
 			devices = append(devices, device)
 		}
 
-		requestedDevices := nascrd.RequestedDevices{
-			Gpu: &nascrd.RequestedGpus{
+		AllocatedDevices := nascrd.AllocatedDevices{
```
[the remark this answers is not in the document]
klueska: search/replace error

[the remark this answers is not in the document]
klueska: Updated in latest

cmd/dra-example-controller/allocations.go
```diff
@@ -0,0 +1,112 @@
...
+		return nascrd.AllocatedDevices{}
+	}
+	return p.allocations[claimUID][node]
+}
+
+func (p *PerNodeAllocatedClaims) VisitNode(node string, visitor func(claimUID string, allocation nascrd.AllocatedDevices)) {
+	p.RLock()
+	for claimUID := range p.allocations {
+		if allocation, exists := p.allocations[claimUID][node]; exists {
+			p.RUnlock()
```
bart0sh, line 63: allocation is a copy. Why do we need to unlock?

cmd/dra-example-controller/allocations.go
```diff
@@ -0,0 +1,112 @@
...
+			visitor(claimUID, allocation)
+			p.RLock()
+		}
+	}
+	p.RUnlock()
+}
+
+func (p *PerNodeAllocatedClaims) Visit(visitor func(claimUID, node string, allocation nascrd.AllocatedDevices)) {
+	p.RLock()
```
bart0sh, line 72: Any reason not to use _defer_ here?

klueska: Just for symmetry, since we unlock and relock across the `visit()` callback.

cmd/dra-example-controller/allocations.go
klueska: We only need the lock to avoid a concurrent map access on `p.allocations[claimUID][node]`. Once we have pulled `allocation` out of it, we unlock, call the visit() call back with it and then relock once the `visit()` callback has completed, so that we hold the lock for the next iteration of the loop.

bart0sh: ok, thanks for the explanations.

cmd/dra-example-controller/allocations.go
bart0sh: I'd suggest to defer just for symmetry with the rest of the methods and to avoid surprises in future, when the logic may include returns. It's a nit, so up to you.

klueska: I think I'd rather leave it as is. Normally you use the pattern of lock() --> defer(unlock) when you want to lock across the whole function and make sure you *always* unlock on every return path. In this case I unlock and relock within the body of the function and there is only one return path, so I'd rather have all locking and unlocking points be explicit.

bart0sh (approved):
/lgtm

PR #7, opened 2023-04-14
Update to K8s 1.27

byako:


klueska:
It looks like 1.27.1 is already out. Should we bump to that instead?

byako:
Yep, will do, didn't even think it would be so fast.

byako:
But not so fast, 1.27.1 was release one hour ago, so packages like https://pkg.go.dev/k8s.io/api?tab=versions will be available in couple of hours. I'll update it then.

PR #8, opened 2023-04-14
Fix CRD after renaming

byako:
This got left behind it seems.


klueska:
This shouldn’t be edited directly. It is autogenerated from the comments in the go file. I must have missed it there.

byako:
I didn't edit it directly, I noticed it after I ran build-driver.sh and create-cluster.sh, they must be regenerating CRD (didn't look inside).

PR #9, opened 2023-05-02
chore: small nit fixes for unused logger, compiler check for interface

toVersus:
This PR contains a set of small nit fixes that I made when implementing my own DRA resource driver based on your great KubeCon Europe 2023 talk:

- Remove unused logrus logger from CDIHandler and its direct dependency
- Add a compiler check for the driver that implements the DRA NodeServer interface as well as the driver that implements the  DRA controller
- ~~Rename misleading function that only returns allocated devices without actually allocating devices~~

### DIFF
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -28,6 +28,8 @@ import (
 	nasclient "github.com/kubernetes-sigs/dra-example-driver/api/example.com/resource/gpu/nas/v1alpha1/client"
 )
 
+var _ drapbv1.NodeServer = &driver{}
+
 type driver struct {
 	nascrd    *nascrd.NodeAllocationState
 	nasclient *nasclient.Client
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,6 @@ require (
 	github.com/container-orchestrated-devices/container-device-interface v0.5.4
 	github.com/google/uuid v1.3.0
 	github.com/prometheus/client_golang v1.14.0
-	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.6.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.14.0
@@ -89,6 +88,7 @@ require (
 	github.com/prometheus/client_model v0.3.0 // indirect
 	github.com/prometheus/common v0.37.0 // indirect
 	github.com/prometheus/procfs v0.8.0 // indirect
+	github.com/sirupsen/logrus v1.9.0 // indirect
 	github.com/spf13/afero v1.9.2 // indirect
 	github.com/spf13/cast v1.5.0 // indirect
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
--- a/cmd/dra-example-controller/driver.go
+++ b/cmd/dra-example-controller/driver.go
@@ -45,7 +45,7 @@ type driver struct {
 	gpu       *gpudriver
 }
 
-var _ controller.Driver = (*driver)(nil)
+var _ controller.Driver = &driver{}
 
 func NewDriver(config *Config) *driver {
 	return &driver{
--- a/cmd/dra-example-kubeletplugin/cdi.go
+++ b/cmd/dra-example-kubeletplugin/cdi.go
@@ -20,8 +20,6 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/sirupsen/logrus"
-
 	cdiapi "github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
 	cdispec "github.com/container-orchestrated-devices/container-device-interface/specs-go"
 	nascrd "github.com/kubernetes-sigs/dra-example-driver/api/example.com/resource/gpu/nas/v1alpha1"
@@ -36,7 +34,6 @@ const (
 )
 
 type CDIHandler struct {
-	logger   *logrus.Logger
 	registry cdiapi.Registry
 }
 
@@ -50,12 +47,7 @@ func NewCDIHandler(config *Config) (*CDIHandler, error) {
 		return nil, fmt.Errorf("unable to refresh the CDI registry: %v", err)
 	}
 
-	logger := logrus.New()
-	logger.Out = os.Stdout
-	logger.Level = logrus.DebugLevel
-
 	handler := &CDIHandler{
-		logger:   logger,
 		registry: registry,
 	}
 


cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -28,6 +28,8 @@ import (
 	nasclient "github.com/kubernetes-sigs/dra-example-driver/api/example.com/resource/gpu/nas/v1alpha1/client"
 )
 
+var _ drapbv1.NodeServer = (*driver)(nil)
```
pohly, line 31:
```suggestion
var _ drapbv1.NodeServer =&driver{}
```

I think this is a bit more idiomatic.

go.mod
```diff
@@ -89,6 +88,7 @@ require (
 	github.com/prometheus/client_model v0.3.0 // indirect
 	github.com/prometheus/common v0.37.0 // indirect
 	github.com/prometheus/procfs v0.8.0 // indirect
+	github.com/sirupsen/logrus v1.9.0 // indirect
```
pohly, line 91: I think we discussed logging when this example-driver was initially created and concluded to defer further work on that.

My own preference would be to do logging as in other Kubernetes components, using k8s.io/component-base/logs, including JSON support.

This doesn't need to be done in this PR, but perhaps you want to pick that up as a next step?

Looking at this diff, there are some other dependencies in go.mod that are questionable: do we really need github.com/spf13/viper?

/cc @byako @klueska

pohly (commented):
Thanks for the contribution.

I'm curious about the DRA driver that you wrote. Can you share more information about it? For the feature to move forward to beta, we need to collect evidence that it is useful also for others.

cmd/dra-example-kubeletplugin/driver.go
toVersus: Fixed on both drivers.

go.mod
toVersus: > This doesn't need to be done in this PR, but perhaps you want to pick that up as a next step?

Yeah, I'll work on that in another PR.

toVersus:
> I'm curious about the DRA driver that you wrote. Can you share more information about it? For the feature to move forward to beta, we need to collect evidence that it is useful also for others.

Sorry for the lack of explanation, it's just a toy / fake resource driver. I'd like to learn how DRA works. I might be working on an HPC project in the near future, and I'm interested in SR-IOV based multiple network interfaces attached to the Pod network (related to KEP-3698: Multi-Network Requirements). Anyway, I'm not working on anything yet.

go.mod
byako: @pohly I think it's a matter of preference. It does not seem to be vital, but there must have been some convenience reason that it was added.

cmd/dra-example-controller/gpu.go
```diff
@@ -73,7 +73,7 @@ func (g *gpudriver) UnsuitableNode(crd *nascrd.NodeAllocationState, pod *corev1.
 		}
 	})
 
-	allocated := g.allocate(crd, pod, gpucas, allcas, potentialNode)
+	allocated := g.getAllocatedDevices(crd, pod, gpucas, allcas, potentialNode)
```
byako, line 76: What this function does, is (in code blocks) : 
1) it builds a list of what devices are available
2) removes what was already allocated previously from the freshly built list of available devices in 1)
3) check if resource claim was previously allocated and there is a record for it
3.1) if so - return the list of devices allocated previously
3.2) otherwise pick requested number of devices from available devices <- this is an actual allocation step, so the function name is not misleading, it is meant to do the allocation if it's absent.

byako (commented):
I disagree with the renaming of the allocate function.

cmd/dra-example-controller/gpu.go
toVersus: Thanks for the detailed explanation! I didn't understand the process inside.
I removed this commit from the PR.

go.mod
pohly: The code needs some logging solution. logrus used to be popular, but it's not a good choice anymore.

@klueska needs to confirm, but I suspect using it wasn't really a conscious decision and "just happened".

Let's discuss in https://github.com/kubernetes-sigs/dra-example-driver/issues/10

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -28,6 +28,8 @@ import (
 	nasclient "github.com/kubernetes-sigs/dra-example-driver/api/example.com/resource/gpu/nas/v1alpha1/client"
 )
 
+var _ drapbv1.NodeServer = &driver{}
```
byako, line 31: Is this necessary at all? The interface conformance is already checked when the driver is passed to kubelet-plugin helper code here: https://github.com/kubernetes-sigs/dra-example-driver/blob/794daff2849e6d3015f3fa1cbb6e01c675ec7e8f/cmd/dra-example-kubeletplugin/main.go#L223

Try renaming NodePrepareResource to nodePrepareResource, for instance, and compiler will fail with 
```
$ make build
GOOS=linux go build ./...
# github.com/kubernetes-sigs/dra-example-driver/cmd/dra-example-kubeletplugin
cmd/dra-example-kubeletplugin/main.go:224:3: cannot use driver (variable of type *driver) as "k8s.io/kubelet/pkg/apis/dra/v1alpha2".NodeServer value in argument to plugin.Start: *driver does not implement "k8s.io/kubelet/pkg/apis/dra/v1alpha2".NodeServer (missing method NodePrepareResource)
make: *** [Makefile:48: build] Error 1
```

byako: Same about the check in controller driver.

byako: Perhaps so.

pohly: It's good Golang practice to have such interface assertions.

pohly: https://github.com/kubernetes-sigs/dra-example-driver/pull/9#discussion_r1182350297 actually shows one reason for it: compilation fails in `cmd/dra-example-kubeletplugin/main.go` but the problem is in `[cmd/dra-example-kubeletplugin/driver.go](https://github.com/kubernetes-sigs/dra-example-driver/pull/9/files/9f59860f2f6b5e057c0d55394f9bca4c8b29c319#diff-37a4b723138c8e7d62ccb364c381c4965f48c5792835f756c1ad1b43b8f58116)`. If that was some separate package, that package itself might build although it's broken.

byako: That's understandable, I wouldn't even ask if it was under pkg/ .

go.mod
klueska: I don't have a preference for logging.

The viper dependency comes from the desire to use environment variables to pass options rather than command line flags. From what I could tell viper was the standard way to integrate with cobra for this functionality.

At NVIDIA we typically use [urfave/cli](https://github.com/urfave/cli) which has this functionality built in, but the go-client libraries use cobra, and this is a kubernetes-sigs project, so I figured it was better to stick with that.

pohly (approved):
/lgtm
/approve

Let's merge this and iterate on further changes in separate PRs. I'll look at loogging.

Issue #12, opened 2023-05-03
enable unit testing and linting


pohly:
Before we merge PRs, we should be certain that `make build` and golangci-lint pass.

My preference would be to use GitHub actions, similar to go-logr: https://github.com/go-logr/logr/tree/master/.github/workflows

Prow doesn't properly work with GitHub actions (might merge before GitHub action is done), but that's an acceptable compromise. Using Prow would be more complicated.

PR #13, opened 2023-05-12
Ignore type conversion safety

byako:
Needed to enable linters #12 , typeassert complains about skipping type conversion return value not checked, let's skip them explicitly.


pohly (approved):
/lgtm

nit: short comment explaining why the type must be right would help people reading the code.

PR #14, opened 2023-05-12
Add GitHub actions

byako:
Implements #12 . Two linters are disabled ATM, plenty of complaints, I'll create an issue to fix them and then we can enable gocritic and revive linters.
Apologies, this includes https://github.com/kubernetes-sigs/dra-example-driver/pull/13 so please review that one first, otherwise linters complain.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,0 +1,30 @@
+name: Run tests
+
+on: [ push, pull_request ]
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        version: ['1.20' ]
+        platform: [ ubuntu-latest, macos-latest ]
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: ${{ matrix.version }}
+    - name: Checkout code
+      uses: actions/checkout@v3
+    - name: Update vendor modules
+      run: go mod vendor
+    - name: Build
+      run: make PREFIX=artifacts cmds
+    - name: List binaries
+      run: ls -al artifacts/
+    # no tests yet
+    # - name: Test
+    #   run: go test -v -race ./...
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,15 @@
+# please keep this alphabetized
+linters:
+  enable:
+    - asciicheck
+    - forcetypeassert
+    - gocritic
+    - godot
+    - gofmt
+    - goimports
+    - misspell
+    - stylecheck
+
+run:
+  tests: true
+  timeout: 1m
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,20 @@
+name: Run lint
+
+on: [ push, pull_request ]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: update vendor
+      run: go mod vendor
+    - name: Lint
+      uses: golangci/golangci-lint-action@v3
+      with:
+        version: latest
+        args: -v --timeout 5m
+        skip-cache: true


.golangci.yaml
```diff
@@ -0,0 +1,26 @@
+run:
+  timeout: 1m
+  tests: true
+
+linters:
+  disable-all: true
```
pohly, line 6: Instead of disabling all linters and then enabling some, can you instead leave the default linters in place and only disable those which currently produce too many warnings?

byako: This was a copy-paste from https://github.com/go-logr/logr/blob/master/.golangci.yaml, apart from two commented out linters that did produce many warnings. 
The disabled linters are not default ones, default linters do not produce any findings at all.
I removed `disable-all` config line, removed default linters from enabled-ones, and added godot. We'll need one more code-changing PR before this can be merged to fix comments to end with a dot.

byako:
PR adding dots to api/ comments https://github.com/kubernetes-sigs/dra-example-driver/pull/16 should go first

.golangci.yaml
```diff
@@ -0,0 +1,19 @@
...
+linters:
+  enable:
+    - asciicheck
+    - forcetypeassert
+    #- gocritic
+    - godot
+    - gofmt
+    - goimports
+    - misspell
+    #- revive
```
pohly, line 14: I'm unsure about godot, but don't mind having it. It's actually a style that I prefer myself...

For reference, linting in Kubernetes uses:
https://github.com/kubernetes/kubernetes/blob/f70c26d49510adf0bed1736e12cd48849b33ed6c/hack/golangci.yaml#L29-L36

Let's add all of those.

pohly: logcheck will not work as golangci-lint plugin and ginkgolinter is not useful here, so exclude those.

byako: Logcheck is not default, is it? https://golangci-lint.run/usage/linters/#ginkgolinter says defaults are: errcheck, gosimple, govet, ineffassign, staticcheck, typecheck, unused

bart0sh:
Do I understand correctly that we're using default golangci-lint checks + enabling these ones on top?
- asciicheck
- forcetypeassert
#- gocritic
- godot
- gofmt
- goimports
- misspell
#- revive

.golangci.yaml
pohly: logcheck is a Kubernetes-specific plugin.

My proposal is to use:
```
linters:
  enable:
    - asciicheck
    - forcetypeassert
    - gocritic
    - godot
    - gofmt
    - goimports
    - misspell
    - stylecheck
```

That's a combination of what go-logr and Kubernetes enabled, with godot enabled because @byako suggested it.

byako:
Yes, that's correct.

.golangci.yaml
byako: lgtm, most of revive complaints was about missing comment for exported entity, should be safe to skip it.

byako: Added style check, removed revive, added section from k/k linter config like @bart0sh suggested.

byako:
PR https://github.com/kubernetes-sigs/dra-example-driver/pull/16 is still needed, and no other changes appear to be needed to enable GH-actions linter, so we can proceed with PR 16 before this one.

byako:
Removed comments about strict linting references from .golangci.yaml, copy-paste artifact, we don't have strict linting in here.

bart0sh:
/lgtm
but we need to merge #16 first

.golangci.yaml
pohly: My understanding was that @bart0sh only linked that additional part for reference and wasn't suggesting that we should do the same.

Kubernetes has to be very conservative for existing code. We can do better.

.golangci.yaml
```diff
@@ -0,0 +1,31 @@
...
+  gocritic:
+    enabled-checks:
+      - equalFold
+      - boolExprSimplify
+  staticcheck:
+    checks:
+      - "all"
+  stylecheck:
+    checks:
+      - "ST1019"   # Importing the same package multiple times - enabled in golangci-strict.yaml.
```
pohly, line 31: I don't think we need any of these special `linters-settings`.

.golangci.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - godot
+    - gofmt
+    - goimports
+    - misspell
+    - stylecheck
+
+issues:
+  exclude-use-default: false
+  max-issues-per-linter: 0
+  max-same-issues: 10
```
pohly, line 19: Same here: let's use the golangci-lint defaults.

.golangci.yaml
pohly: /hold

.golangci.yaml
pohly: I suppose we should? Kubernetes had to choose linter rules to match the code. In this much smaller project it is feasible to choose linter rules, then make the code comply.

byako: After poking around - I disabled one check for gocritic - singleCaseSwitch, it is intended to be a single case switch where it complains due to potentially more resource types implementation.

.golangci.yaml
byako: Removed.

.golangci.yaml
byako: Rest of the config was wiped, good to go IMHO

.golangci.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - asciicheck
+    - forcetypeassert
+    - gocritic
+    - godot
+    - gofmt
+    - goimports
+    - misspell
+    - stylecheck
+
+linters-settings: # please keep this alphabetized
```
pohly, line 16: The comment makes sense, but then we should also have it above.

.golangci.yaml
```diff
@@ -0,0 +1,19 @@
...
+  gocritic:
+    disabled-checks:
+      - singleCaseSwitch
```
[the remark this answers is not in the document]
pohly: https://github.com/kubernetes-sigs/dra-example-driver/pull/14#discussion_r1194690343
> After poking around - I disabled one check for gocritic - singleCaseSwitch, it is intended to be a single case switch where it complains due to potentially more resource types implementation.

Instead of disabling the check globally, disable it for that particular case switch with `//nolint:gocritic // <some comment explaining why>`.

https://golangci-lint.run/usage/false-positives/#nolint-directive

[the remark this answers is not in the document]
byako: Fixed in https://github.com/kubernetes-sigs/dra-example-driver/pull/18. Removing this config section.

.golangci.yaml
byako: Moved to top.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        go-version: ${{ matrix.version }}
+    - name: Checkout code
+      uses: actions/checkout@v3
+    - name: Update vendor modules
+      run: go mod vendor
+    - name: Build
+      run: make PREFIX=artifacts cmds
+    - name: List binaries
+      run: ls -al artifacts/
```
bart0sh, line 27: Why don't we run golangci here?

bart0sh:
@byako Why new checks are not running on this PR?

byako:
> @byako Why new checks are not running on this PR?

Probably because they are not in this git tree, so github does not trigger action for this tree. It does however trigger them in my fork where the PR originates from: https://github.com/byako/dra-example-driver/actions

klueska:
I don't know enough about github actions to properly review this PR. If others that are more familiar think it is ready though, I can approve it.

pohly (approved):
/lgtm
/approve

I verified locally that golangci-lint passes with current master.

byako:
gocritic fix was merged, letting this go through.

/unhold

PR #17, opened 2023-05-15
Ensure that the DRA controller only gets deployed on control plane nodes

klueska:
The taint was set previously (meaning it COULD be scheduled on control plan nodes), but without the nodeSelector it is not guaranteed to be.

### DIFF
--- a/deployments/helm/dra-example-driver/values.yaml
+++ b/deployments/helm/dra-example-driver/values.yaml
@@ -29,7 +29,8 @@ controller:
   priorityClassName: "system-node-critical"
   podAnnotations: {}
   podSecurityContext: {}
-  nodeSelector: {}
+  nodeSelector:
+    node-role.kubernetes.io/control-plane: ""
   tolerations:
   - key: node-role.kubernetes.io/master
     operator: Exists


byako (approved):
lgtm

bart0sh:
@klueska why does the controller have to run only on a control plain node?

byako:
Worker nodes come and go, especially GPU-enabled public cloud ones, to save costs, for instance. And when such a node goes down, deployment does not immediately create new replacement pod for controller that used to reside on the node that has been shut down without drainage. We've seen it too, it's better to keep the controller on the nodes that are guaranteed to be up, control-plane for instance.
Granted, it can be any other node, that the cluster admin knows will not be shutdown, but we are unaware of such nodes of course.

klueska:
Can you give an example of a scenario where you would explicitly want it to run on a _non_-control-plane node? Normally you don't want anything running on your compute / worker nodes other than actual workloads (and daemonsets) so that the resources on those nodes don't get eaten up by control-level processes.

bart0sh:
Sure. Cluster admins may want to have a dedicate node(s) for this kind of services, especially on large clusters where control nodes are quite loaded.
In any case it's up to admins to decide how to setup their clusters. We shouldn't create wrong assumptions for them in my opinion.

klueska:
I still believe this is a worthwhile change based on what I said above.

klueska:
ping

PR #16, opened 2023-05-15
Add dot to api/ comments

byako:
Found by godot linter from https://github.com/kubernetes-sigs/dra-example-driver/pull/14

PR #18, opened 2023-05-16
Fix gocritic findings

byako:
Make gocritic happy.

### DIFF
--- a/cmd/dra-example-controller/driver.go
+++ b/cmd/dra-example-controller/driver.go
@@ -77,6 +77,7 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 	if claim.Spec.ParametersRef.APIGroup != DriverAPIGroup {
 		return nil, fmt.Errorf("incorrect API group: %v", claim.Spec.ParametersRef.APIGroup)
 	}
+
 	switch claim.Spec.ParametersRef.Kind {
 	case gpucrd.GpuClaimParametersKind:
 		gc, err := d.clientset.GpuV1alpha1().GpuClaimParameters(claim.Namespace).Get(ctx, claim.Spec.ParametersRef.Name, metav1.GetOptions{})
@@ -88,8 +89,9 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 			return nil, fmt.Errorf("error validating GpuClaimParameters called '%v' in namespace '%v': %v", claim.Spec.ParametersRef.Name, claim.Namespace, err)
 		}
 		return &gc.Spec, nil
+	default:
+		return nil, fmt.Errorf("unknown ResourceClaim.ParametersRef.Kind: %v", claim.Spec.ParametersRef.Kind)
 	}
-	return nil, fmt.Errorf("unknown ResourceClaim.ParametersRef.Kind: %v", claim.Spec.ParametersRef.Kind)
 }
 
 func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
@@ -244,18 +246,20 @@ func (d driver) unsuitableNode(ctx context.Context, pod *corev1.Pod, allcas []*c
 
 	perKindCas := make(map[string][]*controller.ClaimAllocation)
 	for _, ca := range allcas {
-		var kind string
-		switch ca.ClaimParameters.(type) {
-		case *gpucrd.GpuClaimParametersSpec:
-			kind = gpucrd.GpuClaimParametersKind
+		switch ca.Claim.Spec.ParametersRef.Kind {
+		case gpucrd.GpuClaimParametersKind:
+			perKindCas[gpucrd.GpuClaimParametersKind] = append(perKindCas[gpucrd.GpuClaimParametersKind], ca)
+		default:
+			return fmt.Errorf("unknown ResourceClaimParameters kind: %+v", ca.Claim.Spec.ParametersRef.Kind)
 		}
-		perKindCas[kind] = append(perKindCas[kind], ca)
 	}
 	for _, kind := range []string{gpucrd.GpuClaimParametersKind} {
 		var err error
 		switch kind {
 		case gpucrd.GpuClaimParametersKind:
 			err = d.gpu.UnsuitableNode(crd, pod, perKindCas[kind], allcas, potentialNode)
+		default:
+			err = fmt.Errorf("unknown ResourceClaimParameters kind: %+v", kind)
 		}
 		if err != nil {
 			return fmt.Errorf("error processing '%v': %v", kind, err)
--- a/cmd/dra-example-controller/gpu.go
+++ b/cmd/dra-example-controller/gpu.go
@@ -113,6 +113,8 @@ func (g *gpudriver) allocate(crd *nascrd.NodeAllocationState, pod *corev1.Pod, g
 		switch device.Type() {
 		case nascrd.GpuDeviceType:
 			available[device.Gpu.UUID] = device.Gpu
+		default:
+			// skip other devices
 		}
 	}
 
@@ -122,6 +124,8 @@ func (g *gpudriver) allocate(crd *nascrd.NodeAllocationState, pod *corev1.Pod, g
 			for _, device := range allocation.Gpu.Devices {
 				delete(available, device.UUID)
 			}
+		default:
+			// skip other devices
 		}
 	}
 
--- a/cmd/dra-example-kubeletplugin/cdi.go
+++ b/cmd/dra-example-kubeletplugin/cdi.go
@@ -111,6 +111,8 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices *PreparedDev
 			spec.Devices = append(spec.Devices, cdiDevice)
 			gpuIdx++
 		}
+	default:
+		return fmt.Errorf("unknown device type: %v", devices.Type())
 	}
 
 	minVersion, err := cdiapi.MinimumRequiredVersion(spec)
@@ -127,7 +129,7 @@ func (cdi *CDIHandler) DeleteClaimSpecFile(claimUID string) error {
 	return cdi.registry.SpecDB().RemoveSpec(specName)
 }
 
-func (cdi *CDIHandler) GetClaimDevices(claimUID string, devices *PreparedDevices) []string {
+func (cdi *CDIHandler) GetClaimDevices(claimUID string, devices *PreparedDevices) ([]string, error) {
 	cdiDevices := []string{
 		cdiapi.QualifiedName(cdiVendor, cdiClass, cdiCommonDeviceName),
 	}
@@ -138,7 +140,9 @@ func (cdi *CDIHandler) GetClaimDevices(claimUID string, devices *PreparedDevices
 			cdiDevice := cdiapi.QualifiedName(cdiVendor, cdiClass, device.uuid)
 			cdiDevices = append(cdiDevices, cdiDevice)
 		}
+	default:
+		return nil, fmt.Errorf("unknown device type: %v", devices.Type())
 	}
 
-	return cdiDevices
+	return cdiDevices, nil
 }
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -55,7 +55,12 @@ func NewDriver(config *Config) (*driver, error) {
 			return err
 		}
 
-		err = client.Update(state.GetUpdatedSpec(&config.nascrd.Spec))
+		updatedSpec, err := state.GetUpdatedSpec(&config.nascrd.Spec)
+		if err != nil {
+			return fmt.Errorf("error getting updated CR spec: %v", err)
+		}
+
+		err = client.Update(updatedSpec)
 		if err != nil {
 			return err
 		}
@@ -101,7 +106,12 @@ func (d *driver) NodePrepareResource(ctx context.Context, req *drapbv1.NodePrepa
 			return fmt.Errorf("error allocating devices for claim '%v': %v", req.ClaimUid, err)
 		}
 
-		err = d.nasclient.Update(d.state.GetUpdatedSpec(&d.nascrd.Spec))
+		updatedSpec, err := d.state.GetUpdatedSpec(&d.nascrd.Spec)
+		if err != nil {
+			return fmt.Errorf("error getting updated CR spec: %v", err)
+		}
+
+		err = d.nasclient.Update(updatedSpec)
 		if err != nil {
 			if err := d.state.Unprepare(req.ClaimUid); err != nil {
 				klog.Errorf("Failed to unprepare after claim '%v' Update() error: %v", req.ClaimUid, err)
@@ -128,7 +138,12 @@ func (d *driver) NodeUnprepareResource(ctx context.Context, req *drapbv1.NodeUnp
 			return fmt.Errorf("error unpreparing devices for claim '%v': %v", req.ClaimUid, err)
 		}
 
-		err = d.nasclient.Update(d.state.GetUpdatedSpec(&d.nascrd.Spec))
+		updatedSpec, err := d.state.GetUpdatedSpec(&d.nascrd.Spec)
+		if err != nil {
+			return fmt.Errorf("error getting updated CR spec: %v", err)
+		}
+
+		err = d.nasclient.Update(updatedSpec)
 		if err != nil {
 			return err
 		}
--- a/cmd/dra-example-kubeletplugin/state.go
+++ b/cmd/dra-example-kubeletplugin/state.go
@@ -92,7 +92,11 @@ func (s *DeviceState) Prepare(claimUID string, allocation nascrd.AllocatedDevice
 	defer s.Unlock()
 
 	if s.prepared[claimUID] != nil {
-		return s.cdi.GetClaimDevices(claimUID, s.prepared[claimUID]), nil
+		cdiDevices, err := s.cdi.GetClaimDevices(claimUID, s.prepared[claimUID])
+		if err != nil {
+			return nil, fmt.Errorf("unable to get CDI devices names: %v", err)
+		}
+		return cdiDevices, nil
 	}
 
 	prepared := &PreparedDevices{}
@@ -101,6 +105,8 @@ func (s *DeviceState) Prepare(claimUID string, allocation nascrd.AllocatedDevice
 	switch allocation.Type() {
 	case nascrd.GpuDeviceType:
 		prepared.Gpu, err = s.prepareGpus(claimUID, allocation.Gpu)
+	default:
+		err = fmt.Errorf("unknown device type: %v", allocation.Type())
 	}
 	if err != nil {
 		return nil, fmt.Errorf("allocation failed: %v", err)
@@ -113,7 +119,11 @@ func (s *DeviceState) Prepare(claimUID string, allocation nascrd.AllocatedDevice
 
 	s.prepared[claimUID] = prepared
 
-	return s.cdi.GetClaimDevices(claimUID, s.prepared[claimUID]), nil
+	cdiDevices, err := s.cdi.GetClaimDevices(claimUID, s.prepared[claimUID])
+	if err != nil {
+		return nil, fmt.Errorf("unable to get CDI devices names: %v", err)
+	}
+	return cdiDevices, nil
 }
 
 func (s *DeviceState) Unprepare(claimUID string) error {
@@ -130,6 +140,8 @@ func (s *DeviceState) Unprepare(claimUID string) error {
 		if err != nil {
 			return fmt.Errorf("unprepare failed: %v", err)
 		}
+	default:
+		return fmt.Errorf("unknown device type: %v", s.prepared[claimUID].Type())
 	}
 
 	err := s.cdi.DeleteClaimSpecFile(claimUID)
@@ -142,14 +154,22 @@ func (s *DeviceState) Unprepare(claimUID string) error {
 	return nil
 }
 
-func (s *DeviceState) GetUpdatedSpec(inspec *nascrd.NodeAllocationStateSpec) *nascrd.NodeAllocationStateSpec {
+func (s *DeviceState) GetUpdatedSpec(inspec *nascrd.NodeAllocationStateSpec) (*nascrd.NodeAllocationStateSpec, error) {
 	s.Lock()
 	defer s.Unlock()
 
 	outspec := inspec.DeepCopy()
-	s.syncAllocatableDevicesToCRDSpec(outspec)
-	s.syncPreparedDevicesToCRDSpec(outspec)
-	return outspec
+	err := s.syncAllocatableDevicesToCRDSpec(outspec)
+	if err != nil {
+		return nil, fmt.Errorf("synching allocatable devices to CR spec: %v", err)
+	}
+
+	err = s.syncPreparedDevicesToCRDSpec(outspec)
+	if err != nil {
+		return nil, fmt.Errorf("synching prepared devices to CR spec: %v", err)
+	}
+
+	return outspec, nil
 }
 
 func (s *DeviceState) prepareGpus(claimUID string, allocated *nascrd.AllocatedGpus) (*PreparedGpus, error) {
@@ -172,7 +192,7 @@ func (s *DeviceState) unprepareGpus(claimUID string, devices *PreparedDevices) e
 	return nil
 }
 
-func (s *DeviceState) syncAllocatableDevicesToCRDSpec(spec *nascrd.NodeAllocationStateSpec) {
+func (s *DeviceState) syncAllocatableDevicesToCRDSpec(spec *nascrd.NodeAllocationStateSpec) error {
 	gpus := make(map[string]nascrd.AllocatableDevice)
 	for _, device := range s.allocatable {
 		gpus[device.uuid] = nascrd.AllocatableDevice{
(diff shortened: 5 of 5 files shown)


pohly (changes requested):
I think a better solution is to add a `default:` case with an error when reached.

It's dead code right now, but better than silently doing the wrong thing when new values get added and a switch statement doesn't get updated.

bart0sh:
Can we simply use 'if' instead of the switch with one case? I thought that gocritic suggests exactly that.

byako:
Sure, but as soon as we introduce more types - it will complain again and suggest we replace if - else with switch, this for instance: https://github.com/kubernetes-sigs/dra-example-driver/pull/16/commits/d6932478702688c6130f97549f7e405b08bb12e7

byako:
@pohly for default switch case - as far as I can tell, at least controller code does not benefit from default case - the allocation is called only if getClassParameters and getClaimParameters succeed, which they don't if the type is unsupported by the controller, so having same safety in N places in controller just duplicates dead code.
For kubelet-plugin I'm checking if it's realistic to have a call to it with unsupported type.

pohly:
> so having same safety in N places in controller just duplicates dead code

So why have the switch statement at all?

The `default` is dead code now, but it should keep gocritic happy and serves as safety net when new types get added.

byako:
> So why have the switch statement at all?

Because there _might_ be more (this is how it is implemented in Nvidia GPU resource driver), and this is a reference implementation / suggestion how this can be implemented to support different device types. We can add second type with actual HW discovered, we talked about it with @klueska, and it seemed like out of the two (immediate allocation, device discovery & second device type) - the immediate allocation was doable before conference, and second device should have followed. Here we are. I'm still planning on adding second device type with HW discovery and device file nodes in CDI spec after the immediate allocation is merged, then the default switch case will not be needed for linter to be happy.

I'll rewrite it with default case.

pohly:
> then the default switch case will not be needed for linter to be happy

But it's still good to have it even then because the code then fails more nicely when forgetting to add a new case when adding yet another new type.

bart0sh:
> Because there **might be** more 

I still think that with 'if's the code would be more readable. When there are more cases it would not be a rocket science to replace 'if' with 'switch'.

klueska:
They are only switches because the NVIDIA driver that this is derived from has 3 cases here. I thought it would be easier to leave them as switch cases to make it clear what the pattern would be if more claimparameter objects were added in the future. As of now, now switch statement (or if statement) is needed at all. It's solely for the purpose of making it obvious how to extend it in the future.

cmd/dra-example-controller/driver.go
```diff
@@ -244,17 +245,15 @@ func (d driver) unsuitableNode(ctx context.Context, pod *corev1.Pod, allcas []*c
 
 	perKindCas := make(map[string][]*controller.ClaimAllocation)
 	for _, ca := range allcas {
-		var kind string
-		switch ca.ClaimParameters.(type) {
-		case *gpucrd.GpuClaimParametersSpec:
-			kind = gpucrd.GpuClaimParametersKind
+		if ca.Claim.Spec.ParametersRef.Kind == gpucrd.GpuClaimParametersKind {
+			perKindCas[gpucrd.GpuClaimParametersKind] = append(perKindCas[gpucrd.GpuClaimParametersKind], ca)
+		} else {
```
pohly, line 250: Idiomatic Go would be to check for the error case first, then continue without indention. if/else hardly ever get used.

In this case, a switch with default would be clearer.

pohly (changes requested):
Now the PR uses a mixture of if/else and switch.

Let's use switch consistently, with a default case for the unexpected situations.

cmd/dra-example-controller/driver.go
```diff
@@ -244,18 +246,19 @@ func (d driver) unsuitableNode(ctx context.Context, pod *corev1.Pod, allcas []*c
...
+			return fmt.Errorf("unsupported ResourceClaimParameters kind: %+v", ca.Claim.Spec.ParametersRef.Kind)
 		}
-		perKindCas[kind] = append(perKindCas[kind], ca)
 	}
 	for _, kind := range []string{gpucrd.GpuClaimParametersKind} {
 		var err error
 		switch kind {
 		case gpucrd.GpuClaimParametersKind:
 			err = d.gpu.UnsuitableNode(crd, pod, perKindCas[kind], allcas, potentialNode)
+		default:
```
pohly, line 261: Returning an error missing here?

cmd/dra-example-controller/gpu.go
```diff
@@ -110,15 +110,13 @@ func (g *gpudriver) allocate(crd *nascrd.NodeAllocationState, pod *corev1.Pod, g
 	available := make(map[string]*nascrd.AllocatableGpu)
 
 	for _, device := range crd.Spec.AllocatableDevices {
-		switch device.Type() {
-		case nascrd.GpuDeviceType:
+		if device.Type() == nascrd.GpuDeviceType {
```
pohly, line 113: Let's keep the switch.

cmd/dra-example-controller/gpu.go
```diff
@@ -110,15 +110,13 @@ func (g *gpudriver) allocate(crd *nascrd.NodeAllocationState, pod *corev1.Pod, g
...
 			available[device.Gpu.UUID] = device.Gpu
 		}
 	}
 
 	for _, allocation := range crd.Spec.AllocatedClaims {
-		switch allocation.Type() {
-		case nascrd.GpuDeviceType:
+		if allocation.Type() == nascrd.GpuDeviceType {
```
pohly, line 119: Let's keep the switch here.

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -111,6 +111,8 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices *PreparedDev
 			spec.Devices = append(spec.Devices, cdiDevice)
 			gpuIdx++
 		}
+	default:
+		return fmt.Errorf("unknown device type")
```
pohly, line 115: Include the actual device type in the error?

Otherwise it's just a static string and `errors.New` would be better.

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -138,6 +140,8 @@ func (cdi *CDIHandler) GetClaimDevices(claimUID string, devices *PreparedDevices
 			cdiDevice := cdiapi.QualifiedName(cdiVendor, cdiClass, device.uuid)
 			cdiDevices = append(cdiDevices, cdiDevice)
 		}
+	default:
+		return cdiDevices
```
pohly, line 144: Shouldn't this be an error?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -101,6 +101,8 @@ func (s *DeviceState) Prepare(claimUID string, allocation nascrd.AllocatedDevice
 	switch allocation.Type() {
 	case nascrd.GpuDeviceType:
 		prepared.Gpu, err = s.prepareGpus(claimUID, allocation.Gpu)
+	default:
+		err = fmt.Errorf("unknown allocated device type")
```
pohly, line 105: Include type.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -130,6 +132,8 @@ func (s *DeviceState) Unprepare(claimUID string) error {
 		if err != nil {
 			return fmt.Errorf("unprepare failed: %v", err)
 		}
+	default:
```
pohly, line 135: Include type.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -202,6 +206,8 @@ func (s *DeviceState) syncPreparedDevicesFromCRDSpec(spec *nascrd.NodeAllocation
 			for _, d := range devices.Gpu.Devices {
 				prepared[claim].Gpu.Devices = append(prepared[claim].Gpu.Devices, gpus[d.UUID].GpuInfo)
 			}
+		default:
```
pohly, line 209: Include type.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -222,6 +228,7 @@ func (s *DeviceState) syncPreparedDevicesToCRDSpec(spec *nascrd.NodeAllocationSt
 				}
 				prepared.Gpu.Devices = append(prepared.Gpu.Devices, outdevice)
 			}
+		default:
```
pohly, line 231: Error?

pohly (changes requested):

cmd/dra-example-kubeletplugin/cdi.go
byako: Cannot include actual device type, it's runtime-determined, there's no such field, it's either GPU or literally unknown device type:
https://github.com/kubernetes-sigs/dra-example-driver/blob/e6004d1e93a4ae502c644b69de25e917aa1416a2/api/example.com/resource/gpu/nas/v1alpha1/nas.go#L81

cmd/dra-example-controller/driver.go
byako: We're looping through the static list of items, the only reason the default is here is because linter would request replacing this with if. I think returning an error in here is a bit too self-doubting. because the only error that can possibly take place is somebody extending the list to range-loop through but without adding respective case in switch.

cmd/dra-example-controller/gpu.go
byako: What would the switch switch? We cannot even return an error from default in here, we just trust that there's only one type; so the default action will have to be empty like above, would that be fine ?

cmd/dra-example-controller/gpu.go
byako: What would the switch switch? We cannot even return an error from default in here, we just trust that there's only one type; so the default action will have to be empty like above, would that be fine ?

cmd/dra-example-kubeletplugin/state.go
byako: The type is literally unknown.

cmd/dra-example-kubeletplugin/state.go
byako: The type is literally unknown.

cmd/dra-example-kubeletplugin/state.go
byako: Just loggging, when implemented, there's nothing that expects an error to be returned from here or parent, unsupported devices are ignored and logged.

cmd/dra-example-controller/driver.go
bart0sh: I understand that this is a matter of taste mostly, but I'd agree with linter here.
This
```
if ca.Claim.Spec.ParametersRef.Kind != ca.Claim.Spec.ParametersRef.Kind {
        return fmt.Errorf("unsupported ResourceClaimParameters kind: %+v", ca.Claim.Spec.ParametersRef.Kind)
}
perKindCas[gpucrd.GpuClaimParametersKind] = append(perKindCas[gpucrd.GpuClaimParametersKind], ca)
```
is much more readable in my opinion.
However, I understand that it doesn't suggest that there *might* be other kinds here.

pohly: > because the only error that can possibly take place is somebody extending the list to range-loop through but without adding respective case in switch

And that is exactly what the code should be checking for.

pohly: > However, I understand that it doesn't suggest that there might be other kinds here.

Documenting that is the purpose of using `switch` instead of `if`.

cmd/dra-example-controller/gpu.go
pohly: Then change `allocate` so that it can return an `error`. Assuming that "allocation never fails" already seems suspicious.

cmd/dra-example-controller/gpu.go
pohly: See above.

cmd/dra-example-kubeletplugin/state.go
pohly: See above.

cmd/dra-example-kubeletplugin/state.go
pohly: See above.

cmd/dra-example-kubeletplugin/state.go
pohly: See above.

cmd/dra-example-kubeletplugin/state.go
pohly: As above, let's change the function so that it can return an error.

cmd/dra-example-controller/driver.go
byako: > And that is exactly what the code should be checking for.

I'd say this is a job for unit tests, The code base would explode if we tried to cover all places where the developer can shoot self in the foot.

cmd/dra-example-kubeletplugin/cdi.go
byako: Right.

cmd/dra-example-kubeletplugin/state.go
pohly: Call it defensive programming. Mistakes happen.

cmd/dra-example-controller/gpu.go
pohly: If you feel that a blank default is the right solution, then at least add a comment explaining to the reader/reviewer why the case is empty.

cmd/dra-example-controller/driver.go
byako: Forgot this, had to force-push once more, sorry.
Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/cdi.go
byako: Done.

cmd/dra-example-controller/gpu.go
byako: Done.

cmd/dra-example-controller/gpu.go
byako: That's exactly what I did. Should have put it long ago, you're right.

cmd/dra-example-controller/driver.go
bart0sh: > Documenting that is the purpose of using switch instead of if.

Yep. The question is does it worth it. And the answer is very much a matter of taste in my opinon. My answer is no, just in case it's not yet clear :)

cmd/dra-example-controller/driver.go
```diff
@@ -244,18 +246,20 @@ func (d driver) unsuitableNode(ctx context.Context, pod *corev1.Pod, allcas []*c
...
 	for _, ca := range allcas {
-		var kind string
-		switch ca.ClaimParameters.(type) {
-		case *gpucrd.GpuClaimParametersSpec:
-			kind = gpucrd.GpuClaimParametersKind
+		switch ca.Claim.Spec.ParametersRef.Kind {
+		case gpucrd.GpuClaimParametersKind:
+			perKindCas[gpucrd.GpuClaimParametersKind] = append(perKindCas[gpucrd.GpuClaimParametersKind], ca)
+		default:
+			return fmt.Errorf("unsupported ResourceClaimParameters kind: %+v", ca.Claim.Spec.ParametersRef.Kind)
```
klueska, line 253: Can we use `unknown` instead of `unsupported` to be consistent with line 93.

cmd/dra-example-controller/driver.go
```diff
@@ -244,18 +246,20 @@ func (d driver) unsuitableNode(ctx context.Context, pod *corev1.Pod, allcas []*c
...
 		}
-		perKindCas[kind] = append(perKindCas[kind], ca)
 	}
 	for _, kind := range []string{gpucrd.GpuClaimParametersKind} {
 		var err error
 		switch kind {
 		case gpucrd.GpuClaimParametersKind:
 			err = d.gpu.UnsuitableNode(crd, pod, perKindCas[kind], allcas, potentialNode)
+		default:
+			err = fmt.Errorf("unexpected ResourceClaimParameters kind: %+v", kind)
```
klueska, line 262: Can we make this the following to be consistent with line 92:
```
		err = fmt.Errorf("unknown ResourceClaim.ParametersRef.Kind: %v", claim.Spec.ParametersRef.Kind)
```

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -128,7 +138,12 @@ func (d *driver) NodeUnprepareResource(ctx context.Context, req *drapbv1.NodeUnp
 			return fmt.Errorf("error unpreparing devices for claim '%v': %v", req.ClaimUid, err)
 		}
 
-		err = d.nasclient.Update(d.state.GetUpdatedSpec(&d.nascrd.Spec))
+		updatedSpec, err := d.state.GetUpdatedSpec(&d.nascrd.Spec)
+		if err != nil {
+			return err
```
klueska, line 143: we should wrap this error the same as in `NodePrepareResources()`.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -101,7 +106,12 @@ func (d *driver) NodePrepareResource(ctx context.Context, req *drapbv1.NodePrepa
 			return fmt.Errorf("error allocating devices for claim '%v': %v", req.ClaimUid, err)
 		}
 
-		err = d.nasclient.Update(d.state.GetUpdatedSpec(&d.nascrd.Spec))
+		updatedSpec, err := d.state.GetUpdatedSpec(&d.nascrd.Spec)
+		if err != nil {
+			return fmt.Errorf("error synching devices to CR spec: %v", err)
```
klueska, line 111: This wrapped error should be:
```
			return fmt.Errorf("error getting updated spec: %v", err)
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -101,6 +101,8 @@ func (s *DeviceState) Prepare(claimUID string, allocation nascrd.AllocatedDevice
 	switch allocation.Type() {
 	case nascrd.GpuDeviceType:
 		prepared.Gpu, err = s.prepareGpus(claimUID, allocation.Gpu)
+	default:
+		err = fmt.Errorf("wrong allocated device type: %v", allocation.Type())
```
klueska, line 105: Can we make this the following to be consistent with line 115:
```
		return fmt.Errorf("unknown device type: %v", devices.Type())
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -130,6 +132,8 @@ func (s *DeviceState) Unprepare(claimUID string) error {
 		if err != nil {
 			return fmt.Errorf("unprepare failed: %v", err)
 		}
+	default:
+		return fmt.Errorf("unprepare failed: wrong prepared device type: %v", s.prepared[claimUID].Type())
```
klueska, line 136: Can we just change this to:
```
		return fmt.Errorf("unknown device type: %v", devices.Type())
```
That is the only error from within this function. The unprepare is not called if the type is wrong.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -142,14 +146,18 @@ func (s *DeviceState) Unprepare(claimUID string) error {
...
 	defer s.Unlock()
 
 	outspec := inspec.DeepCopy()
 	s.syncAllocatableDevicesToCRDSpec(outspec)
-	s.syncPreparedDevicesToCRDSpec(outspec)
-	return outspec
+	err := s.syncPreparedDevicesToCRDSpec(outspec)
+	if err != nil {
+		return nil, fmt.Errorf("synching prepared devices to CR spec: %v", err)
+	}
```
klueska, line 158: Can we make the `syncAllocatableDevicesToCRDSpec()` function also return an error for symmetry in the calls here

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -222,8 +232,11 @@ func (s *DeviceState) syncPreparedDevicesToCRDSpec(spec *nascrd.NodeAllocationSt
 				}
 				prepared.Gpu.Devices = append(prepared.Gpu.Devices, outdevice)
 			}
+		default:
+			return fmt.Errorf("wrong prepared device type: %v", devices.Type())
```
klueska, line 236: Can we make this the following to be consistent with line 115:
```
		return fmt.Errorf("unknown device type: %v", devices.Type())
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -222,8 +232,11 @@ func (s *DeviceState) syncPreparedDevicesToCRDSpec(spec *nascrd.NodeAllocationSt
...
 		}
 		outcas[claim] = prepared
 	}
 	spec.PreparedClaims = outcas
+	return nil
```
klueska, line 241: nit: newline before this

cmd/dra-example-kubeletplugin/driver.go
byako: Line 148 below shoould be wrapped too then ?

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/state.go
byako: Done.

cmd/dra-example-kubeletplugin/driver.go
byako: Done.

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -138,7 +140,9 @@ func (cdi *CDIHandler) GetClaimDevices(claimUID string, devices *PreparedDevices
 			cdiDevice := cdiapi.QualifiedName(cdiVendor, cdiClass, device.uuid)
 			cdiDevices = append(cdiDevices, cdiDevice)
 		}
+	default:
+		return nil, fmt.Errorf("wrong device type: %v", devices.Type())
```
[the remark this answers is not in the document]
byako: Done.

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -111,6 +111,8 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices *PreparedDev
 			spec.Devices = append(spec.Devices, cdiDevice)
 			gpuIdx++
 		}
+	default:
+		return fmt.Errorf("wrong device type: %v", devices.Type())
```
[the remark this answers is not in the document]
byako: Done.

cmd/dra-example-controller/driver.go

cmd/dra-example-controller/driver.go

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -55,7 +55,12 @@ func NewDriver(config *Config) (*driver, error) {
 			return err
 		}
 
-		err = client.Update(state.GetUpdatedSpec(&config.nascrd.Spec))
+		updatedSpec, err := state.GetUpdatedSpec(&config.nascrd.Spec)
+		if err != nil {
+			return err
```
klueska, line 60: Can we wrap this?

byako: Wrapped.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -113,7 +115,7 @@ func (s *DeviceState) Prepare(claimUID string, allocation nascrd.AllocatedDevice
 
 	s.prepared[claimUID] = prepared
 
-	return s.cdi.GetClaimDevices(claimUID, s.prepared[claimUID]), nil
+	return s.cdi.GetClaimDevices(claimUID, s.prepared[claimUID])
```
klueska, line 118: This should be wrapped for consistency.

byako: Wrapped.

klueska:
This PR broke the running of the embedeed demo.

Specificalls this:
```
    for _, ca := range allcas {
-       var kind string
-       switch ca.ClaimParameters.(type) {
-       case *gpucrd.GpuClaimParametersSpec:
-           kind = gpucrd.GpuClaimParametersKind
+       switch ca.Claim.Spec.ParametersRef.Kind {
+       case gpucrd.GpuClaimParametersKind:
+           perKindCas[gpucrd.GpuClaimParametersKind] = append(perKindCas[gpucrd.GpuClaimParametersKind], ca)
+       default:
+           return fmt.Errorf("unknown ResourceClaimParameters kind: %+v", ca.Claim.Spec.ParametersRef.Kind)
        }
-       perKindCas[kind] = append(perKindCas[kind], ca)
    }
```

The controller now crashes if `ParametersRef` is `nil` (which it often is).

Has anyone actually ran the demo since this change and had it work?

klueska:
Fix posted here:
https://github.com/kubernetes-sigs/dra-example-driver/pull/21

pohly:
> Has anyone actually ran the demo since this change

Probably not. We know that we need better automated testing :cry:

klueska:
No worries. I just wanted to make sure there wasn't something messed up with my environment somehow where it was working for others, but not me.

PR #22, opened 2023-06-26
Use the same base image as test/e2e/dra does

klueska:
This PR https://github.com/kubernetes/kubernetes/pull/118030 introduced the use of a custom kind base image with containerd preinstalled (instead of building one from scratch).

We now use this same base image for our kind cluster as well.


pohly (approved):
/lgtm
/approve

PR #24, opened 2023-08-07
Controller batch calls support

byako:
Since we don't really have any sophisticated calculations of resource / sub-resources that would require consideration of several claims at once, it should be enough to make controller iterate over batch of claims and do one claim allocation at a time.


cmd/dra-example-controller/driver.go
```diff
@@ -94,9 +94,13 @@ func (d driver) GetClaimParameters(ctx context.Context, claim *resourcev1.Resour
 	}
 }
 
-func (d driver) Allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
+func (d driver) Allocate(ctx context.Context, claimAllocations []*controller.ClaimAllocation, selectedNode string) {
+
 	if selectedNode == "" {
-		return nil, fmt.Errorf("TODO: immediate allocations not yet supported")
+		for _, claim := range claimAllocations {
```
pohly, line 100: This PR adds multiple for loops inside `Allocate` and as part of changes indention of several code lines.

I suspect a much smaller patch would be to implement `Allocate` with a for loop that calls an `allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error)` method.

It's probably also going to be more readable.

pohly: Something like this:
```
func (d driver) Allocate(ctx context.Context, claimAllocations []*controller.ClaimAllocation, selectedNode string) {
   for i := range claimAllocations {
       claimAllocations[i].Allocation, claimAllocations[i].Error = d.allocate(...)
   }
}

func (d driver) allocate(ctx context.Context, claim *resourcev1.ResourceClaim, claimParameters interface{}, class *resourcev1.ResourceClass, classParameters interface{}, selectedNode string) (*resourcev1.AllocationResult, error) {
  <same code as before>
}
```

pohly: I understand that this then does some work multiple times. But I prefer readability over performance here.

byako:
Actually, 0.28.0 is out, I'll bump go.mod to it right away.

go.mod
```diff
@@ -3,75 +3,76 @@ module sigs.k8s.io/dra-example-driver
...
-	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.27.1
-	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.27.1
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.27.1
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.27.1
-	k8s.io/kubectl => k8s.io/kubectl v0.27.1
-	k8s.io/kubelet => k8s.io/kubelet v0.27.1
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.27.1
-	k8s.io/metrics => k8s.io/metrics v0.27.1
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.27.1
+	k8s.io/api => k8s.io/api v0.28.0
```
pohly, line 6: The commit message needs to be updated.

pohly (changes requested):
Looks much simpler than before, just one nit.

This also contains some drive-by changes, but I guess those are fine.

go.mod
byako: Ah, yes, overlooked it, thanks!

pohly (approved):
/lgtm
/approve

PR #28, opened 2023-08-25
[golangci-lint][goimports] add local imports check

tariq1890:


tariq1890:
cc @pohly

pohly (commented):
I remember that you also had some code changes for import statements. Are those not needed to pass the reconfigured linter check?

tariq1890:
@pohly yes that is correct. I chose to exclude them this time as they are auto-generated code. I believe golangci-lint skips them by default. 

Ideally, kube builder should be generating goimports compliant code

pohly (approved):
/lgtm
/approve

PR #29, opened 2023-08-28
make .gitignore only ignore binaries

byako:
#26 broke .gitignore, it now ignores `cmd/<$binarry>/` source as well as ./$<binary>. This fixes the excessive ignore.


pohly (approved):
/lgtm
/approve

PR #30, opened 2023-08-29
Fix null dereference in nas-set-status

byako:
Currently when init container for kubelet plugin tries to start, it fails with SEGFAULT when it tries to add logging flags because the pointer in LoggingConfig is not initialized.

This fixes it.


elezar (approved):
LGTM

PR #35, opened 2023-09-19
Update demo scripts to run with kubernetes 1.28.0

klueska:
As part of this we remove the custom base image that previously allowed us to pull in containerd 1.7. The latest kind images are all running with containerd 1.7 by default.

### DIFF
--- a/demo/scripts/build-kind-image.sh
+++ b/demo/scripts/build-kind-image.sh
@@ -48,4 +48,4 @@ KIND_K8S_DIR="${TMP_DIR}/kubernetes-${KIND_K8S_TAG}"
 git clone --depth 1 --branch ${KIND_K8S_TAG} ${KIND_K8S_REPO} ${KIND_K8S_DIR}
 
 # Build the kind base image
-kind build node-image --base-image "${KIND_IMAGE_BASE}" --image "${KIND_IMAGE}" "${KIND_K8S_DIR}"
+kind build node-image --image "${KIND_IMAGE}" "${KIND_K8S_DIR}"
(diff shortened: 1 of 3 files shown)


pohly (approved):
/lgtm
/approve

Not tested manually! :hot_face:

demo/scripts/build-kind-image.sh
```diff
@@ -48,4 +48,4 @@ KIND_K8S_DIR="${TMP_DIR}/kubernetes-${KIND_K8S_TAG}"
 git clone --depth 1 --branch ${KIND_K8S_TAG} ${KIND_K8S_REPO} ${KIND_K8S_DIR}
 
 # Build the kind base image
-kind build node-image --base-image "${KIND_IMAGE_BASE}" --image "${KIND_IMAGE}" "${KIND_K8S_DIR}"
+kind build node-image --image "${KIND_IMAGE}" "${KIND_K8S_DIR}"
```
elezar, line 51: As a matter of interest, is supporting the `KIND_IMAGE_BASE` something we want to remove? Or are we not too concerned with this at this stage?

elezar: OK. That's fine. Looking at the code I see that there's still a way to build a custom k8s version in. I was looking at this since the images on Dockerhub seem to lag a little.

klueska: Yes, you can still build an image with an arbitrary k8s version in it (i.e. against a `node` image`). I just removed the code for building the `node` image itself against a custom `base` image.

Issue #10, opened 2023-05-02
implement logging best practices


pohly:
Currently the example driver uses logrus for logging. Logrus pioneered structured logging, but nowadays has been superseded by other, more modern solutions. Let's use something else.

pohly:
My proposal is to use the same logging technology as in Kubernetes:
- The main binary calls k8s.io/component-base/logs to set up logging - see https://github.com/kubernetes/component-base/tree/master/logs/example.
- Let's enable contextual logging by default, as in https://github.com/kubernetes-csi/external-provisioner/pull/639
- Everything else uses [contextual logging](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3077-contextual-logging).

byako:
+1 for contextual logging

pohly:
The code already uses component-base/logs...

pohly:
... but not in all binaries.

pohly:
Since f6acdbae37, all commands have these flags:
```
   Logging:

   --feature-gates value  A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
     AllAlpha=true|false (ALPHA - default=false)
     AllBeta=true|false (BETA - default=false)
     ContextualLogging=true|false (ALPHA - default=false)
     LoggingAlphaOptions=true|false (ALPHA - default=false)
     LoggingBetaOptions=true|false (BETA - default=true) (default: ContextualLogging=true) [$FEATURE_GATES]
   --log-flush-frequency value        Maximum number of seconds between log flushes (default: 5s) [$LOG_FLUSH_FREQUENCY]
   --log-json-info-buffer-size value  [Alpha] In JSON format with split output streams, the info messages can be buffered for a while to increase performance. The default value of zero bytes disables buffering. The size can be specified as number of bytes (512), multiples of 1000 (1K), multiples of 1024 (2Ki), or powers of those (3M, 4G, 5Mi, 6Gi). Enable the LoggingAlphaOptions feature gate to use this. (default: 0) [$LOG_JSON_INFO_BUFFER_SIZE]
   --log-json-split-stream value      [Alpha] In JSON format, write error messages to stderr and info messages to stdout. The default is to write a single stream to stdout. Enable the LoggingAlphaOptions feature gate to use this. (default: false) [$LOG_JSON_SPLIT_STREAM]
[...]

Issue #38, opened 2023-11-20
add E2E testing


pohly:
The README explains how to run the demo. We should have automated testing of PRs that ensures that those steps still work.

This needs a capable enough worker that supports kind. A Prow job can handle this. Not sure about GitHub actions. Prow might be better also because it integrates better with Tide.

How to test is open. It might just be a shell script that we run in the right environment.

nojnhuh:
I'll start looking into this one!

/assign

kannon92:
I just saw this.. @nojnhuh I opened up https://github.com/kubernetes-sigs/dra-example-driver/pull/56 to see if it would work with a github action. I was working on a make file to help streamline creation of the dra driver/kind.

nojnhuh:
Awesome! I'll re-assign this issue to you to keep track but I'll definitely take a look at the PR when it's ready. Thanks!

/unassign
/assign @kannon92

kannon92:
@pohly I think with https://github.com/kubernetes-sigs/dra-example-driver/pull/56, we can close this.

PR #39, opened 2023-11-20
fix and test for contextual logging

pohly:
The logcheck finds log calls which don't support contextual logging. It now runs as part of "make test".

Fixes: https://github.com/kubernetes-sigs/dra-example-driver/issues/10


Makefile
```diff
@@ -87,6 +87,13 @@ misspell:
 vet:
 	go vet $(MODULE)/...
 
+# Ensure that all log calls support contextual logging.
+test: logcheck
+.PHONY: logcheck
```
elezar, line 92: Question: Should we run this in the GitHub actions?

elezar (commented):
Thanks @pohly!

One minor question from my side.

Makefile
pohly: We aren't already running `test` for PRs?! We absolutely should.

I created https://github.com/kubernetes-sigs/dra-example-driver/issues/38 for E2E testing. Can you take care of doing something similar for unit testing?

elezar: I have created #41

elezar (approved):
Thanks for this @pohly.

/lgtm

PR #46, opened 2024-04-16
Update to go 1.22.2

klueska:
Bumping to go 1.22 also forced us to bump our controller-gen and client-gen versions.
This, in turn, caused some changes in the generated code that is included in this PR.


kannon92:
Looks like you just need to update github workloads to use 1.22 and this should be good.

byako (commented):
/lgtm

byako (approved):

PR #51, opened 2024-08-07
Add support for Podman

empovit:
Add podman as an option for building images and running on Kind

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DOCKER   ?= docker
+CONTAINER_TOOL ?= docker
 MKDIR    ?= mkdir
 TR       ?= tr
 DIST_DIR ?= $(CURDIR)/dist
@@ -118,7 +118,7 @@ generate-deepcopy: vendor
 .PHONY: .build-image
 .build-image: docker/Dockerfile.devel
 	if [ x"$(SKIP_IMAGE_BUILD)" = x"" ]; then \
-		$(DOCKER) build \
+		$(CONTAINER_TOOL) build \
 			--progress=plain \
 			--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
 			--tag $(BUILDIMAGE) \
@@ -126,29 +126,33 @@ generate-deepcopy: vendor
 			docker; \
 	fi
 
+ifeq ($(CONTAINER_TOOL),podman)
+CONTAINER_TOOL_OPTS=-v $(PWD):$(PWD):Z
+else
+CONTAINER_TOOL_OPTS=-v $(PWD):$(PWD) --user $$(id -u):$$(id -g)
+endif
+
 $(DOCKER_TARGETS): docker-%: .build-image
-	@echo "Running 'make $(*)' in docker container $(BUILDIMAGE)"
-	$(DOCKER) run \
+	@echo "Running 'make $(*)' in container $(BUILDIMAGE)"
+	$(CONTAINER_TOOL) run \
 		--rm \
 		-e HOME=$(PWD) \
 		-e GOCACHE=$(PWD)/.cache/go \
 		-e GOPATH=$(PWD)/.cache/gopath \
-		-v $(PWD):$(PWD) \
+		$(CONTAINER_TOOL_OPTS) \
 		-w $(PWD) \
-		--user $$(id -u):$$(id -g) \
 		$(BUILDIMAGE) \
 			make $(*)
 
 # Start an interactive shell using the development image.
 PHONY: .shell
 .shell:
-	$(DOCKER) run \
+	$(CONTAINER_TOOL) run \
 		--rm \
 		-ti \
 		-e HOME=$(PWD) \
 		-e GOCACHE=$(PWD)/.cache/go \
 		-e GOPATH=$(PWD)/.cache/gopath \
-		-v $(PWD):$(PWD) \
+		$(CONTAINER_TOOL_OPTS) \
 		-w $(PWD) \
-		--user $$(id -u):$$(id -g) \
 		$(BUILDIMAGE)
--- a/demo/scripts/common.sh
+++ b/demo/scripts/common.sh
@@ -22,7 +22,7 @@
 # A reference to the current directory where this script is located
 SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 
-# The name of the example driver 
+# The name of the example driver
 : ${DRIVER_NAME:=dra-example-driver}
 
 # The registry, image and tag for the example driver
@@ -55,3 +55,18 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 # The name of the kind image to build / run
 : ${KIND_IMAGE:="kindest/node:${KIND_K8S_TAG}"}
 
+# Container tool, e.g. docker/podman
+if [[ -z "${CONTAINER_TOOL}" ]]; then
+    if [[ -n "$(which docker)" ]]; then
+        echo "Docker found in PATH."
+        CONTAINER_TOOL=docker
+    elif [[ -n "$(which podman)" ]]; then
+        echo "Podman found in PATH."
+        CONTAINER_TOOL=podman
+    else
+        echo "No container tool detected. Please install Docker or Podman."
+        return 1
+    fi
+fi
+
+: ${KIND:="env KIND_EXPERIMENTAL_PROVIDER=${CONTAINER_TOOL} kind"}
--- a/demo/scripts/load-driver-image-into-kind.sh
+++ b/demo/scripts/load-driver-image-into-kind.sh
@@ -27,6 +27,10 @@ set -o pipefail
 
 source "${CURRENT_DIR}/common.sh"
 
-kind load docker-image \
+# Work around kind not loading image with podman
+IMAGE_ARCHIVE=driver_image.tar
+${CONTAINER_TOOL} save -o "${IMAGE_ARCHIVE}" "${DRIVER_IMAGE}" && \
+${KIND} load image-archive \
 	--name "${KIND_CLUSTER_NAME}" \
-	"${DRIVER_IMAGE}"
+	"${IMAGE_ARCHIVE}"
+rm "${IMAGE_ARCHIVE}"
--- a/deployments/container/Makefile
+++ b/deployments/container/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ##### Global variables #####
-DOCKER ?= docker
+CONTAINER_TOOL ?= docker
 MKDIR ?= mkdir
 DISTRIBUTIONS := ubuntu22.04
 DOCKERFILE := $(CURDIR)/deployments/container/Dockerfile
@@ -31,7 +31,7 @@ ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
 # Use a generic build target to build the relevant images
 $(DISTRIBUTIONS):
 	DOCKER_BUILDKIT=1 \
-		$(DOCKER) build --pull \
+		$(CONTAINER_TOOL) build --pull \
 		--tag $(IMAGE) \
 		--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
 		--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ vendor/
 [._]*.sw[a-p]
 
 ./dra-example-kubeletplugin
+
+driver_image.tar
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The procedure below has been tested and verified on both Linux and Mac.
 
 * [GNU Make 3.81+](https://www.gnu.org/software/make/)
 * [GNU Tar 1.34+](https://www.gnu.org/software/tar/)
-* [docker v20.10+ (including buildx)](https://docs.docker.com/engine/install/)
+* [docker v20.10+ (including buildx)](https://docs.docker.com/engine/install/) or [Podman v4.9+](https://podman.io/docs/installation)
 * [kind v0.17.0+](https://kind.sigs.k8s.io/docs/user/quick-start/)
 * [helm v3.7.0+](https://helm.sh/docs/intro/install/)
 * [kubectl v1.18+](https://kubernetes.io/docs/reference/kubectl/)
@@ -37,6 +37,11 @@ moment to browse through the various files and see what's available:
 git clone https://github.com/kubernetes-sigs/dra-example-driver.git
 cd dra-example-driver
 ```
+
+**Note**: The scripts will automatically use either `docker`, or `podman` as the container tool command, whichever
+can be found in the PATH. To override this behavior, set `CONTAINER_TOOL` environment variable either by calling
+`export CONTAINER_TOOL=docker`, or by prepending `CONTAINER_TOOL=docker` to a script
+(e.g. `CONTAINER_TOOL=docker ./path/to/script.sh`). Keep in mind that building Kind images currently requires Docker.
 
 From here we will build the image for the example resource driver:
 ```bash
--- a/demo/build-driver.sh
+++ b/demo/build-driver.sh
@@ -31,7 +31,7 @@ source "${CURRENT_DIR}/scripts/common.sh"
 ${SCRIPTS_DIR}/build-driver-image.sh
 
 # If a cluster is already running, load the image onto its nodes
-EXISTING_CLUSTER="$(kind get clusters | grep -w "${KIND_CLUSTER_NAME}" || true)"
+EXISTING_CLUSTER="$(${KIND} get clusters | grep -w "${KIND_CLUSTER_NAME}" || true)"
 if [ "${EXISTING_CLUSTER}" != "" ]; then
 	${SCRIPTS_DIR}/load-driver-image-into-kind.sh
 fi
--- a/demo/create-cluster.sh
+++ b/demo/create-cluster.sh
@@ -34,7 +34,7 @@ fi
 ${SCRIPTS_DIR}/create-kind-cluster.sh
 
 # If a driver image already exists load it into the cluster
-EXISTING_IMAGE_ID="$(docker images --filter "reference=${DRIVER_IMAGE}" -q)"
+EXISTING_IMAGE_ID="$(${CONTAINER_TOOL} images --filter "reference=${DRIVER_IMAGE}" -q)"
 if [ "${EXISTING_IMAGE_ID}" != "" ]; then
 	${SCRIPTS_DIR}/load-driver-image-into-kind.sh
 fi
--- a/demo/scripts/build-driver-image.sh
+++ b/demo/scripts/build-driver-image.sh
@@ -41,6 +41,7 @@ cd ${CURRENT_DIR}/../..
 export REGISTRY="${DRIVER_IMAGE_REGISTRY}"
 export IMAGE="${DRIVER_IMAGE_NAME}"
 export VERSION="${DRIVER_IMAGE_TAG}"
+export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
 # Regenerate the CRDs and build the container image
 make docker-generate
--- a/demo/scripts/build-kind-image.sh
+++ b/demo/scripts/build-kind-image.sh
@@ -28,9 +28,14 @@ set -o pipefail
 source "${CURRENT_DIR}/common.sh"
 
 # If an image ID already exists for the image we plan to build, we are done.
-EXISTING_IMAGE_ID="$(docker images --filter "reference=${KIND_IMAGE}" -q)"
+EXISTING_IMAGE_ID="$(${CONTAINER_TOOL} images --filter "reference=${KIND_IMAGE}" -q)"
 if [ "${EXISTING_IMAGE_ID}" != "" ]; then
 	exit 0
+fi
+
+if [[ "${CONTAINER_TOOL}" != "docker" ]]; then
+    echo "Building kind images requires Docker. Cannot use '${CONTAINER_TOOL}'"
+    exit 1
 fi
 
 # Create a temorary directory to hold all the artifacts we need for building the image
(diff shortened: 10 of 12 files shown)


deployments/container/Makefile
```diff
@@ -28,7 +28,9 @@ IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
 
 ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
 
-# Use a generic build target to build the relevant images
+ubi9: BASE_IMAGE = registry.access.redhat.com/ubi9/ubi-micro:latest
```
aojea, line 31: is not good to use latest as it can have random behavior, do you have a static tag?

empovit: Thanks, pinned to 9.4

demo/scripts/common.sh
```diff
@@ -52,3 +52,5 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 # The name of the kind image to build / run
 : ${KIND_IMAGE:="kindest/node:${KIND_K8S_TAG}"}
 
+# Container tool, e.g. docker/podman
+: ${DOCKER:="docker"}
```
elezar, line 56: Not critical for this PR. Is there a more generic term that we can use for `DOCKER`? (may affect more files, but it may be good to do as a follow up)

demo/scripts/load-driver-image-into-kind.sh
```diff
@@ -27,6 +27,16 @@ set -o pipefail
 
 source "${CURRENT_DIR}/common.sh"
 
-kind load docker-image \
-	--name "${KIND_CLUSTER_NAME}" \
-	"${DRIVER_IMAGE}"
+# Work around kind not loading image with podman
+if [[ "${DOCKER}" == "podman" ]]; then
+	IMAGE_ARCHIVE=driver_image.tar
```
elezar, line 32: should we use a file in `/tmp`.

Also in theory this workflow can be used for both runtimes. Should we just use the one instead of having a conditional block?

Makefile
```diff
@@ -146,16 +146,21 @@ generate-crds: vendor
 			docker; \
 	fi
 
+ifeq ($(DOCKER),podman)
+DOCKER_OPTS=-v $(PWD):$(PWD):Z
```
elezar, line 150: Do you happen to have a doc link to describe why this change is required?

deployments/container/Makefile
```diff
@@ -28,7 +28,9 @@ IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
 
 ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
 
-# Use a generic build target to build the relevant images
+ubi9: BASE_IMAGE = registry.access.redhat.com/ubi9/ubi-micro:9.4
```
elezar, line 31: This is not specific to using podman? Could we move this to a different PR. I would prefer to not have to maintain this in this repo and could look at allowing a user to set `BASE_IMAGE` instead to target other distributions.

The concern is keeping the tags up to date, for example.

elezar (commented):
Thanks @empovit

This looks good in general. I have some comments / questions, but I don't see any show-stoppers.

demo/scripts/common.sh
empovit: Yes, usually it's `CONTAINER_TOOL` or something along those lines. As long as you're OK with moving away from DOCKER :) Ideally all Dockerfiles and directory names should change as well.

elezar: I think I would be ok with it, but may require some more discussion on our end if we start renaming things outside of the environment and make variables.

What is the generic standard for `Dockerfile`? From my perspective, this is an implied API and not just a file name.

demo/scripts/load-driver-image-into-kind.sh
empovit: Normally the file will be removed right after loading into Kind, so I don't think it matters.

Correct, the flow can be used for both runtimes, but I couldn't test extensively on Docker so preferred to make as little changes as possible at the moment.

Makefile
empovit: I can't find it right now, but `--user` works differently with podman, and `:z` (`:Z`) is required to write to a host directory via a volume.

demo/scripts/common.sh
empovit: Sounds good.
`Containerfile` is more inclusive :)

elezar: To be clear, I don't think we should rename the `Dockerfile` to `Containerfile` at this stage. See https://github.com/docker/cli/issues/3251

empovit: I've replaced `docker` command with `CONTAINER_TOOL` variable.

empovit: It make the code look better than `if [[ "${DOCKER}" == "podman" ]]` :smile:

klueska:
@empovit what do you mean by:

> NOTE: Currently, the dra-example-driver-kubeletplugin pod crashes if Kind runs as non-root/non-sudo.

Is this true for both docker and podman or just podman? How to reproduce on docker?

Makefile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DOCKER   ?= docker
+CONTAINER_TOOL ?= docker
```
elezar, line 15: Do we want to add a note to the README on how to get started using podman?

empovit:
> @empovit what do you mean by:
> 
> > NOTE: Currently, the dra-example-driver-kubeletplugin pod crashes if Kind runs as non-root/non-sudo.
> 
> Is this true for both docker and podman or just podman? How to reproduce on docker?

@klueska I've described the issue in https://kubernetes.slack.com/archives/C0409NGC1TK/p1721720213996169

```console
$ kubectl logs -n dra-example-driver dra-example-driver-kubeletplugin-q5khp
Defaulted container "plugin" out of: plugin, init (init)
E0723 07:23:56.322197       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x18258e0, 0x2a0aed0})
        /build/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:75 +0x85
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc0000061c0?})
        /build/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:49 +0x6b
panic({0x18258e0?, 0x2a0aed0?})
        /usr/local/go/src/runtime/panic.go:770 +0x132
github.com/fsnotify/fsnotify.(*Watcher).isClosed(...)
        /build/vendor/github.com/fsnotify/fsnotify/backend_inotify.go:176
github.com/fsnotify/fsnotify.(*Watcher).Add(0x0, {0xc000048009?, 0xc00051b4c8?})
        /build/vendor/github.com/fsnotify/fsnotify/backend_inotify.go:240 +0x57
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi.(*watch).update(0xc0001112d0, 0xc0001d6900, {0x0, 0x0, 0x51b720?})
        /build/vendor/github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/cache.go:563 +0xd1
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi.(*Cache).refreshIfRequired(0xc000023180, 0x8?)
        /build/vendor/github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/cache.go:208 +0x38
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi.(*Cache).Refresh(0xc000023180)
        /build/vendor/github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/cache.go:123 +0xa6
main.NewCDIHandler(0x0?)
        /build/cmd/dra-example-kubeletplugin/cdi.go:46 +0x9e
main.NewDeviceState(0xc00051bbd0)
        /build/cmd/dra-example-kubeletplugin/state.go:66 +0x76
main.NewDriver.func1()
        /build/cmd/dra-example-kubeletplugin/driver.go:53 +0x85
k8s.io/client-go/util/retry.OnError.func1()
        /build/vendor/k8s.io/client-go/util/retry/util.go:51 +0x30
k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtection(0x4c9379?)
        /build/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:145 +0x3e
k8s.io/apimachinery/pkg/util/wait.ExponentialBackoff({0x989680, 0x3ff0000000000000, 0x3fb999999999999a, 0x5, 0x0}, 0xc00051b920)
        /build/vendor/k8s.io/apimachinery/pkg/util/wait/backoff.go:461 +0x5a
k8s.io/client-go/util/retry.OnError({0x989680, 0x3ff0000000000000, 0x3fb999999999999a, 0x5, 0x0}, 0x0?, 0xc0002ad6f8?)
        /build/vendor/k8s.io/client-go/util/retry/util.go:50 +0xa5
k8s.io/client-go/util/retry.RetryOnConflict(...)
        /build/vendor/k8s.io/client-go/util/retry/util.go:104
main.NewDriver({0x1d09b90, 0x2aa1e40}, 0xc00051bbd0)
        /build/cmd/dra-example-kubeletplugin/driver.go:42 +0x170
main.StartPlugin({0x1d09b90, 0x2aa1e40}, 0xc00051bbd0)
        /build/cmd/dra-example-kubeletplugin/main.go:139 +0x18b
main.newApp.func2(0xc0002abdc0?)
        /build/cmd/dra-example-kubeletplugin/main.go:113 +0x12e
github.com/urfave/cli/v2.(*Command).Run(0xc0003514a0, 0xc0002abdc0, {0xc000110070, 0x1, 0x1})
        /build/vendor/github.com/urfave/cli/v2/command.go:274 +0x93f
github.com/urfave/cli/v2.(*App).RunContext(0xc000032000, {0x1d09b90, 0x2aa1e40}, {0xc000110070, 0x1, 0x1})
        /build/vendor/github.com/urfave/cli/v2/app.go:332 +0x566
github.com/urfave/cli/v2.(*App).Run(...)
        /build/vendor/github.com/urfave/cli/v2/app.go:309
main.main()
        /build/cmd/dra-example-kubeletplugin/main.go:60 +0x3f
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x518e77]

goroutine 1 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc0000061c0?})
        /build/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:56 +0xcd
panic({0x18258e0?, 0x2a0aed0?})
        /usr/local/go/src/runtime/panic.go:770 +0x132
github.com/fsnotify/fsnotify.(*Watcher).isClosed(...)
        /build/vendor/github.com/fsnotify/fsnotify/backend_inotify.go:176
github.com/fsnotify/fsnotify.(*Watcher).Add(0x0, {0xc000048009?, 0xc00051b4c8?})
        /build/vendor/github.com/fsnotify/fsnotify/backend_inotify.go:240 +0x57
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi.(*watch).update(0xc0001112d0, 0xc0001d6900, {0x0, 0x0, 0x51b720?})
        /build/vendor/github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/cache.go:563 +0xd1
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi.(*Cache).refreshIfRequired(0xc000023180, 0x8?)
        /build/vendor/github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/cache.go:208 +0x38
github.com/container-orchestrated-devices/container-device-interface/pkg/cdi.(*Cache).Refresh(0xc000023180)
        /build/vendor/github.com/container-orchestrated-devices/container-device-interface/pkg/cdi/cache.go:123 +0xa6
main.NewCDIHandler(0x0?)
        /build/cmd/dra-example-kubeletplugin/cdi.go:46 +0x9e
main.NewDeviceState(0xc00051bbd0)
        /build/cmd/dra-example-kubeletplugin/state.go:66 +0x76
main.NewDriver.func1()
        /build/cmd/dra-example-kubeletplugin/driver.go:53 +0x85
k8s.io/client-go/util/retry.OnError.func1()
        /build/vendor/k8s.io/client-go/util/retry/util.go:51 +0x30
k8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtection(0x4c9379?)
        /build/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:145 +0x3e
k8s.io/apimachinery/pkg/util/wait.ExponentialBackoff({0x989680, 0x3ff0000000000000, 0x3fb999999999999a, 0x5, 0x0}, 0xc00051b920)
        /build/vendor/k8s.io/apimachinery/pkg/util/wait/backoff.go:461 +0x5a
k8s.io/client-go/util/retry.OnError({0x989680, 0x3ff0000000000000, 0x3fb999999999999a, 0x5, 0x0}, 0x0?, 0xc0002ad6f8?)
        /build/vendor/k8s.io/client-go/util/retry/util.go:50 +0xa5
k8s.io/client-go/util/retry.RetryOnConflict(...)
        /build/vendor/k8s.io/client-go/util/retry/util.go:104
main.NewDriver({0x1d09b90, 0x2aa1e40}, 0xc00051bbd0)
        /build/cmd/dra-example-kubeletplugin/driver.go:42 +0x170
main.StartPlugin({0x1d09b90, 0x2aa1e40}, 0xc00051bbd0)
        /build/cmd/dra-example-kubeletplugin/main.go:139 +0x18b
main.newApp.func2(0xc0002abdc0?)
        /build/cmd/dra-example-kubeletplugin/main.go:113 +0x12e
github.com/urfave/cli/v2.(*Command).Run(0xc0003514a0, 0xc0002abdc0, {0xc000110070, 0x1, 0x1})
        /build/vendor/github.com/urfave/cli/v2/command.go:274 +0x93f
github.com/urfave/cli/v2.(*App).RunContext(0xc000032000, {0x1d09b90, 0x2aa1e40}, {0xc000110070, 0x1, 0x1})
        /build/vendor/github.com/urfave/cli/v2/app.go:332 +0x566
github.com/urfave/cli/v2.(*App).Run(...)
        /build/vendor/github.com/urfave/cli/v2/app.go:309
main.main()
        /build/cmd/dra-example-kubeletplugin/main.go:60 +0x3f
```
I didn't test with rootless Docker as it's not common to my knowledge. IMO the problem would be best dealt with as a GitHub issue, but can't open one without adding basic support for Podman first.

Tested combinations
* Works with Docker on Ubuntu as regular user
* Works with Podman on RHEL as root user
* Works with Podman on Fedora using sudo
* Crashes with Podman on Fedora as regular user

Makefile
empovit: Yes, that's a good idea. I will update the PR.

elezar:
@empovit I think that's a bug in the example driver, please create an issue for it and I will have a look at it on Monday. I think we need to update the way in which we write the CDI specifications.

klueska:
https://github.com/kubernetes-sigs/dra-example-driver/pull/52 has been merged. Please let me know if it has resolved the issue.

empovit:
Hmm, #52 fixes the issue on 1.30, but some other changes break the Podman support :( 
This can't run because Kind's build has Docker hard-coded:

```console
KUBE_GIT_VERSION=v1.30.0 BUILD_KIND_IMAGE=true KIND_K8S_TAG=v1.31.0-rc.1 ./demo/create-cluster.sh
```

elezar:
Glad to hear that the repo works as-is at least. Let's dig into the build failures and decide whether we can address them here or whether we need to propose changes to kind.

klueska:
So the issue is when building the image itself -- maybe I can / should change things to just have it pull from an image hosted in the github registry until an official kind image comes out.

empovit:
@elezar my understanding is that it's Kind/K8s policy to use only Docker, so proposing changes to Kind won't work. 

@klueska I think that would be great. Alternatively, we can mention as a limitation that Podman works only with 1.30 until a 1.31 image is available.

klueska:
Actually -- it looks like they already published a 1.31.0 image yesterday. Let me just open a PR to update everything to that now.

klueska:
Done, please review:
https://github.com/kubernetes-sigs/dra-example-driver/pull/53

klueska:
I don't see anything in the PR that tells `kind` itself to use `podman`.

I expected to see something like `KIND_EXPERIMENTAL_PROVIDER=podman` in the scripts somehwere...

demo/scripts/load-driver-image-into-kind.sh
```diff
@@ -27,6 +27,16 @@ set -o pipefail
 
 source "${CURRENT_DIR}/common.sh"
 
-kind load docker-image \
-	--name "${KIND_CLUSTER_NAME}" \
-	"${DRIVER_IMAGE}"
+# Work around kind not loading image with podman
+if [[ "${CONTAINER_TOOL}" == "podman" ]]; then
+	IMAGE_ARCHIVE=driver_image.tar
```
klueska, line 32: Can you add this to the .gitignore

empovit:
> I don't see anything in the PR that tells `kind` itself to use `podman`.
> 
> I expected to see something like `KIND_EXPERIMENTAL_PROVIDER=podman` in the scripts somehwere...

It's auto-detected https://kind.sigs.k8s.io/docs/user/quick-start/
> The kind can auto-detect the [docker](https://www.docker.com/), [podman](https://podman.io/), or [nerdctl](https://github.com/containerd/nerdctl) installed and choose the available one. If you want to turn off the auto-detect, use the environment variable KIND_EXPERIMENTAL_PROVIDER=docker, KIND_EXPERIMENTAL_PROVIDER=podman or KIND_EXPERIMENTAL_PROVIDER=nerdctl to select the runtime.

klueska:
OK, so I can override it with `KIND_EXPERIMENTAL_PROVIDER` if I want, but otherwise it will be autodetected?

demo/scripts/load-driver-image-into-kind.sh
empovit: Thanks, done.

empovit:
Correct.

klueska:
Can we instead explicitly set KIND_EXPERIMENTAL_PROVIDER to the value of CONTAINER_TOOL and instead autodetect CONTAINER_TOOL in this environment?

empovit:
Done. PTAL

demo/scripts/load-driver-image-into-kind.sh
```diff
@@ -27,6 +27,16 @@ set -o pipefail
...
+	${CONTAINER_TOOL} save -o "${IMAGE_ARCHIVE}" "${DRIVER_IMAGE}" && \
+	kind load image-archive \
+		--name "${KIND_CLUSTER_NAME}" \
+		"${IMAGE_ARCHIVE}"
+	rm "${IMAGE_ARCHIVE}"
+else
+	kind load docker-image \
+		--name "${KIND_CLUSTER_NAME}" \
+		"${DRIVER_IMAGE}"
+fi
```
klueska, line 42: Do these also need `KIND_EXPERIMENTAL_PROVIDER=${CONTAINER_TOOL} ` in front of them?

klueska: While we're at it, let's just drop the if/else and do it the same for both podman and docker (i.e. the new podman way).

klueska: Also, mabey the better way to do this is to introduce a `KIND` variable instead of calling `kind` directly. Then in common.sh we define `KIND="KIND_EXPERIMENTAL_PROVIDER=${CONTAINER_TOOL} kind"` and replace all direct calls to `kind` with `${KIND}`.

empovit: > While we're at it, let's just drop the if/else and do it the same for both podman and docker (i.e. the new podman way).

I can't currently test it. See my earlier  response.

empovit: > Do these also need `KIND_EXPERIMENTAL_PROVIDER=${CONTAINER_TOOL} ` in front of them?

According to the docs, it only works for `kind create cluster` (looking at the code seems to confirm that), will be ignored with all other commands. We still can implement your suggestion about `${KIND}` variable, hopefully this won't cause confusion.

empovit: Ok, I've made the requested changes, also tested on Ubuntu+Docker. PTAL

demo/scripts/common.sh
```diff
@@ -55,3 +55,18 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
+    elif [[ -n "$(which podman)" ]]; then
+        echo "Podman found in PATH."
+        CONTAINER_TOOL=podman
+    else
+        echo "No container tool detected. Please install Docker or Podman."
+        return 1
+    fi
+fi
+
+: ${KIND:="env KIND_EXPERIMENTAL_PROVIDER=${CONTAINER_TOOL} kind"}
```
klueska, line 72: just for my own understanding -- why is the `env` included here?

empovit: Right, without it substituting `${KIND}` makes the shell treat `KIND_EXPERIMENTAL_PROVIDER=podman` as a command, causing something like `.../demo/scripts/create-kind-cluster.sh: line 30: KIND_EXPERIMENTAL_PROVIDER=podman: command not found`.

If you can think of a better solution, I'll be happy to use it.

klueska: I guess the alternative would be to make sure that `KIND_EXPERIMENTAL_PROVIDER` is exported in every script where `kind` is called (and then go back to not having a `KIND` variable at all).

klueska: To be clear, I don't feel too strongly either way so long as it works (and will work in all supported environments).

empovit: I personally see no harm in using `env` as its purpose is exactly running a command with passed env variables https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html. It's also widely supported https://en.wikipedia.org/wiki/Env. On the other hand, it allows us to set the `KIND` variable in a single place, and keep it alongside other Kind-related variables.

klueska:
@empovit thanks for all the iterations on this. I've tested it on my side with both podman and docker and everything seems to work as expected.

/lgtm
/approve

PR #52, opened 2024-08-09
Update CDI specification to v0.8.0

elezar:
These changes update the imported CDI specification to v0.8.0. This includes the migration to the `tags.cncf.io` repository.


klueska:
I should have probably tested this before merging it -- after this change CDI injection is not working...

PR #54, opened 2024-08-22
Add support for opaque configs with examples

klueska:


README.md
```diff
@@ -254,12 +255,16 @@ below:
...
 ```bash
-for example in $(seq 1 4); do \
+for example in $(seq 1 5); do \
   echo "gpu-test${example}:"
   for pod in $(kubectl get pod -n gpu-test${example} --output=jsonpath='{.items[*].metadata.name}'); do \
     for ctr in $(kubectl get pod -n gpu-test${example} ${pod} -o jsonpath='{.spec.containers[*].name}'); do \
       echo "${pod} ${ctr}:"
-      kubectl logs -n gpu-test${example} ${pod} -c ${ctr}| grep GPU_DEVICE
+      if [ "${example}" -lt 3 ]; then
+        kubectl logs -n gpu-test${example} ${pod} -c ${ctr}| grep GPU_DEVICE
```
elezar, line 264: Does `GPU_` not work here too? Or are there other log entries that start `GPU_`?

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -89,18 +91,37 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices PreparedDevi
...
+		if device.Config.Sharing.IsTimeSlicing() {
+			tsconfig, err := device.Config.Sharing.GetTimeSlicingConfig()
+			if err != nil {
+				return fmt.Errorf("unable to get time slicing config for device %v: %v", device.DeviceName, err)
+			}
+			envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.TimeSlicingStrategy))
+			envs = append(envs, fmt.Sprintf("GPU_TIMESLICE_DURATION=%v", tsconfig.Duration))
+		}
+
+		if device.Config.Sharing.IsSpacePartitioning() {
```
elezar, line 108: nit: Can timeslicing and space partitioning be specified at the same time?

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -89,18 +91,37 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices PreparedDevi
 		Devices: []cdispec.Device{},
 	}
 
-	gpuIdx := 0
-	for _, device := range devices {
+	for i, device := range devices {
```
elezar, line 94: From these changes it seems as if the `GPU_` prefix should catch all relevant output.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -157,11 +167,21 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
-			RequestNames: []string{result.Request},
-			PoolName:     result.Pool,
-			DeviceName:   result.Device,
-			CDIDeviceIDs: s.cdi.GetClaimDevices([]string{result.Device}),
+		config, err := GetMergedOpaqueDeviceConfig[configapi.GpuConfig](result.Request, claim, classes)
+		if err != nil {
+			return nil, fmt.Errorf("error getting GPU configs: %v", err)
+		}
+
+		// Apply any requested configuration here
```
elezar, line 175: If this comment doesn't belong to the block below, then we should expand it and provide more context. Otherwise it just looks as if we have a unnecessary blank line here.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+
+func (pds PreparedDevices) GetDevices() []*drapbv1.Device {
+	var devices []*drapbv1.Device
+	for _, pd := range pds {
+		devices = append(devices, &pd.Device)
+	}
+	return devices
+}
+
+type PtrHasGroupVersionKind[T any] interface {
```
elezar, line 205: Question: Must this be public?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	*T
+	GroupVersionKind() schema.GroupVersionKind
+}
+
+// GetMergedOpaqueDeviceConfig merges configs across device classes and resource claims for a given device request.
+func GetMergedOpaqueDeviceConfig[T any, TI PtrHasGroupVersionKind[T]](request string, claim *resourceapi.ResourceClaim, classes map[string]*resourceapi.DeviceClass) (T, error) {
+	// Collect all configs in reverse priority order. Device class configs have
```
elezar, line 212: Would calling out the priority of the comments be clearer if called out in a bulleted list. Would this information also be relevant for the function documentation?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	return devices
+}
+
+type PtrHasGroupVersionKind[T any] interface {
+	*T
+	GroupVersionKind() schema.GroupVersionKind
+}
+
+// GetMergedOpaqueDeviceConfig merges configs across device classes and resource claims for a given device request.
+func GetMergedOpaqueDeviceConfig[T any, TI PtrHasGroupVersionKind[T]](request string, claim *resourceapi.ResourceClaim, classes map[string]*resourceapi.DeviceClass) (T, error) {
```
elezar, line 211: Should `classes` be `deviceClasses` to make the code more readable below?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+		allConfigs = append(allConfigs, config.DeviceConfiguration)
+	}
+	for _, config := range claim.Spec.Devices.Config {
+		if !slices.Contains(config.Requests, request) {
+			continue
+		}
+		allConfigs = append(allConfigs, config.DeviceConfiguration)
+	}
+
+	// Merge all configs from the allConfigs slice in the order they were
```
elezar, line 227: It might be useful for testing to factor this out into a function.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	// added. Fields in a config that is unmarshaled earlier will be
+	// overwritten by fields in a config that is unmarhaled later.
+	var mergedConfig T
+	for _, config := range allConfigs {
+		if config.Opaque == nil {
+			continue
+		}
+		if config.Opaque.Driver != DriverName {
+			continue
+		}
```
elezar, line 237: Should filtering be a separate operation to make it clearer that it's being applied? Especially since we're using a global variable for filtering here.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+
+		if err := json.Unmarshal(config.Opaque.Parameters.Raw, &mergedConfig); err != nil {
+			return *new(T), fmt.Errorf("error converting config parameters to GpuConfig: %v", err)
```
elezar, line 240: Is `GpuConfig` valid here since we're unmarshalling into a templated type?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +193,67 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+			return *new(T), fmt.Errorf("wrong group for config parameters: %v", gvk.Group)
+		}
+
+		if gvk.Version != configapi.Version {
+			return *new(T), fmt.Errorf("wrong version for config parameters: %v", gvk.Version)
+		}
+
+		if gvk.Kind != configapi.GpuConfigKind {
+			return *new(T), fmt.Errorf("wrong kind for config parameters: %v", gvk.Kind)
+		}
```
elezar, line 255: It's not clear under which conditions these errors would occur. Factoring out into a validate function may also make sense.

demo/gpu-test4.yaml
```diff
@@ -38,7 +35,28 @@ spec:
...
+  namespace: gpu-test4
+  name: pod1
+  labels:
+    app: pod
+spec:
+  containers:
+  - name: ctr0
+    image: ubuntu:22.04
+    command: ["bash", "-c"]
+    args: ["export; sleep 9999"]
```
elezar, line 56: minor: Is there a pause container we could use instead?

elezar (commented):
Thanks @klueska. Some minor questions.

README.md
klueska: It does work, but I didn't want to output the sharing information on examples that didn't have anything to do with sharing.

cmd/dra-example-kubeletplugin/cdi.go
klueska: Not in the way I put together this example driver. One must pick a strategy (either TimeSlicing or SpacePartitioning) and depending on what they picked, only one of these will be valid.

cmd/dra-example-kubeletplugin/cdi.go
klueska: I wanted to avoid printing the sharing information for the first 2 examples.

cmd/dra-example-kubeletplugin/state.go
klueska: Added some more context.

cmd/dra-example-kubeletplugin/state.go
klueska: No, changed.

cmd/dra-example-kubeletplugin/state.go
klueska: I've updated the comment (and made it part of the godoc). Let me know what you think.

The goal is to make this generic enough that it becomes part of the in-tree helpers that ship with k8s itself.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated

cmd/dra-example-kubeletplugin/state.go
klueska: I haven't done this yet, but if you feel strongly I can.

cmd/dra-example-kubeletplugin/state.go
klueska: I've made it a parameter so that the function is more generic.

cmd/dra-example-kubeletplugin/state.go
klueska: I have removed this from the error string

cmd/dra-example-kubeletplugin/state.go
klueska: The function signature has changed to rely on `runtime.Scheme` to do validation for us. Take a look again and let me know if you still have questions.

demo/gpu-test4.yaml
klueska: Maybe -- we can look into it as a follow-up.

cmd/dra-example-kubeletplugin/cdi.go
elezar: But one doesn't seem to set the envvars for those example, but maybe I'm missing something from the code.

cmd/dra-example-kubeletplugin/state.go
elezar: It's not critical to this PR.

demo/gpu-test4.yaml
elezar: Yes, was more of a passing note.

README.md
```diff
@@ -225,10 +225,10 @@ metadata:
 ```
 
 Next, deploy four example apps that demonstrate how `ResourceClaim`s,
-`ResourceClaimTemplate`s, and custom `ClaimParameter` objects can be used to
-request access to resources in various ways:
+`ResourceClaimTemplate`s, and custom `GpuConfig` objects can be used to
```
elezar, line 228: This is more of a general comment, but since the resource driver is not actually GPU-specific we may want to drop the `GPU*` nomenclature in some follow-up. Not a blocker here.

README.md
elezar: Ah, ok. I see that timeslicing is enabled regardless.

cmd/dra-example-kubeletplugin/cdi.go
elezar: Does an `if -- else if` (or even a `switch` block) make more sense then to make the intent clearer?

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -89,18 +91,37 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices PreparedDevi
...
+			envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.TimeSlicingStrategy))
+			envs = append(envs, fmt.Sprintf("GPU_TIMESLICE_DURATION=%v", tsconfig.Duration))
+		}
+
+		if device.Config.Sharing.IsSpacePartitioning() {
+			spconfig, err := device.Config.Sharing.GetSpacePartitioningConfig()
+			if err != nil {
+				return fmt.Errorf("unable to get space partitioning config for device %v: %v", device.DeviceName, err)
+			}
+			envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.SpacePartitioningStrategy))
```
elezar, line 113: why not just use `config.Sharing.Strategy` directly in both cases?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -82,7 +92,7 @@ func NewDeviceState(config *Config) (*DeviceState, error) {
 	return state, nil
 }
 
-func (s *DeviceState) Prepare(claim *resourceapi.ResourceClaim) ([]*drapbv1.Device, error) {
+func (s *DeviceState) Prepare(claim *resourceapi.ResourceClaim, classes map[string]*resourceapi.DeviceClass) ([]*drapbv1.Device, error) {
```
elezar, line 95:
```suggestion
func (s *DeviceState) Prepare(claim *resourceapi.ResourceClaim, deviceClasses map[string]*resourceapi.DeviceClass) ([]*drapbv1.Device, error) {
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -146,7 +156,7 @@ func (s *DeviceState) Unprepare(claimUID string) error {
 	return nil
 }
 
-func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (PreparedDevices, error) {
+func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim, classes map[string]*resourceapi.DeviceClass) (PreparedDevices, error) {
```
elezar, line 159:
```suggestion
func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim, deviceClasses map[string]*resourceapi.DeviceClass) (PreparedDevices, error) {
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -157,11 +167,31 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+		)
+		if err != nil {
+			return nil, fmt.Errorf("error getting GPU configs: %v", err)
+		}
+
+		// Apply any requested configuration here.
+		//
+		// In this example driver there is nothing to do at this point, but a
+		// real driver would likely need to do some sort of hardware
+		// configuration here, based on the config that has been passed in.
```
elezar, line 185: This is clearer than the previous single line comment. Does it make sense to have an empty stub function that we call here instead? We could then properly document this stub to make it clear that this is something that should be extended in real-world cases.

README.md
klueska: Yes, it is the default.

cmd/dra-example-kubeletplugin/cdi.go
klueska: Sure. I just really dislike if/else blocks (for aesthetic reasons) which is why I didn't do it.

klueska: I just find this really hard to read:
```
        if device.Config.Sharing.IsTimeSlicing() {
            tsconfig, err := device.Config.Sharing.GetTimeSlicingConfig()
            if err != nil {
                return fmt.Errorf("unable to get time slicing config for device %v: %v", device.DeviceName, err)
            }
            envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.TimeSlicingStrategy))
            envs = append(envs, fmt.Sprintf("GPU_TIMESLICE_DURATION=%v", tsconfig.Duration))
        } else if device.Config.Sharing.IsSpacePartitioning() {
            spconfig, err := device.Config.Sharing.GetSpacePartitioningConfig()
            if err != nil {
                return fmt.Errorf("unable to get space partitioning config for device %v: %v", device.DeviceName, err)
            }
            envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.SpacePartitioningStrategy))
            envs = append(envs, fmt.Sprintf("GPU_PARTITION_COUNT=%v", spconfig.PartitionCount))
        }
```

How could this be done as a switch? Maybe that will look better.

cmd/dra-example-kubeletplugin/cdi.go
klueska: TimeSlicing is always on as the default (as mimiced from NVIDIA GPUs)

README.md
klueska: What do you mean? We agreed on writing an example GPU driver from the manufacturer "example.com". Everything about the driver is prefaced on this assumption.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated in latest

cmd/dra-example-kubeletplugin/state.go
klueska: Updated in latest.

cmd/dra-example-kubeletplugin/state.go
klueska: Will add in next iteration.

cmd/dra-example-kubeletplugin/cdi.go
klueska: Will do in next iteration.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +203,81 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+// Such configs can either be attached to the resource claim itself or to the
+// device class associated with the request. Configs found directly in the
+// resource claim take precedence over configs found in the device class.
+// Moreover, configs found earlier in the list of configs attached to either a
+// resource claim or a device class take precedence over configs found later in
+// their corresponding list.
+//
+// When merging, a config with higher precedence will overwrite fields of a
+// config with lower precedence. The only exception being that zero-valued
+// fields of higher precedence configs will *not* overwrite non-zero valued
```
elezar, line 232: If zero-valued fields are not overwritten, does a user have an option to clear fields? Also, is there a distinction between zero and `nil` in this case?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +203,81 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	decoder runtime.Decoder,
+	driverName string,
+	request string,
+	resourceClaim *resourceapi.ResourceClaim,
+	deviceClasses map[string]*resourceapi.DeviceClass,
+) (T, error) {
+	// Collect all configs in order of reverse precedence. Device class configs
+	// have lower precedence than resource claim configs. Within either a
+	// device class or a resource claim, configs earlier in the list have lower
+	// precedence than configs later in the list.
```
elezar, line 248: Given the docstring above, I think we can just use:
```suggestion
	// Collect all configs in order of reverse precedence.
```

cmd/dra-example-kubeletplugin/cdi.go
elezar:
```golang
		switch {
		case device.Config.Sharing.IsTimeSlicing():
			tsconfig, err := device.Config.Sharing.GetTimeSlicingConfig()
			if err != nil {
				return fmt.Errorf("unable to get time slicing config for device %v: %v", device.DeviceName, err)
			}
			envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.TimeSlicingStrategy))
			envs = append(envs, fmt.Sprintf("GPU_TIMESLICE_DURATION=%v", tsconfig.Duration))
		case device.Config.Sharing.IsSpacePartitioning():
			spconfig, err := device.Config.Sharing.GetSpacePartitioningConfig()
			if err != nil {
				return fmt.Errorf("unable to get space partitioning config for device %v: %v", device.DeviceName, err)
			}
			envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", configapi.SpacePartitioningStrategy))
			envs = append(envs, fmt.Sprintf("GPU_PARTITION_COUNT=%v", spconfig.PartitionCount))
		}
```

klueska: Updated.

cmd/dra-example-kubeletplugin/cdi.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: Not explicitly. The object will need to be defined with pointer that get populated with actual values when set to support this.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,121 @@
...
+	// SliceCount indicates how many equally sized (memory and compute) slices
+	// the GPU should be divided into. Each client that attaches will get
+	// access to exactly one of these slices.
+	PartitionCount int `json:"partitionCount,omitempty"`
+}
+
+// IsTimeSlicing checks if the TimeSlicing strategy is applied.
+func (s *GpuSharing) IsTimeSlicing() bool {
+	if s == nil {
+		return false
```
elezar, line 82: I thought we said that timeslicing was the default. This seems to indicate that this is not the case?

klueska: I decided to change it for the purpose of this function. The default gets set elsewhere explicitly (such that this would error if it wasn't for some reason).

klueska: https://github.com/kubernetes-sigs/dra-example-driver/pull/54/files#diff-e8cc6726d33b90acdb66a8ebbc67662470c2f60da033553fc14d5dca412a98a5R181

elezar (approved):
I have some thoughts about how the default sharing config is handled, but these are not relevant for movign froward on this PR.

LGTM

api/example.com/resource/gpu/v1alpha1/api.go
```diff
@@ -17,32 +17,59 @@
...
+			Duration: "Default",
 		},
 	}
 }
 
-func DefaultGpuClaimParametersSpec() *GpuClaimParametersSpec {
-	return &GpuClaimParametersSpec{
-		Count: ptr.To(1),
+func init() {
+	// Create a new scheme and add our types to it.
```
pohly, line 55:
```suggestion
	// Create a new scheme and add our types to it.
	// If at some point in the future a new version
	// of the configuration API becomes necessary,
	// then conversion functions can be generated
	// and registered to continue supporting older
	// versions.
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +210,89 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	for _, config := range resourceClaim.Spec.Devices.Config {
+		if !slices.Contains(config.Requests, request) {
+			continue
+		}
+		allConfigs = append(allConfigs, config.DeviceConfiguration)
+	}
+
+	// Merge all configs from the allConfigs slice in the order they were
+	// added. Fields in a config that is decoded earlier will be
+	// overwritten by (non-zero) fields in a config that is decoded later.
```
pohly, line 277: I am a bit worried here that this enters "undefined behavior" territory.

The current "encoding/json" de marshaling might work like that, but I am not sure whether that is guaranteed and whether the Kubernetes apimachinery makes similar guarantees.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -157,11 +167,38 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+			claim,
+			deviceClasses,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("error getting GPU configs: %v", err)
+		}
+
+		if config.Sharing == nil {
+			config.Sharing = configapi.DefaultGpuSharingConfig()
+		}
```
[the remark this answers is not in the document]
pohly: We might be able to generate defaulting code and register that in the schema instead?

The Kubernetes annotation for this is `+default:<some value>`, where `<some value>` can be `configapi.DefaultGpuSharingConfig` if its a constant, but not when it's a function.

Above you have `+kubebuilder:default` - perhaps that also works.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -157,11 +167,38 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+			Device: drapbv1.Device{
+				RequestNames: []string{result.Request},
+				PoolName:     result.Pool,
+				DeviceName:   result.Device,
+				CDIDeviceIDs: s.cdi.GetClaimDevices([]string{result.Device}),
+			},
+			Config: config,
+		}
+
+		// Apply any requested configuration here.
```
pohly, line 195: What about validation?

You have used `+kubebuilder:validation` above, but that doesn't seem to be used for anything.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,121 @@
...
+// +k8s:deepcopy-gen=false
+type Interface interface {
+	IsTimeSlicing() bool
+	IsSpacePartitioning() bool
+	GetTimeSlicingConfig() (*TimeSlicingConfig, error)
+	GetSpacePartitioningConfig() (*SpacePartitioningConfig, error)
+}
+
+// GpuSharingStrategy encodes the valid Sharing strategies as a string.
+// +kubebuilder:validation:Enum=TimeSlicing;SpacePartitioning
```
pohly, line 48: When I first saw the `+kubebuilder` comment, I expected it to have some effect later on. It doesn't.

Therefore I prefer to remove those comments.

api/example.com/resource/gpu/v1alpha1/api.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: This has now been changed to "replace" rather than "merge" semantics in the newest version.

cmd/dra-example-kubeletplugin/state.go
klueska: Manual validation code has now been added.

api/example.com/resource/gpu/v1alpha1/sharing.go
klueska: They are gone in the latest revision.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,121 @@
...
+
+// TimeSliceDuration encodes the valid timeslice duration as a string.
+// +kubebuilder:validation:Enum=Default;Short;Medium;Long
+type TimeSliceDuration string
+
+// GpuSharing holds the current sharing strategy for GPUs and its settings.
+// +kubebuilder:validation:MaxProperties=2
+type GpuSharing struct {
+	// +kubebuilder:default=TimeSlicing
+	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
klueska: Gone.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,115 @@
...
+	DefaultTimeSlice TimeSliceDuration = "Default"
+	ShortTimeSlice   TimeSliceDuration = "Short"
+	MediumTimeSlice  TimeSliceDuration = "Medium"
+	LongTimeSlice    TimeSliceDuration = "Long"
+)
+
+// Sharing provides methods to check if a given sharing strategy is selected
+// and grab its configuration.
+// +k8s:deepcopy-gen=false
+type Interface interface {
```
pohly, line 40: Why is this an interface? As it stands now, I have to find the implementation to know what the semantic of each method is.

Remove it perhaps?

api/example.com/resource/gpu/v1alpha1/validate.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		return s.SpacePartitioningConfig.Validate()
+	}
+	return fmt.Errorf("invalid GPU sharing settings: %v", s)
+}
+
+// Validate ensures that GpuConfig has a valid set of values.
+func (c *GpuConfig) Validate() error {
+	if c.Sharing == nil {
+		c.Sharing = &GpuSharing{
```
pohly, line 81: Mixing defaulting and validation looks odd to me.

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -89,18 +89,39 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices PreparedDevi
...
-	gpuIdx := 0
-	for _, device := range devices {
+	for i, device := range devices {
+		envs := []string{
+			fmt.Sprintf("GPU_DEVICE_%d=%s", i, device.DeviceName),
+		}
+
+		if device.Config.Sharing != nil {
+			envs = append(envs, fmt.Sprintf("GPU_SHARING_STRATEGY=%s", device.Config.Sharing.Strategy))
+		}
```
pohly, line 99: Does it make sense when different devices configure different strategies?

The variable can only be set to one value, so perhaps not?

This is of course just a hypothetical example, but it should still handle corner cases in a sane way.

pohly: Same below.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -97,7 +98,21 @@ func (d *driver) nodePrepareResource(ctx context.Context, claim *drapbv1.Claim)
...
-	prepared, err := d.state.Prepare(resourceClaim)
+	deviceClasses := make(map[string]*resourceapi.DeviceClass)
+	for _, request := range resourceClaim.Spec.Devices.Requests {
+		deviceClass, err := d.client.ResourceV1alpha3().DeviceClasses().Get(
+			ctx,
+			request.DeviceClassName,
+			metav1.GetOptions{})
+		if err != nil {
+			return &drapbv1.NodePrepareResourceResponse{
+				Error: fmt.Sprintf("failed to fetch DeviceClass %s", request.DeviceClassName),
```
pohly, line 109: Device classes may get deleted after allocation is complete. A driver should only depend on the claim status, but not on the device classes.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +206,97 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	defaultConfig T,
+	decoder runtime.Decoder,
+	driverName string,
+	request string,
+	resourceClaim *resourceapi.ResourceClaim,
+	deviceClasses map[string]*resourceapi.DeviceClass,
+) (T, error) {
+	// Collect all configs in order of reverse precedence.
+	var allConfigs []resourceapi.DeviceConfiguration
+	for _, config := range deviceClasses[request].Spec.Config {
```
pohly, line 256: This function should use the claim.status, not the claim.spec. It also shouldn't use DeviceClasses at all.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +206,97 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+
+		config, ok := c.(TI)
+		if !ok {
+			return *new(T), fmt.Errorf("error asserting actual runtime object type: %T", c)
+		}
+
+		candidateConfig = *config
+	}
+
+	// Validate the config before returning it.
```
pohly, line 296: Can't the caller do that? Then ptrImplementsRuntimeObject becomes simpler. The caller can also validate that there are no conflicting configs when multiple devices were allocated.

api/example.com/resource/gpu/v1alpha1/sharing.go
klueska: This was relevant in the NVIDIA driver where this was pulled from. I don't remember the context though and would need to look back at that code base to understand why I did it this way. In any case, I agree, it is irrelevant here. Removing.

api/example.com/resource/gpu/v1alpha1/validate.go
klueska: I have split them for the next revision.

cmd/dra-example-kubeletplugin/cdi.go
klueska: Are you suggesting to have one variable per device since different devices might have different sharing strategiesß Yes, that makes sense. I will make the change.

cmd/dra-example-kubeletplugin/state.go
klueska: Hmm. So are all the configs pre-merged into a single list in the claim status by the scheduler? Are they merged according to the semantics here (i.e. claim configs come before class configs so that their precedence is preserved)?

cmd/dra-example-kubeletplugin/driver.go
klueska: Will update in nexgt revision based on answer to https://github.com/kubernetes-sigs/dra-example-driver/pull/54#discussion_r1732789769

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/driver.go
klueska: Updated.

cmd/dra-example-kubeletplugin/cdi.go
klueska: Updated.

api/example.com/resource/gpu/v1alpha1/validate.go
klueska: Updated.

api/example.com/resource/gpu/v1alpha1/sharing.go
klueska: Updated.

api/example.com/resource/gpu/v1alpha1/api.go
```diff
@@ -17,32 +17,91 @@
...
 	}
 }
 
-func DefaultGpuClaimParametersSpec() *GpuClaimParametersSpec {
-	return &GpuClaimParametersSpec{
-		Count: ptr.To(1),
+// Normalize updates a GpuConfig config with implied default values based on other settings.
+func (c *GpuConfig) Normalize() {
+	if c == nil {
+		*c = DefaultGpuConfig()
```
pohly, line 64: `c` is nil, so this will panic.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +213,97 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+type ptrImplementsRuntimeObject[T any] interface {
+	*T
+	runtime.Object
+}
+
+// GetOpaqueDeviceConfig returns the config with the highest precedence specified for a device request in a resource claim.
+//
+// Configs can either come from the resource claim itself or from the device
+// class associated with the request. Configs coming directly from the resource
+// claim take precedence over configs coming from the device class. Moreover,
```
pohly, line 243: This is something that needs to be documented for users, so I added it to the API documentation.

But it is more subtle than "one config wins". Suppose `GpuConfig` had two fields A and B which are independent. It would be reasonable to set A in the DeviceClass and B in the claim. `GetOpaqueDeviceConfig` as it stands doesn't support this.

I think what it can do is return all configs that apply to the device. Then the caller can implement whatever conflict resolution strategy makes sense for each field. For example, `GetTimeSlicingConfig` could iterate over the set of configs.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +213,97 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	allConfigs = append(allConfigs, claimConfigs...)
+
+	// Walk all configs from the allConfigs slice in the order they were added.
+	// The latest config matching the GroupVersionKind of the defaultConfig
+	// will be returned. If no configs match, then the original defaultConfig
+	// is returned.
+	candidateConfig := defaultConfig
+	for _, config := range allConfigs {
+		if config.Opaque == nil {
+			continue
```
pohly, line 284: I think this should be an error. The semantic of a one-of is that "empty = some future extension". In this case here, the user is asking for something that the driver cannot do. It shouldn't silently continue because the driver needs to be updated.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +213,97 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+}
+
+// ptrImplementsRuntimeObject is a helper interface for GetOpaqueDeviceConfig.
+//
+// It enforces that a pointer to the type parameter passed to
+// GetOpaqueDeviceConfig implements the runtime.Object interface.
+type ptrImplementsRuntimeObject[T any] interface {
+	*T
+	runtime.Object
+}
```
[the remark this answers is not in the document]
klueska: `config, ok := c.(T)` will return `!ok` because T is not a `runtime.Object` (`*T` is).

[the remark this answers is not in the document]
klueska: I don't see how I would be able to get rid of this check.
```
		if c.GetObjectKind().GroupVersionKind() != TI(&defaultConfig).GetObjectKind().GroupVersionKind() {
			continue
		}
```
I am specifically looking for type `T`, but the scheme might be able to decode multiple types (i.e. any types that get registered with the scheme), not just the `T` type.

cmd/dra-example-kubeletplugin/state.go
klueska: This is exactly what I brought up in the meeting on Monday.

I (thought) we decided that strict "replace" semantics were preferred over ambiguous "merge"semantics (that would allow what you describe above). I don't know how I feel about simply "leaving this up to the driver". I think there should be strict, unambiguous semantics about how multiple (overlapping) configs are treated.

cmd/dra-example-kubeletplugin/state.go
klueska: But we don't even know if this config is meant for this driver yet. We only know that later on when we check `if config.Opaque.Driver != driverName {` below.

klueska: I take it back -- we pre-filter by the request that we know is meant for this driver, so actually both this check and the check below should throw an error if they are not what is expected.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,105 @@
...
+	DefaultTimeSlice TimeSliceDuration = "Default"
+	ShortTimeSlice   TimeSliceDuration = "Short"
+	MediumTimeSlice  TimeSliceDuration = "Medium"
+	LongTimeSlice    TimeSliceDuration = "Long"
+)
+
+// GpuSharingStrategy encodes the valid Sharing strategies as a string.
+type GpuSharingStrategy string
+
+// TimeSliceDuration encodes the valid timeslice duration as a string.
```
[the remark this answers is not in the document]
klueska: I can change the wording in the comment / the name of this field to something other than duration.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,105 @@
...
+type TimeSliceDuration string
+
+// GpuSharing holds the current sharing strategy for GPUs and its settings.
```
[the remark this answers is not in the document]
klueska: I forgot that we allow an empty `requests` slice at all. I don't handle this properly at the moment.

That said, I don't like the notion of "It's up to the vendors to define some same semantics for this."

I know we can't *force* a vendor to adhere to last-one-wins semantics, but we should *highly* encourage it (for consistency) as well as provide an API in our in-tree library to treat them this way.

That is actually the purpose of `GetOpaqueDeviceConfig()` -- I would like to see us move this to the in-tree library at some point.

cmd/dra-example-kubeletplugin/state.go
pohly: When returning a slice of configs, returning nil would be natural. Then the caller can handle the defaulting and this method doesn't need to be generic at all, unless we want to keep the type conversion. We might want to leave it out to keep it generic: theoretically, a driver could also support different kinds (sic!) of configs, in which case the `[]any` result may contain more than one type.

api/example.com/resource/gpu/v1alpha1/api.go
pohly: Or simply document it as "mutates the instance and thus must not be called with nil"?

pohly: Then calling it for nil will panic because its a programming mistake.

api/example.com/resource/gpu/v1alpha1/sharing.go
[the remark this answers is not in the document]
pohly: This may need further discussions in WG Device Management.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
pohly: The Go type conversion makes sure that you get what you are looking for. That aside, why should a driver register more than one type and then only parse one of them? At the moment, the code silently ignores other types (`continue`).

Regarding "multiple types" see https://github.com/kubernetes-sigs/dra-example-driver/pull/54#discussion_r1734505814: I think this helper should support that and decode all of them.

cmd/dra-example-kubeletplugin/state.go
pohly: Encountering configs meant for other drivers is not an error. The wording of the API is:

https://github.com/kubernetes/kubernetes/blob/95b3fe9f15cdcaf98098be398478e70365b12dd7/pkg/apis/resource/types.go#L583-L584

"Driver is used to determine which kubelet plugin needs to be passed these configuration parameters."

This needs to be updated, kubelet no longer "passes" the configuration. But the intent is the same: it allows requesting a device that could come from driver A or driver B and providing configurations for both. That wouldn't work if driver A errors out on the config for driver B and vice versa.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
pohly: > `config, ok := c.(T)` will return `!ok` because T is not a `runtime.Object` (`*T` is).

Right, so the type cast has to use `(*T)`, if we keep doing the cast inside this helper.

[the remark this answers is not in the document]
klueska: Technically it has to use `(TI)` (which the typing system knows is a `(*T)` that implements the `runtime.Object` interface because of the private interface I defined for `ptrImplementsRuntimeObject`. Casting to `(*T)` directly errors out because `T` itself is just `any`.

cmd/dra-example-kubeletplugin/state.go
klueska: That's fair. I wrote this before you reminded me that we can have an empty list for `requests`.  Otherwise, I would say it should error if the passed in request was not meant for this driver.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
pohly: I see. Then replace `T any` with `T runtime.Object`? That changes the API a bit, but at the end of the day the methods is working with pointers, not values, so the rest of the code could as well.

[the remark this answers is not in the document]
pohly: There is no benefit in passing the config instance by value because there is no guarantee that copying it is the same as a deep copy. A pointer makes the sharing more obvious.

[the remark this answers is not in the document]
klueska: Unfortunately I can't pass the type parameter as a pointer because then I have no way of instantiating the non-pointer version of it (the generic function doesn't *know* its a pointer).

In any case, i have refactored things as you suggested and this function will no longer be generic in the latest revision.

api/example.com/resource/gpu/v1alpha1/api.go
klueska: I just added an error -- it makes it more symmetrical with other code around it anyway.

api/example.com/resource/gpu/v1alpha1/sharing.go
```diff
@@ -0,0 +1,105 @@
...
+
+// These constants represent the different TimeSlicing configurations.
+const (
+	DefaultTimeSlice TimeSliceDuration = "Default"
+	ShortTimeSlice   TimeSliceDuration = "Short"
+	MediumTimeSlice  TimeSliceDuration = "Medium"
+	LongTimeSlice    TimeSliceDuration = "Long"
+)
+
+// GpuSharingStrategy encodes the valid Sharing strategies as a string.
```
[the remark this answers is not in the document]
klueska: Done.

api/example.com/resource/gpu/v1alpha1/sharing.go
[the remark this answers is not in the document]
klueska: Changed to `Interval` instead of `Duration`

api/example.com/resource/gpu/v1alpha1/sharing.go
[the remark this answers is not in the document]
klueska: For now I have implemented these semantics outside of the "helper" function (so they are driver specific).

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
klueska: Latest revision pushed.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
pohly: One way to avoid "merging" from different structs would be to make each config struct entirely self-contained. In other words, a struct would never have "two fields"A and B which are independent". Just a thought.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
pohly: > then I have no way of instantiating the non-pointer version of it

You shouldn't have to. The decoder will create a new instance if called without some pre-allocated instance.

cmd/dra-example-kubeletplugin/state.go
pohly: > Otherwise, I would say it should error if the passed in request was not meant for this driver.

Same situation: the request might have been defined such that it could be satisfied by two different drivers.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
klueska: Yeah, there would probably be a way if I did things slightly differently. In any case -- its gone now.

cmd/dra-example-kubeletplugin/state.go
klueska: It still wouldn't force a user to do this...

cmd/dra-example-kubeletplugin/state.go
klueska: True. And then you provide a different config for each driver you know might handle it (knowing only one of them will actually have picked it up). In any case -- its fixed now.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
klueska: Not using autogenerated code.

cmd/dra-example-kubeletplugin/state.go
klueska: Not following this "merge" approach anymore.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +230,78 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	driverName,
+	request string,
+	possibleConfigs []resourceapi.DeviceAllocationConfiguration,
+) ([]runtime.Object, error) {
+	// Collect all configs in order of reverse precedence.
+	var classConfigs []resourceapi.DeviceConfiguration
+	var claimConfigs []resourceapi.DeviceConfiguration
+	var candidateConfigs []resourceapi.DeviceConfiguration
+	for _, config := range possibleConfigs {
+		if !(len(config.Requests) == 0 || slices.Contains(config.Requests, request)) {
```
pohly, line 269:
```suggestion
		// If the config is for specific requests and the current
		// request isn't one of those, the config can be ignored.
		if len(config.Requests) != 0 && !slices.Contains(config.Requests, request) {
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +230,78 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	// The latest config matching the GroupVersionKind of the defaultConfig
+	// will be returned. If no configs match, then the original defaultConfig
+	// is returned.
+	var resultConfigs []runtime.Object
+	for _, config := range candidateConfigs {
+		if config.Opaque == nil {
+			return nil, fmt.Errorf("only opaque parameters are supported by this driver")
+		}
+
+		if config.Opaque.Driver != driverName {
```
pohly, line 294:
```suggestion
		// Configs for different drivers may have
		// been specified because a request might
		// get satisfied by different drivers. This is
		// not an error, drivers must skip over other
		// driver's configs to support this.
		if config.Opaque.Driver != driverName {
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +230,78 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	candidateConfigs = append(candidateConfigs, classConfigs...)
+	candidateConfigs = append(candidateConfigs, claimConfigs...)
+
+	// Walk all configs from the candidateConfigs slice in the order they were added.
+	// The latest config matching the GroupVersionKind of the defaultConfig
+	// will be returned. If no configs match, then the original defaultConfig
+	// is returned.
+	var resultConfigs []runtime.Object
+	for _, config := range candidateConfigs {
+		if config.Opaque == nil {
```
pohly, line 290:
```suggestion
		// If this is nil, then the driver doesn't support
		// some future API extension and has to be updated.
		if config.Opaque == nil {
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -173,3 +230,78 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+			return nil, fmt.Errorf("invalid config source: %v", config.Source)
+		}
+	}
+	candidateConfigs = append(candidateConfigs, classConfigs...)
+	candidateConfigs = append(candidateConfigs, claimConfigs...)
+
+	// Walk all configs from the candidateConfigs slice in the order they were added.
+	// The latest config matching the GroupVersionKind of the defaultConfig
+	// will be returned. If no configs match, then the original defaultConfig
+	// is returned.
```
pohly, line 287:
```suggestion
	// Decode all configs that are relevant for the driver.
```

I think this comment is mostly out-dated.

api/example.com/resource/gpu/v1alpha1/api.go
```diff
@@ -17,32 +17,94 @@
...
 }
 
-func DefaultGpuClaimParametersSpec() *GpuClaimParametersSpec {
-	return &GpuClaimParametersSpec{
-		Count: ptr.To(1),
+// Normalize updates a GpuConfig config with implied default values based on other settings.
+func (c *GpuConfig) Normalize() error {
+	if c == nil {
+		return fmt.Errorf("config is 'nil'")
+	}
```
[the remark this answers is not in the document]
klueska: Will leave as is then with the check in place.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -151,17 +160,65 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+		)
+		if err != nil {
+			return nil, fmt.Errorf("error getting GPU configs: %v", err)
+		}
+
+		// Select the GPU config with the highest precedence from those retrieved.
+		// Use the default GPU config if no GPU configs have been returned.
+		config := configapi.DefaultGpuConfig()
+		for _, c := range configs {
+			if c.GetObjectKind().GroupVersionKind() != config.GetObjectKind().GroupVersionKind() {
```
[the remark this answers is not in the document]
klueska: If a new type was added to the scheme that is associated with the Decoder, but is not a `GpuConfig` (for example a separate `MigConfig` in the case of NVIDIA GPUs). We don't have another one yet, but we could / maybe should in the future as an example.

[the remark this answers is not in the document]
klueska: Actually, wait -- I see your point now. The Decoder will already filter by *some* valid config that is registered with it, and then it is just up to the cast below to pull out the right one. This would also work for other types that are registered as well. 

Will remove the check now.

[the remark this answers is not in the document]
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
klueska: Updated.

cmd/dra-example-kubeletplugin/state.go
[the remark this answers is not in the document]
pohly: > The Decoder will already filter by some valid config that is registered with it

Exactly. There is exactly one GroupKind for each type, so if you get that type, it has to match the expected GroupKind.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -151,17 +160,63 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+
+		// Select the GPU config with the highest precedence from those retrieved.
+		// Use the default GPU config if no GPU configs have been returned.
+		config := configapi.DefaultGpuConfig()
+		for _, c := range configs {
+			switch castConfig := c.(type) {
+			case *configapi.GpuConfig:
+				config = castConfig
+			default:
+				return nil, fmt.Errorf("runtime object is not a regognized configuration")
```
pohly, line 189: One could also call this an "internal error" because why did the driver register a type that it doesn't recognize here?

But let's leave it as-is.

/lgtm
/approve
/hold

In case you want to get a second opinion (@byako ?), otherwise feel free to lift the hold.

klueska: Evan did an initial review and had approved it previously. Let's unhold it and I will make any changes as a follow-up if required.

PR #58, opened 2024-09-07
Refactor how opaque device configs are handled

klueska:
Previously, each config was being applied independently to each request
that referenced it. However, some configs may need to operate
collectively on all of the requests they are associated with it.

The code has been refactored to handle this situation. Additionally, the
code to define the ContainerEdits for any custom config has been moved
into the config code itself to better encapsulate it.


pohly (approved):
/lgtm
/approve

PR #61, opened 2024-09-16
demos: handle SIGTERM to avoid long termination

bart0sh:
This should fix https://github.com/kubernetes/kubernetes/issues/127188
Please, have a look at the issue for further details, thanks!

### DIFF
--- a/demo/gpu-test2.yaml
+++ b/demo/gpu-test2.yaml
@@ -35,7 +35,7 @@ spec:
   - name: ctr0
     image: ubuntu:22.04
     command: ["bash", "-c"]
-    args: ["export; sleep 9999"]
+    args: ["export; trap 'exit 0' TERM; sleep 9999 & wait"]
     resources:
       claims:
       - name: gpus
--- a/test/e2e/e2e.sh
+++ b/test/e2e/e2e.sh
@@ -63,4 +63,12 @@ if [ $gpu_test_5 != 1 ]; then
     exit 1
 fi
 
+# test that deletion is fast (less than the default grace period of 30s)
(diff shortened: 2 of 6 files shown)


demo/gpu-test2.yaml
```diff
@@ -31,11 +31,12 @@ metadata:
   labels:
     app: pod
 spec:
+  shareProcessNamespace: true
```
klueska, line 34: This shouldn't be necessary.

test/e2e/e2e.sh
```diff
@@ -63,4 +63,12 @@ if [ $gpu_test_5 != 1 ]; then
     exit 1
 fi
 
+# test that deletion is fast (less than the default grace period of 30s)
+# see https://github.com/kubernetes/kubernetes/issues/127188 for details
+kubectl delete -f demo/gpu-test1.yaml --timeout=20s
+kubectl delete -f demo/gpu-test2.yaml --timeout=20s
+kubectl delete -f demo/gpu-test3.yaml --timeout=20s
+kubectl delete -f demo/gpu-test4.yaml --timeout=20s
+kubectl delete -f demo/gpu-test5.yaml --timeout=20s
```
klueska, line 72: Let's make it 25 just to further reduce the chance of flakes

bart0sh: done

demo/gpu-test2.yaml
bart0sh: yeh, it's a leftover from my experiments. Thank you for pointing out! Fixed.

klueska (commented):
/lgtm

PR #62, opened 2024-09-16
Update demo specs to replace sleep command with one that catches SIGTERM

klueska:
This allows all of the demo pods to terminate immediately, without needing to wait for the grace-period of 30s before they receive a SIGKILL.


klueska:
Looks like @bart0sh beat me to it with #61

Issue #48, opened 2024-06-06
update deployment for 1.31


pohly:
https://github.com/kubernetes/kubernetes/pull/125163 is making the drivers responsible for publishing ResourceSlices and getting the ResourceClaim. The example driver needs to be updated to work with that change.

Furthermore, we need to demonstrate how apiserver access can be limited. See https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dra/test-driver/deploy/example/plugin-permissions.yaml.

We also should document that deployments are expected to limit access to admin access as in https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dra/test-driver/deploy/example/admin-access-policy.yaml

pohly:
Some relevant documentation:
- https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#additional-metadata-in-pod-bound-tokens (provides the correlation with the node, if I am not mistaken)

carlory:
/unassign

For some reason, I may not have time to complete this task. If I am free and the task still needs volunteers, I will come back to claim it. sorry for the delay @pohly

carlory:
If someone raises a PR, I am very happy to review together

jackfrancis:
should we close this based on https://github.com/kubernetes-sigs/dra-example-driver/pull/50 ?

klueska:
Yes. Closed by https://github.com/kubernetes-sigs/dra-example-driver/pull/53

klueska:
Actually, let's reopen this, as not all all of the criteria that @pohly mentions in the issue has been fulfilled. That is to say that basic support for 1.31 has been merged (i.e. it is functional), but it doesn't check all of the boxes that @pohly has for being done the "right" way.

pohly:
I agree, let's add at least the RBAC rules. I've updated the description with more recent links.

jackfrancis:
cc @nojnhuh @ritazh

PR #66, opened 2024-10-20
Update deployment for Kubernetes 1.31

bart0sh:
- Modified ClusterRole to include only API objects that the driver can access 
- Added ValidatingAdmissionPolicy for resourceslices

Fixes https://github.com/kubernetes-sigs/dra-example-driver/issues/48

### DIFF
--- a/deployments/helm/dra-example-driver/templates/validatingadmissionpolicy.yaml
+++ b/deployments/helm/dra-example-driver/templates/validatingadmissionpolicy.yaml
@@ -1,0 +1,31 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: resourceslices-policy-{{ include "dra-example-driver.fullname" . }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["resource.k8s.io"]
+      apiVersions: ["v1alpha3"]
+      operations:  ["CREATE", "UPDATE", "DELETE"]
+      resources:   ["resourceslices"]
+  matchConditions:
+  - name: isRestrictedUser
+    expression: >-
+      request.userInfo.username == "system:serviceaccount:{{ include "dra-example-driver.namespace" . }}:{{ include "dra-example-driver.serviceAccountName" . }}"
+  variables:
+  - name: userNodeName
+    expression: >-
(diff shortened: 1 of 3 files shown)


klueska:
Can we add the adminssion policy stuff as a separate PR where we also add support for requesting resources using AdminAccess?

bart0sh:
@klueska done. Any ideas on where to add support for requesting resources using AdminAccess?

deployments/helm/dra-example-driver/templates/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
```
klueska, line 2: I'm actually not that familiar with `ValidatingAdmissionPolicy`. Can you explain to me in english what this file (in conjunction with [validatingadmissionpolicybinding.yaml](https://github.com/kubernetes-sigs/dra-example-driver/pull/66/files#diff-ae0655a5f1431437c30dc2197f4993c35a566fff2838db7a59992eebcd9a1d62)) is meant to do?

bart0sh: I understand it as a set of rules to restrict access from the client (DRA plugin in this case) to the API server. I used [test driver admin-access policy](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dra/test-driver/deploy/example/admin-access-policy.yaml) as an example.

klueska: My question is more around what this gives us beyond what we get from the ClusterRole already? At a quick glance it looks as if it also limits each instance of a driver daemonset to only modify the resourceslices on the node it is running on. Is that correct? Is there more to it than that?

bart0sh: It also sets a list of acceptable API versions, `v1alpha3` in this particular case.

pohly: This is  just a selector for what the policy applies to. Once we have v1beta1 and a client uses it, this policy would be ignored.

klueska (approved):
/approve
/lgtm

PR #70, opened 2024-12-10
add unit tests

### LINKED ISSUE #41: Add unit testing
We should start adding basic unit tests and include a GitHub action that runs these.

nojnhuh:
This PR adds unit tests for at least the low-hanging fruit and enables the tests in CI. I think what I didn't include here already has decent coverage from the e2e tests, but please let me know if I missed something that would be valuable to add unit tests for.

Fixes #41

### DIFF
--- a/api/example.com/resource/gpu/v1alpha1/api_test.go
+++ b/api/example.com/resource/gpu/v1alpha1/api_test.go
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2025 The Kubernetes Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v1alpha1
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGpuConfigNormalize(t *testing.T) {
+	tests := map[string]struct {
+		gpuConfig   *GpuConfig
+		expected    *GpuConfig
+		expectedErr error
+	}{
+		"nil GpuConfig": {
+			gpuConfig:   nil,
+			expectedErr: errors.New("config is 'nil'"),
+		},
+		"empty GpuConfig": {
+			gpuConfig: &GpuConfig{},
+			expected: &GpuConfig{
+				Sharing: &GpuSharing{
+					Strategy: TimeSlicingStrategy,
+					TimeSlicingConfig: &TimeSlicingConfig{
+						Interval: DefaultTimeSlice,
+					},
+				},
+			},
+		},
+		"empty GpuConfig with SpacePartitioning": {
+			gpuConfig: &GpuConfig{
+				Sharing: &GpuSharing{
+					Strategy: SpacePartitioningStrategy,
+				},
+			},
+			expected: &GpuConfig{
+				Sharing: &GpuSharing{
+					Strategy: SpacePartitioningStrategy,
+					SpacePartitioningConfig: &SpacePartitioningConfig{
+						PartitionCount: 1,
+					},
+				},
+			},
+		},
+		"full GpuConfig": {
+			gpuConfig: &GpuConfig{
+				Sharing: &GpuSharing{
+					Strategy: SpacePartitioningStrategy,
+					TimeSlicingConfig: &TimeSlicingConfig{
+						Interval: ShortTimeSlice,
+					},
+					SpacePartitioningConfig: &SpacePartitioningConfig{
+						PartitionCount: 5,
+					},
+				},
+			},
+			expected: &GpuConfig{
+				Sharing: &GpuSharing{
+					Strategy: SpacePartitioningStrategy,
+					TimeSlicingConfig: &TimeSlicingConfig{
+						Interval: ShortTimeSlice,
+					},
+					SpacePartitioningConfig: &SpacePartitioningConfig{
+						PartitionCount: 5,
+					},
+				},
+			},
+		},
+		"default GpuConfig is already normalized": {
+			gpuConfig: DefaultGpuConfig(),
+			expected:  DefaultGpuConfig(),
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			err := test.gpuConfig.Normalize()
+			assert.Equal(t, test.expected, test.gpuConfig)
+			assert.Equal(t, test.expectedErr, err)
+		})
+	}
+}
--- a/api/example.com/resource/gpu/v1alpha1/sharing_test.go
+++ b/api/example.com/resource/gpu/v1alpha1/sharing_test.go
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2025 The Kubernetes Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


api/example.com/resource/gpu/v1alpha1/api_test.go
```diff
@@ -0,0 +1,116 @@
...
+				if err == nil {
+					t.Errorf("expected error %q, but got no error", test.expectedErr)
+				} else if err.Error() != test.expectedErr {
+					t.Errorf("expected error %q, but got %v", test.expectedErr, err)
+				}
+			} else {
+				if err != nil {
+					t.Errorf("expected Normalize to succeed, got error %v", err)
+				}
+			}
```
elezar, line 110: This can get quite verbose if repeated for each test. @pohly is there a recommended way to handle errors in tests in other k8s projects?

api/example.com/resource/gpu/v1alpha1/sharing_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}{
+		{
+			name:        "nil GpuSharing",
+			gpuSharing:  nil,
+			expectedErr: "no sharing set to get config from",
+		},
+		{
+			name: "strategy is not TimeSlicing",
+			gpuSharing: &GpuSharing{
+				Strategy: "not" + TimeSlicingStrategy,
```
elezar, line 40: Question: Does using a valid non-TimeSlicingStrategy here make sense?

api/example.com/resource/gpu/v1alpha1/sharing_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}{
+		{
+			name:        "nil GpuSharing",
+			gpuSharing:  nil,
+			expectedErr: "no sharing set to get config from",
+		},
+		{
+			name: "strategy is not SpacePartitioning",
+			gpuSharing: &GpuSharing{
+				Strategy: "not" + SpacePartitioningStrategy,
```
elezar, line 102: similar comment as above.

api/example.com/resource/gpu/v1alpha1/validate_test.go
```diff
@@ -0,0 +1,110 @@
...
+ * limitations under the License.
+ */
+
+package v1alpha1
+
+import (
+	"testing"
+)
+
+func TestGpuConfigValidate(t *testing.T) {
```
elezar, line 23: One test case that is not covered is that if both TimeSlicing and SpacePartitioning are defined, an invalid SpacePartitioning config is ignored.

This might not be intended behaviour, but it may be behaviour that users start to depend on. (although since this is an example driver it's probably not critical) cc @pohly @klueska

elezar (commented):
Thanks @nojnhuh. I've made a first pass.

I have a few minor questions and comments.

api/example.com/resource/gpu/v1alpha1/api_test.go
```diff
@@ -0,0 +1,116 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
```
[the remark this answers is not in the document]
pohly: I find assertions with github.com/stretchr/testify easier to read and write, but that might just be me.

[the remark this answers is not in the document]
nojnhuh: I like using cmp for the diff you get in the failure message, but I see testify's `Equal` also has that built in, so I can switch these over to that.

api/example.com/resource/gpu/v1alpha1/api_test.go
nojnhuh: This gets cleaned up a fair bit using testify's `ErrorEqual` and `NoError`, but if there's a premade wrapper around those together that should work here. I could make that kind of wrapper here too.

api/example.com/resource/gpu/v1alpha1/sharing_test.go
nojnhuh: Yeah that would be equivalent here. I tend toward this kind of mangling of the "correct" value make it "incorrect" to guard against the extreme edge case where two constants like `TimeSlicingStrategy` and `SpacePartitioningStrategy` actually have the same value. Overall though I just want to make these tests as least surprising or interesting as possible, so I can change this.

api/example.com/resource/gpu/v1alpha1/validate_test.go
nojnhuh: Agree, that seems worth recording in a test, even if just to signal to authors of real drivers that that's something to look out for and potentially not recreate.

nojnhuh:
Pushed a merge commit to fix the conflict in case that's easier to review than a rebase, but I'll plan to squash before merging this.

/hold

api/example.com/resource/gpu/v1alpha1/api_test.go
elezar: let's just start with:
```
if test.expecteError != "" {
   require.EqualError(t, err, test.expecteError)
} else {
   require.NoError(t, err)
}
```
We can look at a more comprehensive wrapper later if required.

go.mod
```diff
@@ -3,8 +3,10 @@ module sigs.k8s.io/dra-example-driver
 go 1.23.1
 
 require (
+	github.com/google/go-cmp v0.6.0
```
elezar, line 6: I thought we removed the use of `cmp`?

elezar (commented):
Minor comment on the go modules, but otherwise this looks good.

api/example.com/resource/gpu/v1alpha1/api_test.go
elezar: (I see that's what you've already done!).

api/example.com/resource/gpu/v1alpha1/api_test.go
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2024 The Kubernetes Authors.
```
bart0sh, line 2: 2025?

api/example.com/resource/gpu/v1alpha1/api_test.go
```diff
@@ -0,0 +1,108 @@
...
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGpuConfigNormalize(t *testing.T) {
+	tests := []struct {
```
bart0sh, line 26: nit: Making it a map would change an order of test cases for each run, which can help ensuring that they don't influence each other.
```suggestion
	tests := map[string]struct {
```

go.mod
nojnhuh: Ah yep, just missed a `go mod tidy`.

api/example.com/resource/gpu/v1alpha1/api_test.go
nojnhuh: Happy New Year!

nojnhuh:
I took the liberty of squashing after addressing the latest round of feedback.

/hold cancel

api/example.com/resource/gpu/v1alpha1/validate_test.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGpuConfigValidate(t *testing.T) {
+	tests := map[string]struct {
+		gpuConfig *GpuConfig
+		expected  string
```
bart0sh, line 28: nit: changing type to `error` would simplify the code a little bit and make it more logical in my opinion as `expected` is what `gpuConfig.Validate()` returns.
```suggestion
		expected  error
```

api/example.com/resource/gpu/v1alpha1/validate_test.go
```diff
@@ -0,0 +1,124 @@
...
+		"empty GpuConfig": {
+			gpuConfig: &GpuConfig{},
+			expected:  "no sharing strategy set",
+		},
+		"empty GpuConfig.Sharing": {
+			gpuConfig: &GpuConfig{
+				Sharing: &GpuSharing{},
+			},
+			expected: "unknown GPU sharing strategy: ",
+		},
```
bart0sh, line 39: I'd also add this test case:
```
		"unknown GPU sharing strategy": {
			gpuConfig: &GpuConfig{
				Sharing: &GpuSharing{
					Strategy: "unknown",
				},
			},
			expected: errors.New("unknown GPU sharing strategy: unknown"),
		},
```
BTW, this test case shows that[ this code](https://github.com/kubernetes-sigs/dra-example-driver/blob/main/api/example.com/resource/gpu/v1alpha1/validate.go#L65) is unreachable.

elezar (approved):
Lgtm

PR #73, opened 2025-01-14
Verify environment variables in e2e test

### LINKED ISSUE #57: Enhance e2e tests to detect GPU devices
Currently, we are verifying that the pods went Running. This is not technically correct as it could be possible to schedule without having DRA set up.

See https://github.com/kubernetes-sigs/dra-example-driver/pull/56#discussion_r1740080606.

nojnhuh:
This change adds new steps to the e2e test to verify that each container in each pod contains logs showing the presence of any `GPU_DEVICE_*` environment variables. I checked that this at least catches where the CDI config is not setup properly, but didn't try all of the other possible failure modes here.

Fixes #57

### DIFF
--- a/test/e2e/e2e.sh
+++ b/test/e2e/e2e.sh
@@ -27,6 +27,43 @@ kubectl create -f demo/gpu-test3.yaml
 kubectl create -f demo/gpu-test4.yaml
 kubectl create -f demo/gpu-test5.yaml
 
+function gpus-from-logs {
+  local logs="$1"
+  echo "$logs" | sed -nE "s/^declare -x GPU_DEVICE_[[:digit:]]+=\"(.+)\"$/\1/p"
+}
+
+function gpu-id {
+  local gpu="$1"
+  echo "$gpu" | sed -nE "s/^gpu-([[:digit:]]+)$/\1/p"
+}
+
+function gpu-sharing-strategy-from-logs {
+  local logs="$1"
+  local id="$2"
+  echo "$logs" | sed -nE "s/^declare -x GPU_DEVICE_${id}_SHARING_STRATEGY=\"(.+)\"$/\1/p"
+}
+
+function gpu-timeslice-interval-from-logs {
+  local logs="$1"
+  local id="$2"
+  echo "$logs" | sed -nE "s/^declare -x GPU_DEVICE_${id}_TIMESLICE_INTERVAL=\"(.+)\"$/\1/p"
+}
+
+function gpu-partition-count-from-logs {
+  local logs="$1"
+  local id="$2"
+  echo "$logs" | sed -nE "s/^declare -x GPU_DEVICE_${id}_PARTITION_COUNT=\"(.+)\"$/\1/p"
+}
+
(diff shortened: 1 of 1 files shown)


test/e2e/e2e.sh
```diff
@@ -42,28 +55,33 @@ if [ $gpu_test_2 != 1 ]; then
...
     exit 1
 fi
+verify-env gpu-test3 pod0
 
 kubectl wait --for=condition=Ready -n gpu-test4 pod/pod0 --timeout=120s
 kubectl wait --for=condition=Ready -n gpu-test4 pod/pod1 --timeout=120s
 gpu_test_4=$(kubectl get pods -n gpu-test4 | grep -c 'Running')
 if [ $gpu_test_4 != 2 ]; then
-    echo "gpu_test_4 $gpu_test_4 failed to match against 1 expected pods"
+    echo "gpu_test_4 $gpu_test_4 failed to match against 2 expected pods"
```
nojnhuh, line 72: FYI I did sneak in this typo fix also.

elezar:
@nojnhuh this is definitely an improvement, thanks.

Do you have a feel for the effort required to actually test the expected behaviour as called out in the README?

nojnhuh:
> > Do you have a feel for the effort required to actually test the expected behaviour as called out in the README?
> 
> Do you mean checking at least as far as whether certain containers have the same or different GPUs allocated based on the environment variables defined in each container? And that the correct TimeSlicing/SpacePartitioning parameters are set? I could definitely do that, but it might get a little messy in a bash script.

I pushed these changes in a new commit which I plan to squash if we're good with those changes.

/hold

test/e2e/e2e.sh
```diff
@@ -35,6 +72,36 @@ if [ $gpu_test_1 != 2 ]; then
     exit 1
 fi
 
+gpu_test1_pod0_ctr0_logs=$(kubectl logs -n gpu-test1 pod0 -c ctr0)
```
nojnhuh, line 75: I'm very open to suggestions for ways to make this less of a copy-paste nightmare.

nojnhuh:
@elezar This should be ready for another review, PTAL.

elezar (approved):
Thanks @nojnhuh.

This is quite verbose, but I suppose that's a side-effect of using bash instead of writing the tests in Ginkgo -- which is out of scope for this PR.

LGTM

nojnhuh:
Squashed commits, didn't make any more changes in the process like the [diff](https://github.com/kubernetes-sigs/dra-example-driver/compare/a4795f1e51aca3fef219c54494529c08573833da..9c39b1ebfb6705b287092d86cefc68c04edb4a25) shows.

elezar (approved):
Lgtm

nojnhuh:
Forgot to remove the hold here once I squashed.

/hold cancel

pohly:
/lgtm

Based on https://github.com/kubernetes-sigs/dra-example-driver/pull/73#pullrequestreview-2589524009.

PR #74, opened 2025-01-14
Update copyright year in generated code

nojnhuh:
Since the Makefile instructs controller-gen to inject the current year into the code it generates, the turn of the calendar has caused it to become out of date. Alternatively, we could remove the year entirely to keep from having to do this annually: https://github.com/kubernetes-sigs/dra-example-driver/blob/f0bee95eb22239bbb06bc0d4b21b1c5a3d7ee6a7/Makefile#L111
https://github.com/kubernetes-sigs/dra-example-driver/blob/f0bee95eb22239bbb06bc0d4b21b1c5a3d7ee6a7/hack/boilerplate.go.txt#L2


klueska (approved):

byako:
The generated code would get new year if there was a change in the code. If we want to keep copyright year up to date even in files that were not changed since last year - would you consider also update the year in the rest of the files? 

Anyone knows what's the policy in general? I have been updating copyright year to current only in files that are were changed in the current year.

pohly:
The usual approach in Kubernetes is to set the copyright year the first time that a file gets created and then not update it again later, not even when editing the file.

Instead of injecting a varying year, I would copy some fixed year from the input files into the generated files - but I haven't looked into details here.

nojnhuh:
It seems the pattern at least in k/k is to omit the copyright year in generated code, so I applied that here as well.

s/Update/Remove/:

/retitle Remove copyright year in generated code

hack/boilerplate.generatego.txt
nojnhuh, line 1: hack/boilerplate.go.txt is no longer referenced, but I left it in the repo in case that's useful for when something like a `make verify-boilerplate` target is introduced. Happy to remove it though now if we only want to add it back once we need it.

PR #79, opened 2025-02-26
use _GIT_TAG for staging repo image tag

nojnhuh:
`$COMMIT_SHA` seems to be empty as the resulting image was tagged with the default `v0.1.0` tag: https://console.cloud.google.com/artifacts/docker/k8s-staging-images/us-central1/dra-example-driver/dra-example-driver/sha256:0e8f5df34c196e5900c4d29667ce5b21ba482ef9aecc5dd19cea14fa6dde2d0c;tab=attachments?project=k8s-staging-images

`$_GIT_TAG` is a variable injected by the Prow job that seems to be what's generally used for image tags: https://github.com/kubernetes/test-infra/blob/3e2b92318e41a47e569a90f0c71db3bcb81b9902/images/builder/README.md

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,7 +7,7 @@ steps:
   entrypoint: bash
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
-  - DRIVER_IMAGE_TAG=$COMMIT_SHA
+  - DRIVER_IMAGE_TAG=$_GIT_TAG # _GIT_TAG is injected by Prow
   args:
     - -ec
     - |


cloudbuild.yaml
```diff
@@ -7,7 +7,7 @@ steps:
   entrypoint: bash
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
-  - DRIVER_IMAGE_TAG=$COMMIT_SHA
+  - DRIVER_IMAGE_TAG=$_GIT_TAG
```
pohly, line 10: I am unsure how expansion works in the `env` section. This might have been the problem with the original `$COMMIT_SHA`, too.

Let's do it as in https://github.com/kubernetes-csi/csi-release-tools/blob/90efb2ca59900f19eba05e65da28beda79c5bb28/cloudbuild.yaml#L32-L47 ?

pohly (changes requested):

cloudbuild.yaml
nojnhuh: I've updated this to follow that pattern.

The [docs](https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values) show at least one example of substitutions of these automatic variables without an explicit mapping in `substitutions`. I think `COMMIT_SHA` was empty because Prow invokes a "manually triggered" build which doesn't seem to include that variable.

nojnhuh: Sounds good, changed this back.

cloudbuild.yaml
```diff
@@ -7,7 +7,7 @@ steps:
   entrypoint: bash
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
-  - DRIVER_IMAGE_TAG=$COMMIT_SHA
+  - DRIVER_IMAGE_TAG=$_GIT_TAG # _GIT_TAG is injected by Prow
```
pohly, line 10: Nit: `_GIT_TAG` doesn't come from Prow, it's from the builder tool.

Magic.

Let's try it, we can always iterate further.

/lgtm
/approve

PR #81, opened 2025-03-01
Add Helm chart OCI push to cloudbuild

nojnhuh:
These changes enable the Helm chart to be pushed as an OCI artifact to the staging repository. This is the rough release process I had in mind, hoping to gather feedback and iterate in this PR:

1. We open a PR here to update the Chart.yaml's `appVersion` to match the new tag that will be created.
2. A maintainer pushes a semver Git tag (e.g. `v0.1.0`) which matches the chart's new `appVersion`.
    - Prow kicks off an image build and pushes to the staging repo, resulting in the image `us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/dra-example-driver:v0.1.0`.
3. For the same commit, a maintainer pushes a `chart/<semver>` Git tag, (e.g. `chart/0.12.0`, [no leading "v"](https://github.com/kubernetes-sigs/kueue/issues/4445)).
    - Prow similarly pushes the chart to `us-central1-docker.pkg.dev/k8s-staging-images/charts/dra-example-driver:0.12.0`
4. Anyone can then open a PR to k8s.io that will promote the image and chart to registry.k8s.io (e.g. https://github.com/kubernetes/k8s.io/pull/7830).
5. A maintainer publishes a GitHub release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor/
 ./dra-example-kubeletplugin
 
 driver_image.tar
+*.tgz
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ $(DOCKER_TARGETS): docker-%: .build-image
 			make $(*)
 
 # Start an interactive shell using the development image.
-PHONY: .shell
+.PHONY: .shell
 .shell:
 	$(CONTAINER_TOOL) run \
 		--rm \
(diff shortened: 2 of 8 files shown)


kannon92:
I would just publish the images and charts in the same step.

nojnhuh:
> I would just publish the images and charts in the same step.

Makes sense. That might not be possible if the chart refers to the image by its digest if we can't determine that before the image is built. But that does cut out a few steps, so I think that's worthwhile. I'll update the description.

byako (approved):
lgtm, not adding a label because I'm not familiar with registry.k8s.io permissions.

pohly:
> lgtm, not adding a label because I'm not familiar with registry.k8s.io permissions.

Neither am I, but I am not sure whether we have anyone better suited for a formal approval. Let's go ahead and refine if something doesn't work...

/lgtm

PR #83, opened 2025-03-06
Fix image and chart tags used for staging repo

nojnhuh:
This PR replaces our use of the automatically-injected `_GIT_TAG` substitution in cloudbuild.yaml with the plain Git tag, not prefixed by a date. When the tag is prefixed by the date, the old logic in the Makefile to detect a `chart/` prefix failed to trigger, so the chart did not get pushed to the staging repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ MKDIR    ?= mkdir
 TR       ?= tr
 DIST_DIR ?= $(CURDIR)/dist
 
+export IMAGE_GIT_TAG ?= $(shell git describe --tags --always --dirty --match 'v*')
+export CHART_GIT_TAG ?= $(shell git describe --tags --always --dirty --match 'chart/*')
+
 include $(CURDIR)/common.mk
 
 BUILDIMAGE_TAG ?= golang$(GOLANG_VERSION)
@@ -168,11 +171,8 @@ $(DOCKER_TARGETS): docker-%: .build-image
 
 .PHONY: push-release-artifacts
 push-release-artifacts:
-	if echo -n "${GIT_TAG}" | grep -q "^chart/"; then \
-		export CHART_VERSION="$${GIT_TAG##chart/}"; \
-		demo/scripts/push-driver-chart.sh; \
-	else \
-		export DRIVER_IMAGE_TAG="${GIT_TAG}"; \
-		demo/scripts/build-driver-image.sh; \
-		demo/scripts/push-driver-image.sh; \
-	fi
+	CHART_VERSION="$${CHART_GIT_TAG##chart/}" \
+		demo/scripts/push-driver-chart.sh
+	export DRIVER_IMAGE_TAG="${IMAGE_GIT_TAG}"; \
+	demo/scripts/build-driver-image.sh && \
+	demo/scripts/push-driver-image.sh
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -8,6 +8,5 @@ steps:
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
   - DRIVER_CHART_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/charts
-  - GIT_TAG=$_GIT_TAG # _GIT_TAG is injected by the image builder running in Prow
   args:
     - push-release-artifacts


deployments/helm/dra-example-driver/Chart.yaml
```diff
@@ -23,6 +23,6 @@ version: 0.0.0-dev
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "v0.1.0"
+appVersion: "v0.1.1"
```
nojnhuh, line 26: @pohly If you could please cut a `v0.1.1` (and `chart/0.1.1`) tag after this merges that would be great! Sorry again for the churn here.

pohly: Do those always have to match? Could we release chart/0.1.1 for appVersion v0.1.0?

Or perhaps chart/0.1.1.1 for appVersion v0.1.0 and then chart/0.1.2 for appVersion v0.1.2?

klueska: They don't have to match, but they should in this case.

nojnhuh: > Do those always have to match? Could we release chart/0.1.1 for appVersion v0.1.0?

I tried to set things up so that these don't always have to match. I think we could still promote the image that got built from the v0.1.0 and give it a `v0.1.0` tag (and skip a `v0.1.1` tag), assuming there isn't some extra check during image promotion that the tag names in the [image lists](https://github.com/kubernetes/k8s.io/blob/535b8841c6c5d6f6a45f56bf5f4859dc54c101a5/registry.k8s.io/images/k8s-staging-kueue/images.yaml) match the tag names in the staging repo. If we want to try that, I can remove the `appVersion` bump from this PR.

pohly: I'm just trying to rationalize when the versions should match and when it's okay if they don't.

If we merge this and release v0.1.1 as update of v0.1.0, the only change in the example driver itself will be this PR. That's not really a reason to release a new example driver, is it?

nojnhuh: You're right that it's not, and maybe it's actually good to rip off the "the versions are coincidentally in sync" bandaid early.

I couldn't find any evidence that promoting a `v0.1.0` tag requires that same tag to exist in the staging repo, so I think we should be good to keep v0.1.0 of the image and only push a new `chart/0.1.1` tag after this merges. I'll update the PR to omit the `appVersion` change.

klueska: I'll reiterate that while you can bring them out of sync, I would highly recommend we don't. It leads to people talking about version X of the component and then having to clarify if you mean version X of the helm chart or version X of the image. We started off trying to do the "right" thing by versioning all of the helm charts separate from the components we own for the NVIDIA container stack, and eventually brought them back in sync because of all the confusion it caused with our user-base.

klueska: It also means, as a developer, you don't need to think much about which version needs to be updated when. If you are doing a release, you are always bumping both, rather than having to think about 2 different release cycles for the same project.

klueska: Also, what does the git-tag mean if your bring them out of sync? Is it the tag for the helm chart or the tag for the image? If you choose one of them, why that one? Both of them are reflected in your code, so why is one treated different than the other?

klueska: I see you tried to solve this partially with a separate tag for `v0.1.0` and `chart/v0.1.0`. I am strongly against doing this and propose we remove the `chart/v*` variants and stick with a single version tag as proposed in my other comments.

nojnhuh: Helm thinks `v`-prefixed chart versions are invalid, so it fails to do any kind of semver-based matching or sorting against those: https://github.com/kubernetes-sigs/kueue/issues/4445. If keeping the versions identical is important, then we should probably go the other way and leave off the `v`.

klueska: Hmm, that used to be an issue, but I was under the impression it was fixed quite a while ago. What version of helm was this issue seen in?

klueska: If this is still an issue, I feel less strongly about having a v on the chartVersion than keeping the appVersion and chartVersion in sync.

nojnhuh: I see that issue with the latest:
```
% helm version
version.BuildInfo{Version:"v3.17.1", GitCommit:"980d8ac1939e39138101364400756af2bdee1da5", GitTreeState:"", GoVersion:"go1.24.0"}
```

nojnhuh: The issue I see with keeping the chart and app versions in sync is that the chart and container image are fundamentally different artifacts with different interfaces, so bumping the version for both the same way for the same change doesn't always make sense. You end up in a situation where version bumps for one of those artifacts are either overly conservative (bumping the major version of the image for a breaking change in the chart even if the image hasn't changed) or outright break semver (fixing a bug in the image requiring a breaking change to the chart, but only bumping the patch version of the chart). Those situations are more confusing IMO than having separate versions for the image and the chart.

If we expect users will only consume the image through the chart, then maybe we can flip things a bit where the chart version becomes "the version" of the driver and the image version isn't really advertised. We can't guarantee or enforce that users only use the image with the chart though, so following semver for the image independent from the chart still seems worth the little bit of extra effort.

klueska: I understand the implications that you point out, but (at least for our components at NVIDIA), being overly conservative is worth the tradeoff for simplicity and ease of understanding for our users.

I'm not going to block you if you feel strongly about keeping the versions separate. But my biggest hangup is that this project is meant to provide the recommended way to structure a DRA driver, and (in my opinion), I would not recommend separating them. The NVIDIA DRA drivers won't follow this example.

nojnhuh: Totally understand one approach isn't going to work for everyone. I think I'd still prefer separating the versions here for now, but open to aligning those later if that makes more sense then.

@pohly I've removed the `appVersion` bump here so this PR is ready if we decide to continue with v0.1.0 of the image and 0.1.1 for the chart.

pohly: Let's proceed with different versions for now. Once this works, we should document the approach taken here and what the alternative is. Then DRA driver authors can decide for themselves.

pohly: /lgtm
/approve

pohly: -> https://github.com/kubernetes-sigs/dra-example-driver/releases/tag/chart%2F0.1.1

nojnhuh: Thanks! We hit a new bump in the road: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-dra-example-driver-image/1899547129681022976. While looking at how kueue has things set up I didn't catch that their scripts install Helm since it doesn't come baked into the docker image.

I can see at least that chart version looks correct this time. Sorry this is taking quite a few iterations, but we're making progress!

nojnhuh: I opened a fix for that issue: #84

PR #84, opened 2025-03-11
invoke Helm with `go run`

nojnhuh:
The docker image used to push the Helm chart doesn't include `helm`, so we have to ensure that's available ourselves. This PR adds that by invoking Helm with `go run`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ CONTAINER_TOOL ?= docker
 MKDIR    ?= mkdir
 TR       ?= tr
 DIST_DIR ?= $(CURDIR)/dist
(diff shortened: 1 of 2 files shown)


nojnhuh:
> I want to see how much of this I can run locally to make sure things are working as expected.
> /hold

I tried invoking a build through the image in cloudbuild.yaml but didn't quite get through some docker-in-docker issues that don't seem to occur in Prow. I could at least see the chart and image builds invoked with the expected tags, so I feel ready to try this in Prow.

/hold cancel

pohly (approved):
/lgtm
/approve

pohly:
https://github.com/kubernetes-sigs/dra-example-driver/releases/tag/chart%2F0.1.2

PR #85, opened 2025-03-12
Fix image repo used in chart

nojnhuh:
This PR adds a missing `/dra-example-driver` to make the default `image.repository` value for the chart `registry.k8s.io/dra-example-driver/dra-example-driver` which will be the correct name of the image that ultimately gets promoted.

(I think this is the last fix before we can promote the image and chart to prod, but I shouldn't make any promises.)


pohly (approved):
/lgtm
/approve

pohly:
-> https://github.com/kubernetes-sigs/dra-example-driver/releases/tag/chart%2F0.1.3

Issue #64, opened 2024-09-23
Improve the experience for using this in e2e tests


kannon92:
### Motivation

In Kueue, we have discussed using this repo has a way to add testing for DRA support.  There are a few problems with using this in e2e tests.

#### Problem 1

There is no release branch corresponding to each release so this will likely break if one tries to pull the main after a while. For Kueue, we would like to have testing for each k8s release (1.29, 1.30 and 1.31) and I expect this will fail when we transition to beta.

We can focus on supporting only 1.30 and 1.31 for now.

#### Problem 2

If one wanted to consume this in a e2e test, they must clone the library, build the driver, create a cluster with the DRA feature gate and helm install the driver into the cluster. Maybe as we publish release branches for this, we could consider publishing a chart or at least an image so other projects don't have the build the drivers on every test run.

nojnhuh:
I'm also interested in using the example driver in some new documentation (https://github.com/kubernetes/website/pull/49079), where having a `kubectl apply` or `helm install` command would save the new docs from having to recreate the build steps listed in this repo.

nojnhuh:
I'm happy to pick this up if I could get some pointers as to what the preferred hosting provider for a Helm chart would be. Using a GitHub Pages site or raw artifacts in GitHub has worked well enough for me before, but open to suggestions. @klueska @pohly Is that a sensible option here?

pohly:
Kubernetes publishes images in its own registry. We should do the same for the example driver. Perhaps the image publishing process for CSI sidecars can be adapted, see https://github.com/kubernetes-csi/csi-release-tools/blob/master/SIDECAR_RELEASE_PROCESS.md#release-process, in particular step 11.

I don't think we have any central location for Helm charts. What you suggested sounds fine to me.

nojnhuh:
Sounds good, I'll get started on that. Thanks!

/assign

nojnhuh:
PR open to create the staging repo for container images: https://github.com/kubernetes/k8s.io/pull/7703

nojnhuh:
> I'm happy to pick this up if I could get some pointers as to what the preferred hosting provider for a Helm chart would be. Using a GitHub Pages site or raw artifacts in GitHub has worked well enough for me before, but open to suggestions. 

[...]

PR #87, opened 2025-03-24
Update RELEASE.md

nojnhuh:
This PR updates RELEASE.md to reflect the process followed when cutting the first release for #64.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,9 +1,43 @@
 # Release Process
 
-The Kubernetes Template Project is released on an as-needed basis. The process is as follows:
+The DRA example driver is released on an as-needed basis. Published release
+artifacts include:
+
+- The dra-example-driver Helm chart
+- Container images
+
+The Helm chart may be released independently from the container images. When
+releasing new container images, a new release of the Helm chart should usually
+be cut at the same time.
+
+The process for both is as follows:
 
 1. An issue is proposing a new release with a changelog since the last release
-1. All [OWNERS](OWNERS) must LGTM this release
-1. An OWNER runs `git tag -s $VERSION` and inserts the changelog and pushes the tag with `git push $VERSION`
+1. At least two of the [OWNERS](OWNERS) must LGTM this release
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -1,9 +1,43 @@
...
+      [SemVer], e.g. `v0.1.0`
+    - When releasing the Helm chart, `$VERSION` is a `chart/`-prefixed [SemVer],
+      e.g. `chart/0.1.0`
+    - The same commit may be tagged with a tag of each form to release both the
+      Helm chart and container images at the same time
+    - Each pushed tag triggers builds of artifacts pushed to the staging
+      repository for the [dra-example-driver container image][container-staging]
+      and [Helm chart][chart-staging]
+1. A PR updating the [registry.k8s.io image list][image-list] is opened with the
+   SHA digests and tags of the new artifacts from the staging repo
```
nojnhuh, line 31: I'm hopeful this can be better automated in the future, but there are some kinks to get something up and running at the moment.

[`kpromo`](https://github.com/kubernetes-sigs/promo-tools/blob/main/docs/promotion-pull-requests.md) is hardcoded to expect images are staged at `gcr.io/k8s-staging-<project>` and doesn't work with the new way staging repos have been getting setup: https://github.com/kubernetes-sigs/promo-tools/issues/1512

I also looked at @pohly's [SIG Storage script](https://github.com/kubernetes/k8s.io/blob/61631f4b7fdd6c44058606e5517728deb0efbffc/registry.k8s.io/images/k8s-staging-sig-storage/generate.sh) but our staging repo is in an unfortunate spot right now where a `v0.1.0` tag exists in the staging repo but isn't the same as the `v0.1.0` tag that got promoted, so our version of that script would have to account for that until that image gets cleaned up in ~2 months. I'm thinking the volume of dra-example-driver releases until then will be low enough that this might be more worth revisiting then.

pohly (commented):
/lgtm

I'll leave the approve to @klueska for a second opinion.

PR #90, opened 2025-04-03
Encapsulate Checkpoint internal state

bg-chun:
This PR encapsulates the internal state of Checkpoint and prevents direct access to CheckpointV1 from DeviceState.


cmd/dra-example-kubeletplugin/state.go
```diff
@@ -119,10 +105,9 @@ func (s *DeviceState) Prepare(claim *resourceapi.ResourceClaim) ([]*drapbv1.Devi
 	if err := s.checkpointManager.GetCheckpoint(DriverPluginCheckpointFile, checkpoint); err != nil {
 		return nil, fmt.Errorf("unable to sync from checkpoint: %v", err)
 	}
-	preparedClaims := checkpoint.V1.PreparedClaims
 
-	if preparedClaims[claimUID] != nil {
-		return preparedClaims[claimUID].GetDevices(), nil
+	if exist := checkpoint.GetPreparedDevices(claimUID); exist != nil {
```
nojnhuh, line 109: nit: Could we call this `existingDevices` or something like that? `exist` is a good name for a boolean, but since this isn't a bool then `exist != nil` looks a little more confusing to me than something like `existingDevices != nil`.

nojnhuh (commented):
I left a couple nits but definitely a positive change overall I think. Thanks!

cmd/dra-example-kubeletplugin/checkpoint.go
```diff
@@ -3,14 +3,19 @@ package main
...
 )
 
+type PreparedClaims map[string]PreparedDevices
+
 type Checkpoint struct {
 	Checksum checksum.Checksum `json:"checksum"`
 	V1       *CheckpointV1     `json:"v1,omitempty"`
 }
 
+var _ checkpointmanager.Checkpoint = (*Checkpoint)(nil)
```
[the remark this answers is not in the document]
bg-furiosa: True, it’s not strictly necessary. But it helps readers new to the code understand which interface checkpoint must implement.

cmd/dra-example-kubeletplugin/state.go
bg-furiosa: updated to reuse preparedDevices in 96e98a65d26b64497ce0d77fb8c344e951f6a17d

cmd/dra-example-kubeletplugin/checkpoint.go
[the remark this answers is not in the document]
bg-chun: True, it’s not strictly necessary. But it helps readers new to the code understand which interface checkpoint must implement.

cmd/dra-example-kubeletplugin/state.go
bg-chun: updated to reuse preparedDevices in 96e98a65d26b64497ce0d77fb8c344e951f6a17d

cmd/dra-example-kubeletplugin/checkpoint.go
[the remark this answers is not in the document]
pohly: I agree that this interface assertions can help readers of the code.

I think `= &Checkpoint{}` is a bit more idiomatic.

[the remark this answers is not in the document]
bg-chun: updated in e14b2a7ace8d1afe721912f64302d58cb88bcfbf

bg-chun:
commits has been squashed

nojnhuh (approved):
> commits has been squashed

Thanks!

/lgtm

cmd/dra-example-kubeletplugin/checkpoint.go
```diff
@@ -3,14 +3,19 @@ package main
 import (
 	"encoding/json"
 
+	"k8s.io/kubernetes/pkg/kubelet/checkpointmanager"
```
pohly, line 6: `k8s.io/kubernetes` should not be imported. Any code inside is considered internal and not meant for public consumption. There are some exceptions (most notably the scheduler framework for building custom schedulers), but not this one here.

It's not even a particularly good package. We had huge issues with figuring out how checksumming was meant to be used and what the purpose of checksumming was in the first place.

Can we perhaps use this opportunity to drop the dependency?

bg-chun: cc @pohly 
> Can we perhaps use this opportunity to drop the dependency?

=> In terms of dependency, I'm on the same page. I will update PR to introduce simple checkpoint util to drop the dependency.
But for checksum do you mean we don't need checksum here?  Seems there was some issues with dra_manager_state in the past. Or do you mean we need well designed checkpoint implementation along with checksum. Since it is example dra driver, maybe simple checkpointing without checksum is fine.

pohly: The only purpose of checksumming that I could imagine is to detect bit flips in the file. As a DRA driver author, is that important to you on top of whatever potential checksumming and error correcting the OS might do?

Are there other reasons for it?

bg-chun: I haven’t thought about it seriously.
The checkpoint manager already existed when I was involved with resource managers in kubelet around 2019.

Seems, checksum is originated from PodSandbox checkpointer of dockershim.
- https://github.com/kubernetes/kubernetes/commit/d62bd9ef6571c13b134162da2f7ae41785c6a59e#diff-7cf7a43176686c06054024ad9d7226062787c188c66fd17d837d7e821b164ab0

Approaching conservatively, there seem to be a few possible cases:
- The process could restart while writing a file (e.g., OOM, kill, restart).
- The file might be corrupted unintentionally by a person or script.

pohly: > The process could restart while writing a file (e.g., OOM, kill, restart).

The usual approach is to write a temp file, sync, then rename. But I suppose a checksum is easier.

> The file might be corrupted unintentionally by a person or script.

True, albeit a bit unlikely.

pohly: > I will update PR to introduce simple checkpoint util to drop the dependency.

Gentle reminder that this is pending.

/lgtm cancel

bg-chun: cc @pohly 
I will resume this PR soon. Before I do, I have a couple of quick questions:

1. My understanding from your feedback is that checksums might be unnecessary here. Would implementing a simple checkpointing mechanism without checksums address this concern?

2. Aside from that, do you have any additional suggestions for an ideal checkpointing mechanism?

pohly: I'm fine without a checkpoint checksum. For the "writing file fails" case I think the "write tmp file, sync, close, remove, rename" approach would be useful. I don't know about other existing mechanisms that could be used here.

cc @nojnhuh

pohly: @nojnhuh: can you perhaps take over from @bg-chun?

He said that he's not going to be able to finish.

nojnhuh: Yep, I'll pick this up.

/remove-lifecycle rotten

nojnhuh: Finally got back to this in #148, let's follow up there.

/close

PR #93, opened 2025-04-09
WIP: Add instructions for running the example driver on GKE

mortent:
We currently only provide instructions for running the example driver on Kind. We should provide instructions for how to run the driver on other Kubernetes distributions, as they will include some different steps. This PR adds instructions for how to get the example driver running on GKE.

For now this just adds a new section with the new content. Ideally we should restructure the README file, so that we can provide separate instructions for getting the driver running on a cluster, and then a shared section for running the examples. I decided not to do that in this PR, as I want to make sure we have agreement on a structure first. Created an issue for this: https://github.com/kubernetes-sigs/dra-example-driver/issues/94

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,6 +383,49 @@ Finally, you can run the following to cleanup your environment and delete the
 ./demo/delete-cluster.sh
 ```
 
+## Installing the example driver on a GKE cluster
+It is also possible to run the example driver on a GKE cluster. For this, we
+will use the pre-built image for the kubelet plugin, so there is no need
+to build anything. All that is needed is a Google Cloud Platform account,
+the gcloud CLI and Helm.
+
+To keep things simple and identical to the Kind example, we will use a
+single-node GKE cluster.
+
+CDI must be enabled in containerd for the DRA driver to work. CDI is
+enabled by default in GKE since 1.32.1-gke.1489001, so we will create
+a cluster in the rapid channel to make sure we get a recent version.
+
+Since DRA is still a beta feature, we need to explicitely enable it
+when the cluster is created.
+
+First, create a GKE cluster with gcloud.
+```bash
+gcloud container clusters create dra-example-driver-cluster \
+--location=us-central1-c \
+--release-channel=rapid \
+--num-nodes=1 \
+--enable-kubernetes-unstable-apis=resource.k8s.io/v1beta1/deviceclasses,resource.k8s.io/v1beta1/resourceclaims,resource.k8s.io/v1beta1/resourceclaimtemplates,resource.k8s.io/v1beta1/resourceslices
+```
+
+Once the cluster is ready, we can install the DRA using Helm.
+
+The kubelet plugin in the example driver is set up to run with priority class
+`system-node-critical`. On GKE, pods are by default restricted from running
+with this priority class, so we need to use a ResourceQuota to allow it. The
+Helm chart supports, this, we just have to enable it.
+
+```bash
+helm upgrade -i \
+  --create-namespace \
+  --namespace dra-example-driver \
+  --set=resourcequota.enabled=true \
+  dra-example-driver \
+  deployments/helm/dra-example-driver
+```
+
+The examples in `demo/gpu-test{1,2,3,4,5}.yaml` works just like with Kind.
+
 ## Anatomy of a DRA resource driver
 
 TBD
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -383,6 +383,49 @@ Finally, you can run the following to cleanup your environment and delete the
...
+the gcloud CLI and Helm.
+
+To keep things simple and identical to the Kind example, we will use a
+single-node GKE cluster.
+
+CDI must be enabled in containerd for the DRA driver to work. CDI is
+enabled by default in GKE since 1.32.1-gke.1489001, so we will create
+a cluster in the rapid channel to make sure we get a recent version.
+
+Since DRA is still a beta feature, we need to explicitely enable it
```
nojnhuh, line 399:
```suggestion
Since DRA is still a beta feature, we need to explicitly enable it
```

README.md
```diff
@@ -383,6 +383,49 @@ Finally, you can run the following to cleanup your environment and delete the
...
+```bash
+helm upgrade -i \
+  --create-namespace \
+  --namespace dra-example-driver \
+  --set=resourcequota.enabled=true \
+  dra-example-driver \
+  deployments/helm/dra-example-driver
+```
+
+The examples in `demo/gpu-test{1,2,3,4,5}.yaml` works just like with Kind.
```
nojnhuh, line 427:
```suggestion
The examples in `demo/gpu-test{1,2,3,4,5}.yaml` work just like with Kind.
```

README.md
```diff
@@ -383,6 +383,49 @@ Finally, you can run the following to cleanup your environment and delete the
 ./demo/delete-cluster.sh
 ```
 
+## Installing the example driver on a GKE cluster
```
nojnhuh, line 386: It would be great to also run the e2e test on GKE to make sure we don't accidentally break this. I'm not sure if there's an existing pattern to do that in Prow or if GitHub Actions makes that easy. We can definitely address that later.

README.md
```diff
@@ -383,6 +383,49 @@ Finally, you can run the following to cleanup your environment and delete the
...
+with this priority class, so we need to use a ResourceQuota to allow it. The
+Helm chart supports, this, we just have to enable it.
+
+```bash
+helm upgrade -i \
+  --create-namespace \
+  --namespace dra-example-driver \
+  --set=resourcequota.enabled=true \
+  dra-example-driver \
+  deployments/helm/dra-example-driver
```
nojnhuh, line 424: Using the chart from the main branch might not always be compatible with the latest released version of the container image. Ideally we'd stick to either building the image from source for the checked out version of the chart (like the e2e CI tests do), or use compatible published release versions of each. I see published releases won't work here immediately without these unreleased changes to the chart, but we could merge the chart changes, cut a chart release, then merge the docs.

Seamless upgrade support for 1.33 is one thing I'm anticipating which will involve changes to both the image and the chart. Installing a chart that is set up for seamless upgrades and an image that isn't will likely cause issues, though that might only be in marginal cases users walking through the demo wouldn't normally hit. I haven't thought that particular scenario all the way through. For sensitive changes like that I'm confident we can work things out such that a little bit of skew is probably fine, but zero skew is of course preferred.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -87,3 +87,14 @@ webhook:
     # The name of the service account to use.
     # If not set and create is true, a name is generated using the fullname template
     name: ""
+
+resourcequota:
+  enabled: false
```
nojnhuh, line 92: I'd vote to enable this for e2e tests for that extra bit of coverage like we do for the webhook which is disabled by default: https://github.com/kubernetes-sigs/dra-example-driver/blob/e82f2915b7af8ade4124f255ec9d847c343ef565/test/e2e/setup-e2e.sh#L36

nojnhuh (commented):
I don't have a GCP account where I could test this myself, but the docs look good overall, thanks! I left a few small comments inline.

pohly:
@mortent: please address the comments and rebase.

pohly:
/retitle WIP: Add instructions for running the example driver on GKE

pohly:
/reopen

@mortent: do you want to hand this over to someone else? The PR still seems relevant.

/cc @vishalanarase

vishalanarase (commented):
I've started working on this and will raise a PR by tomorrow.

vishalanarase:
We can close this PR, because I have raised PR and it was merged.
https://github.com/kubernetes-sigs/dra-example-driver/pull/177
/close

PR #95, opened 2025-04-14
Bump k8s.io dependencies to 1.33

nojnhuh:
This PR bumps k8s.io dependencies to v1.33.0 and handles the collateral damage associated with that:
- Bump Go to 1.24
- Adopt the DRA Helper: https://github.com/kubernetes/kubernetes/pull/130700

~~There's an additional commit here to bump the k8s version used in CI to v1.33.0-rc.0 which I plan to drop once v1.33.0 is available.~~

Commit is dropped!


deployments/helm/dra-example-driver/Chart.yaml
```diff
@@ -25,4 +25,4 @@ version: 0.0.0-dev
 # It is recommended to use it with quotes.
 appVersion: "v0.1.0"
 
-kubeVersion: "1.32.x"
+kubeVersion: ">=1.32.0-0"
```
nojnhuh, line 28: The e2e test passes for me with a 1.32 cluster still, so I didn't opt to make 1.33 a minimum here. Even when the driver can handle seamless upgrades, it might be possible to maintain compatibility with 1.32 if we allow users to disable the upgrade-related changes like `maxSurge`.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -49,87 +49,103 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
 		ctx,
 		claim.Name,
 		metav1.GetOptions{})
 	if err != nil {
-		return &drapbv1.NodePrepareResourceResponse{
-			Error: fmt.Sprintf("failed to fetch ResourceClaim %s in namespace %s", claim.Name, claim.Namespace),
+		return kubeletplugin.PrepareResult{
+			Err: fmt.Errorf("failed to fetch ResourceClaim %s in namespace %s: %w", claim.Name, claim.Namespace, err),
 		}
 	}
```
bg-chun, line 112: Do we still need this? The helper ensures presence of claim in the given namespace.
- https://github.com/kubernetes/kubernetes/blob/b53b9fb5573323484af9a19cf3f5bfe80760abba/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L731

nojnhuh: It looks like we don't. I missed that the helper already does this for us. Thanks!

nojnhuh:
I updated this with the final v1.33.0 dependencies, so this should be ready for review.

/retitle Bump k8s.io dependencies to 1.33

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -49,87 +48,93 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
+	result := make(map[types.UID]kubeletplugin.PrepareResult)
 
-	for _, claim := range req.Claims {
-		preparedResources.Claims[claim.UID] = d.nodePrepareResource(ctx, claim)
+	for _, claim := range claims {
+		result[claim.UID] = d.prepareResourceClaim(ctx, claim)
 	}
 
-	return preparedResources, nil
+	return result, nil
```
aojea, line 99: does this never return error now?

aojea: nvm, it didn't return before

jackfrancis (approved):
/lgtm

/assign @pohly

to sign off on the changes to fulfill the modified DRAPlugin interface

also cc @elezar who collaborated on reviewing those changes

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -33,9 +33,7 @@ import (
 )
 
 const (
-	PluginRegistrationPath     = "/var/lib/kubelet/plugins_registry/" + consts.DriverName + ".sock"
 	DriverPluginPath           = "/var/lib/kubelet/plugins/" + consts.DriverName
```
pohly, line 36: It's good practice to have a configuration option for `/var/lib/kubelet/plugins` and `/var/lib/kubelet/plugins_registry`. That makes a DRA driver usable on Kubernetes distros with unusual paths.

This wasn't done before, so this can also be enhanced in a follow-up PR.

pohly (approved):
/approve

One item to follow up on, but that should better be a separate PR.

PR #98, opened 2025-05-06
Make kubelet plugin dirs configurable

nojnhuh:
Follow-up from https://github.com/kubernetes-sigs/dra-example-driver/pull/95#discussion_r2073367665

This PR adds two new values to the Helm chart to configure paths needed to initialize kubelet plugins in case a cluster uses non-default paths for kubelet data.

I tested locally with these changes to use a non-default value for each:
```diff
diff --git a/demo/scripts/kind-cluster-config.yaml b/demo/scripts/kind-cluster-config.yaml
index 997c306..04cfd13 100644
--- a/demo/scripts/kind-cluster-config.yaml
+++ b/demo/scripts/kind-cluster-config.yaml
@@ -34,3 +34,4 @@ nodes:
     nodeRegistration:
       kubeletExtraArgs:
         v: "1"
+        root-dir: /var/lib/kubelet2
diff --git a/test/e2e/setup-e2e.sh b/test/e2e/setup-e2e.sh
index c45fddf..bcc5b1a 100755
--- a/test/e2e/setup-e2e.sh
+++ b/test/e2e/setup-e2e.sh
@@ -34,5 +34,7 @@ helm upgrade -i \
   --create-namespace \
   --namespace dra-example-driver \
   --set webhook.enabled=true \
+  --set kubeletPlugin.kubeletRegistrarDirectoryPath=/var/lib/kubelet2/plugins_registry \
+  --set kubeletPlugin.kubeletPluginsDirectoryPath=/var/lib/kubelet2/plugins \
   dra-example-driver \
   deployments/helm/dra-example-driver
```

### DIFF
--- a/deployments/helm/dra-example-driver/values.yaml
+++ b/deployments/helm/dra-example-driver/values.yaml
@@ -53,6 +53,8 @@ kubeletPlugin:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  kubeletRegistrarDirectoryPath: /var/lib/kubelet/plugins_registry
+  kubeletPluginsDirectoryPath: /var/lib/kubelet/plugins
   containers:
     init:
       securityContext: {}
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -54,6 +54,8 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
 		kubeletplugin.KubeClient(config.coreclient),
 		kubeletplugin.NodeName(config.flags.nodeName),
 		kubeletplugin.DriverName(consts.DriverName),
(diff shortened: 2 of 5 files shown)


deployments/helm/dra-example-driver/values.yaml
```diff
@@ -53,6 +53,8 @@ kubeletPlugin:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  kubeletRegistrarDirectoryPath: /var/lib/kubelet/plugins_registry
+  kubeletPluginsDirectoryPath: /var/lib/kubelet/plugins
```
pohly, line 57: This is the only place where a default is stored and Helm is the only supported installation method, right? Otherwise we would need similar defaults elsewhere.

Do we have documentation of values that can be set?

pohly: Right. I was wondering whether the helper package should have those constants... and it already does!

/lgtm
/approve

PR #104, opened 2025-06-03
Add livenessProbe to kubelet plugin

### LINKED ISSUE #96: Add livenessProbe to kubelet plugin DaemonSet
Adding a livenessProbe will allow the kubelet to restart an unhealthy kubelet plugin so it can re-register in the event that kubelet unregisters it because of https://github.com/kubernetes/kubernetes/pull/132058. Ideally the livenessProbe will fail based on the same criteria being evaluated by the plugin manager.

/assign

nojnhuh:
This PR adds a gRPC livenessProbe to the kubelet plugin which invokes both the registration socket's GetInfo method and the DRA socket's NodePrepareResources method with an empty list of claims.

Fixes #96

### DIFF
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -32,9 +32,10 @@ import (
 )
 
 type driver struct {
-	client coreclientset.Interface
-	helper *kubeletplugin.Helper
-	state  *DeviceState
+	client      coreclientset.Interface
+	helper      *kubeletplugin.Helper
+	state       *DeviceState
+	healthcheck *healthcheck
 }
 
 func NewDriver(ctx context.Context, config *Config) (*driver, error) {
@@ -78,6 +79,11 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
 		},
 	}
 
+	driver.healthcheck, err = startHealthcheck(ctx, config)
+	if err != nil {
+		return nil, fmt.Errorf("start healthcheck: %w", err)
+	}
+
 	if err := helper.PublishResources(ctx, resources); err != nil {
 		return nil, err
 	}
@@ -85,7 +91,10 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
 	return driver, nil
 }
 
-func (d *driver) Shutdown() error {
+func (d *driver) Shutdown(logger klog.Logger) error {
+	if d.healthcheck != nil {
+		d.healthcheck.Stop(logger)
+	}
 	d.helper.Stop()
 	return nil
 }
--- a/cmd/dra-example-kubeletplugin/health.go
+++ b/cmd/dra-example-kubeletplugin/health.go
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2025 The Kubernetes Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/url"
+	"path"
+	"strconv"
+	"sync"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/status"
+	"k8s.io/klog/v2"
+	drapb "k8s.io/kubelet/pkg/apis/dra/v1beta1"
+	registerapi "k8s.io/kubelet/pkg/apis/pluginregistration/v1"
+
+	"sigs.k8s.io/dra-example-driver/pkg/consts"
+)
+
+type healthcheck struct {
+	grpc_health_v1.UnimplementedHealthServer
+
+	server *grpc.Server
+	wg     sync.WaitGroup
+
+	regClient registerapi.RegistrationClient
+	draClient drapb.DRAPluginClient
+}
+
+func startHealthcheck(ctx context.Context, config *Config) (*healthcheck, error) {
+	log := klog.FromContext(ctx)
+
+	port := config.flags.healthcheckPort
+	if port < 0 {
+		return nil, nil
+	}
+
+	addr := net.JoinHostPort("", strconv.Itoa(port))
+	lis, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to listen for healthcheck service at %s: %w", addr, err)
+	}
+
+	regSockPath := (&url.URL{
+		Scheme: "unix",
+		// TODO: this needs to adapt when seamless upgrades
+		// are enabled and the filename includes a uid.
+		Path: path.Join(config.flags.kubeletRegistrarDirectoryPath, consts.DriverName+"-reg.sock"),
+	}).String()
+	log.Info("connecting to registration socket", "path", regSockPath)
+	regConn, err := grpc.NewClient(
+		regSockPath,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("connect to registration socket: %w", err)
+	}
+
+	draSockPath := (&url.URL{
+		Scheme: "unix",
+		Path:   path.Join(config.DriverPluginPath(), "dra.sock"),
+	}).String()
+	log.Info("connecting to DRA socket", "path", draSockPath)
+	draConn, err := grpc.NewClient(
+		draSockPath,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("connect to DRA socket: %w", err)
+	}
+
+	server := grpc.NewServer()
+	healthcheck := &healthcheck{
+		server:    server,
+		regClient: registerapi.NewRegistrationClient(regConn),
+		draClient: drapb.NewDRAPluginClient(draConn),
+	}
+	grpc_health_v1.RegisterHealthServer(server, healthcheck)
+
+	healthcheck.wg.Add(1)
+	go func() {
+		defer healthcheck.wg.Done()
+		log.Info("starting healthcheck service", "addr", lis.Addr().String())
+		if err := server.Serve(lis); err != nil {
+			log.Error(err, "failed to serve healthcheck service", "addr", addr)
+		}
+	}()
+
+	return healthcheck, nil
+}
+
+func (h *healthcheck) Stop(logger klog.Logger) {
+	if h.server != nil {
+		logger.Info("stopping healthcheck service")
+		h.server.GracefulStop()
+	}
+	h.wg.Wait()
+}
+
+// Check implements [grpc_health_v1.HealthServer].
+func (h *healthcheck) Check(ctx context.Context, req *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
+	log := klog.FromContext(ctx)
+
+	knownServices := map[string]struct{}{"": {}, "liveness": {}}
+	if _, known := knownServices[req.GetService()]; !known {
+		return nil, status.Error(codes.NotFound, "unknown service")
+	}
+
+	status := &grpc_health_v1.HealthCheckResponse{
+		Status: grpc_health_v1.HealthCheckResponse_NOT_SERVING,
+	}
+
+	info, err := h.regClient.GetInfo(ctx, &registerapi.InfoRequest{})
+	if err != nil {
+		log.Error(err, "failed to call GetInfo")
+		return status, nil
+	}
+	log.V(5).Info("Successfully invoked GetInfo", "info", info)
+
+	_, err = h.draClient.NodePrepareResources(ctx, &drapb.NodePrepareResourcesRequest{})
+	if err != nil {
+		log.Error(err, "failed to call NodePrepareResources")
+		return status, nil
+	}
+	log.V(5).Info("Successfully invoked NodePrepareResources")
+
+	status.Status = grpc_health_v1.HealthCheckResponse_SERVING
+	return status, nil
+}
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -48,6 +48,7 @@ type Flags struct {
 	numDevices                    int
 	kubeletRegistrarDirectoryPath string
 	kubeletPluginsDirectoryPath   string
+	healthcheckPort               int
 }
 
 type Config struct {
@@ -105,6 +106,13 @@ func newApp() *cli.App {
 			Value:       kubeletplugin.KubeletPluginsDir,
 			Destination: &flags.kubeletPluginsDirectoryPath,
 			EnvVars:     []string{"KUBELET_PLUGINS_DIRECTORY_PATH"},
+		},
+		&cli.IntFlag{
+			Name:        "healthcheck-port",
+			Usage:       "Port to start a gRPC healthcheck service. When positive, a literal port number. When zero, a random port is allocated. When negative, the healthcheck service is disabled.",
+			Value:       -1,
+			Destination: &flags.healthcheckPort,
+			EnvVars:     []string{"HEALTHCHECK_PORT"},
 		},
 	}
 	cliFlags = append(cliFlags, flags.kubeClientConfig.Flags()...)
@@ -180,7 +188,7 @@ func RunPlugin(ctx context.Context, config *Config) error {
 		logger.Error(err, "error from context")
 	}
 
-	err = driver.Shutdown()
+	err = driver.Shutdown(logger)
 	if err != nil {
 		logger.Error(err, "Unable to cleanly shutdown driver")
 	}
--- a/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
+++ b/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
@@ -45,6 +45,19 @@ spec:
         command: ["dra-example-kubeletplugin"]
         resources:
           {{- toYaml .Values.kubeletPlugin.containers.plugin.resources | nindent 10 }}
+        {{/*
+          A literal "0" will allocate a random port. Don't configure the probe
+          with the same literal "0" since that won't match where the service is
+          actually running.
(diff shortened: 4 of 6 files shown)


cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -78,6 +86,31 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
 	}
 
+	if port := config.flags.healthcheckPort; port > 0 {
+		addr := net.JoinHostPort("", strconv.Itoa(port))
+		lis, err := net.Listen("tcp", addr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to listen for healthcheck service at %s: %w", addr, err)
+		}
+
+		driver.healthcheckService = grpc.NewServer()
```
nojnhuh: Since the DRA [`Helper`](https://pkg.go.dev/k8s.io/dynamic-resource-allocation@v0.33.1/kubeletplugin#Helper) doesn't expose any gRPC details, here we start a new gRPC server, advertise a healthy status, and that status then is never changed.

This by itself doesn't seem very useful alongside https://github.com/kubernetes/kubernetes/pull/132058, where kubelet will remove ResourceSlices for drivers it deems unresponsive or unreachable. In that context, this new independent gRPC socket could still be perfectly responsive while the kubelet plugin service socket is not, where the livenessProbe continues to succeed while kubelet removes ResourceSlices.

Ideally, the driver's probe will start failing exactly when the kubelet also observes an unhealthy state, but I'm not sure that's feasible without implementing some probe-accessible health indicator mostly in the `Helper`.

@pohly Do you have any initial thoughts on if this kind of driver self-monitoring could/should be provided by the `Helper`? I wonder if it would make sense for the helper to start a client to itself and implement the same kind of connection monitoring as https://github.com/kubernetes/kubernetes/pull/132058.

lauralorenz: A few ideas
1. k8s liveness probe is the GetInfo rpc for the registration server already started by the helper
2. k8s liveness probe is some kinda new Healthcheck rpc spawned by the helper (as you suggest) in which the helper rolls up checks it does on its own service and registration server objects
3. k8s liveness probe is some kinda new Healthcheck rpc spawned by the driver (as in this PR) in which the driver rolls up pings to service and registration server endpoints

pohly: I think using the GetInfo call over the registration socket as liveness check makes sense. That already works because GetInfo is read-only, reliable and fast, which is what one wants for a  liveness probe. But it does not cover whether the DRA driver is really fulfilling its function.

Checking responsiveness over the service socket is harder. We would need to issue some kind of NodePrepareResources call to hit the same code paths as a "real" request. The first thing that comes to mind is using an invalid ResourceClaim name (not just non-existant, literally invalid). The expected outcome then is [an error](https://github.com/kubernetes/kubernetes/blob/2d3ca18f645edd9065c32f07f02de2ba6273ab1d/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L709).

But that then returns before reaching the more interesting [serialization](https://github.com/kubernetes/kubernetes/blob/2d3ca18f645edd9065c32f07f02de2ba6273ab1d/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L712).

Asking to prepare *no* claims is better. It's not something that the kubelet would normally do, but this is our liveness probe talking to our implementation, so we can do unusual things. We need to call it out in https://github.com/kubernetes/kubernetes/blob/2d3ca18f645edd9065c32f07f02de2ba6273ab1d/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L55-L93 to ensure that DRA driver implementers are not surprised.

This complements connection monitoring. When connection monitoring detects a problem, it cannot restart the container. The liveness probe can, which may be useful.

I don't think a separate gRPC service makes sense. That just tests that the process is still there, which is already covered by the GetInfo call.

nojnhuh: > I don't think a separate gRPC service makes sense. That just tests that the process is still there, which is already covered by the GetInfo call.

Since the Helper totally hides its gRPC server and gRPC probes expect [specific request and response messages](https://github.com/grpc/grpc/blob/90d4f1a52fdbf6fb874504cfc1ba1ab718f4fc7e/doc/health-checking.md) served on a [TCP socket](https://github.com/kubernetes/kubernetes/blob/447efd4c0f45f13d262423265a8a960666ed3778/pkg/probe/grpc/grpc.go#L61), is it possible to implement this outside the Helper without a new gRPC service?

And are you suggesting we should have probes check both the registration socket's GetInfo and also the DRA socket's NodePrepareResources, or one or the other? Would both of those checks be part of the same liveness probe?

I've updated this PR to wire up GetInfo to the existing new gRPC service. I also already had an implementation of the probe which does the same connection monitoring on the service socket as https://github.com/kubernetes/kubernetes/pull/132058 which is in previous commits in case that's interesting.

nojnhuh: Talking with Laura, I'll try swapping out the GetInfo in the probe for NodePrepareResources with an empty list of claims. lmk if there's any more tweaks to that approach that I'm not fully gleaning from your comment @pohly.

pohly: My suggestion was to probe both sockets, without introducing a new one.
- registration socket: call GetInfo
- service socket: call NodePrepareResources with no claims

But your are right, we cannot simply have kubelet call those sockets directly. We have to start serving the health interface separately. So whatever responds there should connect to the other two sockets and call GetInfo resp. NodePrepareResources.

nojnhuh: Thanks! Just pushed those changes. I've also squashed but the messy history is available at 9d515fe5404d125a1a19e704a093c991273f3ff9 for reference.

nojnhuh:
Now that I think we've settled on an overall approach I've marked this PR as ready for review.

Also since this isn't really a "basic" probe anymore :)

/retitle Add livenessProbe to kubelet plugin

cmd/dra-example-kubeletplugin/health.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	addr := net.JoinHostPort("", strconv.Itoa(port))
+	lis, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to listen for healthcheck service at %s: %w", addr, err)
+	}
+
+	// TODO: this needs to adapt when seamless upgrades
+	// are enabled and the filename includes a uid.
```
bart0sh: Ideally "k8s.io/dynamic-resource-allocation/kubeletplugin" should provide all required path APIs to avoid unnecessary code modifications if/when certain path is changed. Currently this logic is scattered over [this code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L463-L537). However, I understand that this is out of scope of this PR.

cmd/dra-example-kubeletplugin/health.go
```diff
@@ -0,0 +1,148 @@
...
+	regSockPath, err := url.JoinPath("unix:///", config.flags.kubeletRegistrarDirectoryPath, consts.DriverName+"-reg.sock")
```
bart0sh: Should it be "unix://" ? Third slash is a part of an absolute path and `config.flags.kubeletRegistrarDirectoryPath` **is** an absolute path, right?
Alternatively you can use `(&url.URL{Scheme: "unix", Path: ... }).String()`.

cmd/dra-example-kubeletplugin/health.go
```diff
@@ -0,0 +1,148 @@
...
+
+	regClient registerapi.RegistrationClient
+	draClient drapb.DRAPluginClient
+}
+
+func startHealthcheck(ctx context.Context, config *Config) (*healthcheck, error) {
+	log := klog.FromContext(ctx)
+
+	port := config.flags.healthcheckPort
+	if port == 0 {
```
bart0sh: Port 0 is a [valid option for a listener](https://pkg.go.dev/net#Listen). Should we use another default value?

cmd/dra-example-kubeletplugin/health.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil, fmt.Errorf("connect to DRA socket: %w", err)
+	}
+
+	server := grpc.NewServer()
+	healthcheck := &healthcheck{
+		server:    server,
+		regClient: registerapi.NewRegistrationClient(regConn),
+		draClient: drapb.NewDRAPluginClient(draConn),
+	}
+	grpc_health_v1.RegisterHealthServer(server, healthcheck)
```
bart0sh, line 98: As we're registering it as a HealthServer would it make sense to implement [gRPC HealthServer](https://pkg.go.dev/google.golang.org/grpc/health/grpc_health_v1#HealthServer) interface?

cmd/dra-example-kubeletplugin/health.go
nojnhuh: `url.JoinPath` swallows the leading slash on the first path element beyond the base, so this was kind of an ugly hack to compensate for that: https://go.dev/play/p/Hrrzu0qCQu0
```go
func main() {
	u, _ := url.JoinPath("unix://", "/var")
	fmt.Println(u)
}
```
```
unix://var
```

I agree that constructing a literal `url.URL` looks better than this so I've updated to do that.

cmd/dra-example-kubeletplugin/health.go
nojnhuh: 0's "pick any port" semantics probably aren't useful here since the probe is configured with a specific port, but I've updated this to be disabled when strictly less than 0 and pass through those 0 semantics in case that's useful in some other case.

cmd/dra-example-kubeletplugin/health.go
nojnhuh: Yes, that's what this does by embedding the `grpc_health_v1.UnimplementedHealthServer` type, otherwise it wouldn't compile. The kubelet only invokes `Check` AFAICT, which is the only method we override here. `List` and `Watch` are left unimplemented and will return errors indicating that if they're called.

bart0sh: I'd suggest to implement `List` as well just in case someone would want to use this API. `Watch` could be left unimplemented.

nojnhuh: I see now that we import an older version of the gRPC module which doesn't include `List` in the `HealthServer` interface: https://pkg.go.dev/google.golang.org/grpc@v1.68.1/health/grpc_health_v1#HealthServer

Since the `List` method uses defined `HealthListRequest` and `HealthListResponse` types, I don't think we can reasonably jump ahead and implement it anyway. Should I update the module in this PR to be able to do that?

bart0sh: No, you shouldn't update it just for this reason.

cmd/dra-example-kubeletplugin/health.go
```diff
@@ -0,0 +1,149 @@
...
+
+	addr := net.JoinHostPort("", strconv.Itoa(port))
+	lis, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to listen for healthcheck service at %s: %w", addr, err)
+	}
+
+	regSockPath := (&url.URL{
+		Scheme: "unix",
+		// TODO: this needs to adapt when seamless upgrades
```
lauralorenz, line 66: Do we have a way forward for this (does this need to detect that seamless upgrades was enabled?), or intended as a follow on?

nojnhuh: I think this should be doable, but might involve making some assumptions about the format of the socket paths like we're already doing here. The only extra piece of data we should need is the Pod UID, but that's easy to pass through the downward API. I've called this out specifically in #107 to make sure we don't lose track.

bart0sh: > Repeatedly crashing liveness probes and server endpoint cleanup – will we exhaust the number of sockets? Should be hard to do but at this point dra drivers are at 3 sockets each time it spawns

Unix sockets(registration and service) have pre-determined paths and don't change on restart, so can't be exhausted. 3rd socket (liveness) is a tcp socket and doesn't have any file associated with it.

nojnhuh: > Unix sockets(registration and service) have pre-determined paths and don't change on restart, so can't be exhausted. 3rd socket (liveness) is a tcp socket and doesn't have any file associated with it.

This matches what I observe when I hack around the driver to do no cleanup and let the probe restart it a bunch. The TCP socket for the liveness probe also persists across container restarts.

If I continuously create new Pods for the plugin though, I do see sockets start to pile up in `/var/lib/kubelet/plugins_registry` and `/var/lib/kubelet/plugins/gpu.example.com` and are never removed (at least after over 20 minutes). I'm running a kind cluster built off of k/k master plus kubernetes/kubernetes#132058. I see a bunch of logs like this in the kubelet referring to endpoints for deleted pods, which I'm not sure would be expected in this scenario:
```
Jun 26 21:57:56 dra-example-driver-cluster-worker kubelet[304]: W0626 21:57:56.602715     304 logging.go:55] [core] [Channel #35 SubChannel #36]grpc: addrConn.createTransport failed to connect to {Addr: "/var/lib/kubelet/plugins/gpu.example.com/dra-93c24dd0-f276-4436-a1de-357eb87008a6.sock", ServerName: "localhost", Attributes: {"<%!p(networktype.keyType=grpc.internal.transport.networktype)>": "unix" }, BalancerAttributes: {"<%!p(pickfirstleaf.managedByPickfirstKeyType={})>": "<%!p(bool=true)>" }}. Err: connection error: desc = "transport: Error while dialing: dial unix /var/lib/kubelet/plugins/gpu.example.com/dra-93c24dd0-f276-4436-a1de-357eb87008a6.sock: connect: connection refused"
```
Even after I manually delete those stale sockets, I see kubelet continue to log similar messages.

With cleanup in place, I see the old sockets get removed on shutdown but still see similar kubelet logs for old instances.

bart0sh: Thank you for the investigation!
Asked @pohly to re-open https://github.com/kubernetes/kubernetes/issues/128696 and started working on fixing it (again :)).

bart0sh: @nojnhuh 
> If I continuously create new Pods for the plugin though, I do see sockets start to pile up in /var/lib/kubelet/plugins_registry and /var/lib/kubelet/plugins/gpu.example.com and are never removed (at least after over 20 minutes).

I can't reproduce it. I only see two sockets for the plugin: `/var/lib/kubelet/plugins_registry/ gpu.example.com.sock` and ` /var/lib/kubelet/plugins/plugin.sock `and it gets deleted even if I force remove plugin pod. I guess different names are used in case of plugin upgrade. Need to investigate further.

nojnhuh: To clarify, I was only seeing that behavior when I additionally implement seamless upgrades and remove the cleanup steps from the example driver entirely to simulate a poorly-implemented driver. The changes where I was seeing this behavior are here: https://github.com/kubernetes-sigs/dra-example-driver/compare/main...nojnhuh:dra-example-driver:liveness-probe-hacking

bart0sh: Yep, I understand that. We need to check if the issue still exists when seamless upgrades are enabled.

bart0sh: > If I continuously create new Pods for the plugin though, I do see sockets start to pile up in /var/lib/kubelet/plugins_registry and /var/lib/kubelet/plugins/gpu.example.com

My wild guess is that the socket files were not removed because of [not calling driver.Shutdown](https://github.com/kubernetes-sigs/dra-example-driver/compare/main...nojnhuh:dra-example-driver:liveness-probe-hacking#diff-9086799987110fadc6ea3dbe02aef06610ab790c2b9a204e0c762141e8491609R186).

cmd/dra-example-kubeletplugin/health.go
```diff
@@ -0,0 +1,149 @@
...
+	grpc_health_v1.UnimplementedHealthServer
+
+	server *grpc.Server
+	wg     sync.WaitGroup
+
+	regClient registerapi.RegistrationClient
+	draClient drapb.DRAPluginClient
+}
+
+func startHealthcheck(ctx context.Context, config *Config) (*healthcheck, error) {
```
bart0sh, line 50: Would it make sense to move this (and possible other) APIs to `k8s.io/dynamic-resource-allocation/kubeletplugin` so that plugin authors won't have to duplicate this code in their plugins?

cmd/dra-example-kubeletplugin/health.go
nojnhuh: > My wild guess is that the socket files were not removed because of [not calling driver.Shutdown](https://github.com/kubernetes-sigs/dra-example-driver/compare/main...nojnhuh:dra-example-driver:liveness-probe-hacking#diff-9086799987110fadc6ea3dbe02aef06610ab790c2b9a204e0c762141e8491609R186).

Right, but should the kubelet be mitigating that by removing the sockets itself after some period of time?

cmd/dra-example-kubeletplugin/health.go
nojnhuh: I think that's the eventual goal. It's easier to experiment quicker here and without the risk of affecting other "real" drivers though. As soon as we determine this is a reasonable solution I think it would make sense to make this part of the `Helper`.

cmd/dra-example-kubeletplugin/health.go
bart0sh: It could be racy as Kubelet can remove a socket, which is expected to exist by the plugin. It's better to take care of this on the plugin side as sockets are created by the plugin and should be maintained by the plugin.

bart0sh: @nojnhuh 
> If I continuously create new Pods for the plugin though, I do see sockets start to pile up in /var/lib/kubelet/plugins_registry and /var/lib/kubelet/plugins/gpu.example.com 

What puzzles me here is that unix sockets are removed by the listener code when the listener is closed. How did you get them not removed/stalled? Did you force removed plugin pods?

> Even after I manually delete those stale sockets, I see kubelet continue to log similar messages.

Can you create an issue for that? I'll try to investigate.

nojnhuh: > What puzzles me here is that unix sockets are removed by the listener code when the listener is closed. How did you get them not removed/stalled? Did you force removed plugin pods?

I had run `kubectl rollout restart daemonset` a few times to create new pods.

> > Even after I manually delete those stale sockets, I see kubelet continue to log similar messages.
>
> Can you create an issue for that? I'll try to investigate.

Opened https://github.com/kubernetes/kubernetes/issues/132660

nojnhuh:
> @nojnhuh The PR looks good to me. Please squash your commits and unhold. I'll re-lgtm it.

Done, thanks!

bart0sh:
/lgtm

/assign @klueska @pohly 
for approval

nojnhuh:
> /hold for squash

I'll leave the hold in case @klueska or @pohly have any other feedback, otherwise I think this is good to go.

PR #106, opened 2025-06-19
add initContainer claim-sharing example

jackfrancis:
This PR adds a new example that demonstrates a pod's initContainer sharing a claim w/ its container.

A potential use of this example would be adding "pre-validation" to a GPU device before performing work.

### DIFF
--- a/demo/gpu-test6.yaml
+++ b/demo/gpu-test6.yaml
@@ -1,0 +1,48 @@
+# One pod: one initContainer + one container
+# Each asking for shared access to a single GPU
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gpu-test6
+
+---
+apiVersion: resource.k8s.io/v1beta1
+kind: ResourceClaimTemplate
+metadata:
+  namespace: gpu-test6
+  name: single-gpu
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        deviceClassName: gpu.example.com
+
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: gpu-test6
+  name: pod0
+spec:
+  initContainers:
+  - name: init0
+    image: ubuntu:22.04
+    command: ["bash", "-c"]
+    args: ["export"]
+    resources:
+      claims:
+      - name: shared-gpu
+  containers:
+  - name: ctr0
+    image: ubuntu:22.04
+    command: ["bash", "-c"]
+    args: ["export; trap 'exit 0' TERM; sleep 9999 & wait"]
+    resources:
+      claims:
+      - name: shared-gpu
+  resourceClaims:
+  - name: shared-gpu
+    resourceClaimTemplateName: single-gpu
--- a/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
+++ b/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
@@ -58,7 +58,7 @@ spec:
               fieldPath: metadata.namespace
         # Simulated number of devices the example driver will pretend to have.
         - name: NUM_DEVICES
-          value: "8"
+          value: "9"
         volumeMounts:
         - name: plugins-registry
           mountPath: /var/lib/kubelet/plugins_registry
--- a/test/e2e/e2e.sh
+++ b/test/e2e/e2e.sh
@@ -53,6 +53,7 @@ kubectl create -f demo/gpu-test2.yaml
 kubectl create -f demo/gpu-test3.yaml
 kubectl create -f demo/gpu-test4.yaml
 kubectl create -f demo/gpu-test5.yaml
+kubectl create -f demo/gpu-test6.yaml
 
 function gpus-from-logs {
   local logs="$1"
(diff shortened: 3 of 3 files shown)


demo/gpu-test6.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: Pod
+metadata:
+  namespace: gpu-test6
+  name: pod0
+spec:
+  initContainers:
+  - name: init0
+    image: ubuntu:22.04
+    command: ["bash", "-c"]
+    args: ["export; trap 'exit 0' TERM; sleep 1 & wait"]
```
nojnhuh: Is there value in sleeping for 1 second over not sleeping at all? i.e. only run `export`?

demo/gpu-test6.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: ResourceClaimTemplate
+metadata:
+  namespace: gpu-test6
+  name: single-gpu
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        deviceClassName: gpu.example.com
```
nojnhuh, line 21: Right now all of the examples together take up exactly the 8 GPUs exposed by default. I don't think that should block adding new examples though.

Maybe one short-term workaround is to leave the default count exposed by the driver to 8 and override that to 9 for the e2e tests so we can accommodate this new example there? https://github.com/kubernetes-sigs/dra-example-driver/blob/4a1bf91c32d7ccf43e93d7cd6dfaaf7564a7fae2/test/e2e/e2e.sh

I don't think the chart currently exposes a way to override that, so that would also need to be added. I think we can integrate this with tests later though.

@jackfrancis Could you please open an issue to add this to the e2e tests so we don't lose track of that?

demo/gpu-test6.yaml
```diff
@@ -0,0 +1,48 @@
+# One pod: one initContainer + one container
+# Each asking for shared access to a single GPU
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gpu-test6
```
nojnhuh, line 8: @klueska Is there any source material for this image where we could integrate this new example? https://github.com/kubernetes-sigs/dra-example-driver/blob/4a1bf91c32d7ccf43e93d7cd6dfaaf7564a7fae2/demo/demo-apps.png

demo/gpu-test6.yaml
jackfrancis: I suppose the value would be in better representing the work that an initContainer might do (probably not instantaneous?) But whatever best expresses and demonstrates a successful initContainer execution and then progression onto the primary container is fine, lemme know!

nojnhuh: I think a plain `export` is sufficient here. I don't think we need to simulate extra work in the init container to demonstrate the use case.

nojnhuh:
LGTM pending squash

demo/gpu-test6.yaml
jackfrancis: I'd be happy to add this to the E2E test as part of merging as well if that makes sense?

nojnhuh: That'd be great!

jackfrancis: Added E2E matter for this, and updated the hard-coded num devices to 9. I think this should literally work once we build an updated image w/ these changes?

Maybe the larger question is: is this the right way to integrate an initContainer example? Would updating one of the existing examples be easier and clearer in terms of the convenient image representation that describes the current set of six scenarios?

Happy to take whatever approach that is best for keeping the overall set of reference scenarios the cleanest for maintenance, and easiest to present to users.

test/e2e/e2e.sh
```diff
@@ -374,6 +375,62 @@ if [[ "$gpu_test5_pod0_sp_ctr1_partition_count" != "10" ]]; then
...
+fi
+gpu_test6_pod0_ctr0_timeslice_interval=$(gpu-timeslice-interval-from-logs "$gpu_test6_pod0_ctr0_logs" $(gpu-id "$gpu_test6_pod0_ctr0_gpu"))
+if [[ "$gpu_test6_pod0_ctr0_timeslice_interval" != "Default" ]]; then
+  echo "Expected Pod gpu-test6/pod0, container ctr0 to have timeslice interval Default, got $gpu_test6_pod0_ctr0_timeslice_interval"
+  exit 1
+fi
+
 # test that deletion is fast (less than the default grace period of 30s)
 # see https://github.com/kubernetes/kubernetes/issues/127188 for details
 kubectl delete -f demo/gpu-test1.yaml --timeout=25s
```
nojnhuh, line 436: We should delete the new example here too.

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -80,7 +80,7 @@ func newApp() *cli.App {
 		&cli.IntFlag{
 			Name:        "num-devices",
 			Usage:       "The number of devices to be generated.",
-			Value:       8,
+			Value:       9,
 			Destination: &flags.numDevices,
 			EnvVars:     []string{"NUM_DEVICES"},
```
nojnhuh: This environment variable gets set in the Helm chart which takes precedence over this default:
https://github.com/kubernetes-sigs/dra-example-driver/blob/4a1bf91c32d7ccf43e93d7cd6dfaaf7564a7fae2/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml#L65

test/e2e/e2e.sh
```diff
@@ -374,6 +375,62 @@ if [[ "$gpu_test5_pod0_sp_ctr1_partition_count" != "10" ]]; then
...
+gpu_test6_pod0_init0_logs=$(kubectl logs -n gpu-test6 pod0 -c init0)
+gpu_test6_pod0_init0_gpus=$(gpus-from-logs "$gpu_test6_pod0_init0_logs")
+gpu_test6_pod0_init0_gpus_count=$(echo "$gpu_test6_pod0_init0_gpus" | wc -w)
+if [[ $gpu_test6_pod0_init0_gpus_count != 1 ]]; then
+  echo "Expected Pod gpu-test6/pod0, container init0 to have 1 GPU, but got $gpu_test6_pod0_init0_gpus_count: $gpu_test6_pod0_init0_gpus"
+  exit 1
+fi
+gpu_test6_pod0_init0_gpu="$gpu_test6_pod0_init0_gpus"
+echo "Pod gpu-test6/pod0, container init0 claimed $gpu_test6_pod0_init0_gpu"
+if [[ "$gpu_test6_pod0_init0_gpu" != "$gpu_test6_pod0_init0_gpu" ]]; then
```
nojnhuh:
```suggestion
if [[ "$gpu_test6_pod0_init0_gpu" != "$gpu_test6_pod0_ctr0_gpu" ]]; then
```

We'll have to move this further down to after we've set `gpu_test6_pod0_ctr0_gpu`

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
```diff
@@ -58,7 +58,7 @@ spec:
               fieldPath: metadata.namespace
         # Simulated number of devices the example driver will pretend to have.
         - name: NUM_DEVICES
-          value: "8"
+          value: "9"
```
nojnhuh, line 61: If changing the default here means we can skip changing the default for the command line flag in the binary then I'd say let's keep the default of 8 in the binary just to avoid any churn there that we can. We could do that as a follow-up.

nojnhuh (commented):
/lgtm

I left one more non-blocking comment in case you'd like to address that now, otherwise I think this is ready for squash.

nojnhuh (approved):
/lgtm
/approve

Thanks!

PR #109, opened 2025-06-26
Update kubelet plugin signal handling on shutdown

nojnhuh:
This PR supersedes #68 to update how shutdown signals are handled by the kubelet plugin. The main difference from that PR is the use of [`signal.NotifyContext`](https://pkg.go.dev/os/signal#NotifyContext) to tie the signal handler to a `Context` instead of managing a channel directly. I haven't observed any functional differences between this PR and either `main` or #68.

Closes #68

### DIFF
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -133,14 +134,16 @@ func newApp() *cli.App {
 				coreclient: clientSets.Core,
 			}
 
-			return StartPlugin(ctx, config)
+			return RunPlugin(ctx, config)
 		},
 	}
 
 	return app
 }
 
-func StartPlugin(ctx context.Context, config *Config) error {
+func RunPlugin(ctx context.Context, config *Config) error {
+	logger := klog.FromContext(ctx)
+
 	err := os.MkdirAll(config.DriverPluginPath(), 0750)
 	if err != nil {
 		return err
@@ -159,18 +162,27 @@ func StartPlugin(ctx context.Context, config *Config) error {
 		return fmt.Errorf("path for cdi file generation is not a directory: '%v'", err)
 	}
 
+	ctx, stop := signal.NotifyContext(ctx, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	defer stop()
+
 	driver, err := NewDriver(ctx, config)
 	if err != nil {
 		return err
 	}
 
-	sigc := make(chan os.Signal, 1)
-	signal.Notify(sigc, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
-	<-sigc
+	<-ctx.Done()
+	// restore default signal behavior as soon as possible in case graceful
+	// shutdown gets stuck.
+	stop()
+	if err := ctx.Err(); err != nil && !errors.Is(err, context.Canceled) {
+		// A canceled context is the normal case here when the process receives
+		// a signal. Only log the error for more interesting cases.
+		logger.Error(err, "error from context")
+	}
 
-	err = driver.Shutdown(ctx)
+	err = driver.Shutdown()
 	if err != nil {
-		klog.FromContext(ctx).Error(err, "Unable to cleanly shutdown driver")
+		logger.Error(err, "Unable to cleanly shutdown driver")
 	}
 
 	return nil
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -85,7 +85,7 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
 	return driver, nil
 }
 
-func (d *driver) Shutdown(ctx context.Context) error {
+func (d *driver) Shutdown() error {
 	d.helper.Stop()
 	return nil
 }


cmd/dra-example-kubeletplugin/main.go
```diff
@@ -141,6 +142,8 @@ func newApp() *cli.App {
 }
 
 func StartPlugin(ctx context.Context, config *Config) error {
+	log := klog.FromContext(ctx)
```
pohly: It's odd to have a function called "Start" which blocks. Can we renamed to "Run"?

The functions also, with and without this change, ignores cancellation of the context that it is passed. If the context is only for passing a logger, then let's pass a logger instead of the context. Or honor cancellation...

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -159,18 +162,32 @@ func StartPlugin(ctx context.Context, config *Config) error {
...
+		defer stop()
+
+		driver, err = NewDriver(ctx, config)
+		if err != nil {
+			return err
+		}
 
-	sigc := make(chan os.Signal, 1)
-	signal.Notify(sigc, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
-	<-sigc
```
pohly, line 169: My comment from https://github.com/kubernetes-sigs/dra-example-driver/pull/68/files#r1882041569 still applies: why does different signal handling help with ResourceSlice removal, which was given as justification for that PR?

cc @klueska

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -159,18 +162,32 @@ func StartPlugin(ctx context.Context, config *Config) error {
...
-	driver, err := NewDriver(ctx, config)
-	if err != nil {
-		return err
-	}
+	var driver *driver
+	{
+		// A new context only for NewDriver. We don't want to use an
+		// already-done context after that finishes.
+		ctx, stop := signal.NotifyContext(ctx, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+		defer stop()
```
pohly:
```suggestion
		defer stop()
		ctx = klog.NewContext(ctx, logger)
```

cmd/dra-example-kubeletplugin/main.go
pohly:
```suggestion
	logger := klog.FromContext(ctx)
```

pohly: That's what we normally call a `logr.Logger` instance.

cmd/dra-example-kubeletplugin/main.go
nojnhuh: TBH I'm also still confused about what exact scenario the original PR was targeting.

cmd/dra-example-kubeletplugin/main.go
nojnhuh: Can you expand on what this should look like in order to honor cancellation of the context? I tried adding this change to invoke `StartPlugin` with a cancelled context and I can see the container crash loop with `Error: start ResourceSlice controller: create controller: sync ResourceSlice informer: context canceled`. Is that the desired behavior here, i.e. that if the parent context is cancelled, then `StartPlugin` exits?

```go
ctx, cancel := context.WithCancel(ctx)
cancel()

return StartPlugin(ctx, config)
```

I also tried cancelling the context passed to `StartPlugin` in a background goroutine after the plugin starts running and that also leads down the graceful shutdown route.

cmd/dra-example-kubeletplugin/main.go
nojnhuh: Does this only add the logger to the context? Is the parent's logger not inherited?

pohly: Please ignore my comment. The `NotifyContext` parent context already has the logger.

cmd/dra-example-kubeletplugin/main.go
pohly: > Can you expand on what this should look like in order to honor cancellation of the context?

I think using this `ctx` in `signal.NotifyContext` adds support for cancellation via the context. My comment must have been based on the original version of `StartPlugin`, which uses `signal.Notify` and then blocks on `<-sigc` - that ignores the cancellation of `ctx`.

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -159,18 +162,32 @@ func StartPlugin(ctx context.Context, config *Config) error {
...
 	}
 
-	driver, err := NewDriver(ctx, config)
-	if err != nil {
-		return err
-	}
+	var driver *driver
+	{
+		// A new context only for NewDriver. We don't want to use an
+		// already-done context after that finishes.
```
pohly: Does this comment still make sense?

I have the nagging feeling that this code at some point used `signal.NotifyContext(context.Background(), ...`, but I don't see it in the PR history.

pohly: I think I get it: what this says that if a signal is received, we want to call `driver.Shutdown(ctx)` below with a context that isn't canceled.

But what if the `ctx` passed to `RunPlugin` gets canceled? Then we still end up calling `driver.Shutdown(ctx)` with a canceled context.

We could replace  `driver.Shutdown(ctx)`:

```
// Ignore parent context cancellation and instead give driver.Shutdown some (arbitrarily chosen)
// time to finish.
ctx, cancel := context.WithTimeout(context.WithoutCancellation(ctx), 5 * time.Minute)
defer cancel()
err := driver.Shutdown(ctx)
...
```

If we do that, then the local scope here isn't needed anymore.

nojnhuh: Makes sense. Actually, since the context isn't actually used at all I think I'll take it out of `Shutdown` entirely. #109 adds some logs that derive a logger from the context, but we could just pass the logger instead.

nojnhuh:
squashed! Needs one more LGTM.

/hold cancel

Issue #97, opened 2025-04-29
Add support for DRAAdminAccess


ritazh:
Add support and demo for the [DRA Admin Access](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#admin-access) feature in Kubernetes.

PR #112, opened 2025-07-21
DRAAdminAccess: add example

ritazh:
Fixes: #97 

- Add support for the DRAAdminAccess feature
- Add `resource.k8s.io/v1beta2=true` to kind config
- Update cdi handler to inject host hardware information


cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -84,3 +87,181 @@ func hash(s string) int64 {
...
+	NetworkInfo string
+	StorageInfo string
+}
+
+// GetHostHardwareInfo gathers hardware information from the host system.
+func GetHostHardwareInfo() (*HostHardwareInfo, error) {
+	info := &HostHardwareInfo{}
+
+	// Get CPU information
+	if cpuInfo, err := readFile("/proc/cpuinfo"); err == nil {
```
nojnhuh: Since all the rest of the data coming out of the driver is entirely OS- and hardware-agnostic AFAIK, could we follow that here? Even if admin access is implemented here as simply as adding only `DRA_ADMIN_ACCESS=true`, I think that's enough to demonstrate the end-to-end flow. And it doesn't assume any details about the underlying OS or hardware, which I think is an important part of making the example driver usable in as many places as possible.

e.g. @lauralorenz just brought up the question of whether the example driver works on Windows. Without this change the answer is "maybe?" but reading `/proc` would make that a more conclusive "no." I know the driver wouldn't fail and we'd only get the "unavailable" value in that case, but I'd rather avoid any evidence of assumptions like that if possible.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -101,6 +101,9 @@ func (d *driver) Shutdown(logger klog.Logger) error {
 
 func (d *driver) PrepareResourceClaims(ctx context.Context, claims []*resourceapi.ResourceClaim) (map[types.UID]kubeletplugin.PrepareResult, error) {
 	klog.Infof("PrepareResourceClaims is called: number of claims: %d", len(claims))
+	for i, claim := range claims {
+		klog.Infof("Claim %d: UID=%s, Namespace=%s, Name=%s", i, claim.UID, claim.Namespace, claim.Name)
```
nojnhuh: Could we add this log to `prepareResourceClaim` so we don't need this extra loop?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -117,14 +126,17 @@ func (s *DeviceState) Prepare(claim *resourceapi.ResourceClaim) ([]*drapbv1.Devi
 
 	checkpoint := newCheckpoint()
 	if err := s.checkpointManager.GetCheckpoint(DriverPluginCheckpointFile, checkpoint); err != nil {
-		return nil, fmt.Errorf("unable to sync from checkpoint: %v", err)
+		// Create a new checkpoint if the old one is corrupted or in a different format
```
nojnhuh: Is this necessary here because `AdminAccess` is now part of the checkpoint?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -117,14 +126,17 @@ func (s *DeviceState) Prepare(claim *resourceapi.ResourceClaim) ([]*drapbv1.Devi
...
+		checkpoint = newCheckpoint()
+		if err := s.checkpointManager.CreateCheckpoint(DriverPluginCheckpointFile, checkpoint); err != nil {
```
nojnhuh: FYI @bg-chun this might conflict with #90.

demo/test-admin-access.sh
nojnhuh, line 1: Could we integrate this with the e2e job?
https://github.com/kubernetes-sigs/dra-example-driver/blob/b399cbaa49b9d52d8266cf604e5f3f873bdd9fdd/.github/workflows/e2e.yaml

That will help make sure once we update for 1.34 that everything here is working properly.

demo/gpu-test7.yaml
```diff
@@ -0,0 +1,97 @@
...
+# Each container requesting 1 distinct GPU with admin access
+# This demo shows host hardware information exposed via environment variables
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gpu-test7
+  labels:
+    resource.k8s.io/admin-access: "true"
```
nojnhuh: Would you suggest that new users set both of these labels in anticipation of upgrading to 1.34? If so, I think we should do that here too.

demo/gpu-test7.yaml
```diff
@@ -0,0 +1,97 @@
...
+      echo "Host System Info: $HOST_SYSTEM_INFO"
+      echo "Host Network Info: $HOST_NETWORK_INFO"
+      echo "Host Storage Info: $HOST_STORAGE_INFO"
+      echo ""
+      echo "GPU Environment Variables:"
+      env | grep GPU_ | sort
+      echo ""
+      echo "=== Container will sleep for 5 minutes to allow inspection ==="
+      trap 'exit 0' TERM
+      sleep 300 & wait
```
nojnhuh: Is there a reason these only sleep for 5 minutes? The other examples sleep for 9999 seconds which I think I slightly prefer so I don't feel like I have to race against the clock and look at this right away.

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2023 The Kubernetes Authors.
```
nojnhuh: We can drop the year from the boilerplate in all of the new files.
```suggestion
 * Copyright The Kubernetes Authors.
```

demo/scripts/kind-cluster-config.yaml
```diff
@@ -15,7 +16,7 @@ nodes:
     kind: ClusterConfiguration
     apiServer:
       extraArgs:
-        runtime-config: "resource.k8s.io/v1beta1=true"
+        runtime-config: "resource.k8s.io/v1beta1=true,resource.k8s.io/v1beta2=true"
```
nojnhuh: Is v1beta2 still required here or can we drop this? It looks like the new manifests all use v1.

cmd/dra-example-kubeletplugin/cdi.go
```diff
@@ -99,9 +101,25 @@ func (cdi *CDIHandler) CreateClaimSpecFile(claimUID string, devices profiles.Pre
 			ContainerEdits: &cdispec.ContainerEdits{
 				Env: []string{
 					fmt.Sprintf("%s_DEVICE_%s_RESOURCE_CLAIM=%s", strings.ToUpper(cdi.class), deviceEnvKey, claimUID),
+					fmt.Sprintf("DRA_ADMIN_ACCESS=%t", device.AdminAccess),
 				},
 			},
 		}
+
+		// If this device has admin access, inject OS-agnostic host hardware information
```
nojnhuh: I'm still wondering if the `DRA_ADMIN_ACCESS` environment variable is a good enough demonstration of how this works without having to introduce and maintain the concept of "host hardware info" at all.

If someone is looking at this trying to figure out where to plug in their admin access-specific container edits, I think that's good enough. If getting hardware details is a common use case for admin access, I think I'd rather explain that in a comment here than actual code.

demo/test-admin-access.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
nojnhuh: The other bash scripts have the copyright boilerplate, so let's add that here too (without the year).

demo/gpu-test7.yaml
```diff
@@ -0,0 +1,105 @@
...
+  resourceClaims:
+  - name: admin-gpus
+    resourceClaimTemplateName: multiple-gpus-admin
+
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: gpu-test7
+  name: pod1
```
nojnhuh: Is this second pod substantially different from the first one? Do you think we can demonstrate admin access with just one?

test/e2e/e2e.sh
```diff
@@ -55,6 +55,7 @@ kubectl create -f demo/gpu-test3.yaml
 kubectl create -f demo/gpu-test4.yaml
 kubectl create -f demo/gpu-test5.yaml
 kubectl create -f demo/gpu-test6.yaml
+kubectl create -f demo/gpu-test7.yaml
```
nojnhuh: Is it possible that the scheduler might be working on these pods out of order such that it schedules the admin access ones first, then none of the others can be scheduled since the devices are already allocated for admin access? Should we create this example after the tests for the others have run? We could still delete them all together to show that admin access can claim already-allocated devices.

ritazh: I actually think we should move gpu-test7 earlier to show how admin access workloads not only can access already-allocated devices but also should not block future allocations. 

Per the KEP:
```
No impact on availability of claims:

The scheduler ignores claims with adminAccess: true, normal usage is not impacted as claims in other namespaces can still be allocated using the same devices that are also accessed by workloads in the admin namespace.
```

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -37,7 +37,7 @@ serviceAccount:
 kubeletPlugin:
   # numDevices describes how many GPUs to advertise on each node when the "gpu"
   # deviceProfile is used. Not relevant for other profiles.
-  numDevices: 8
+  numDevices: 9
```
ritazh, line 40: gpu-test 1-6 now require 9 GPUs

nojnhuh: I'd like to avoid having to bump this every time we add an example: #144. I also don't want simply running through the README to encounter errors, so I think this is fine for now.

demo/scripts/kind-cluster-config.yaml
```diff
@@ -15,13 +16,13 @@ nodes:
     kind: ClusterConfiguration
     apiServer:
       extraArgs:
-        runtime-config: "resource.k8s.io/v1beta1=true"
```
nojnhuh, line 18: Let's still keep v1beta1 so we can at least validate the webhooks against objects with that API version in e2e.sh.

test/e2e/e2e.sh
nojnhuh: Cool. I mostly just want to make sure that the examples will work reliably and we're not making any assumptions about the exact order things happen when all the examples are `kubectl apply`-ed all at once.

test/e2e/e2e.sh
```diff
@@ -52,6 +52,7 @@ timeout --foreground 15s bash -c verify-webhook
 kubectl create -f demo/gpu-test1.yaml
 kubectl create -f demo/gpu-test2.yaml
 kubectl create -f demo/gpu-test3.yaml
+kubectl create -f demo/gpu-test7.yaml
```
nojnhuh: Someone (probably me in a month) will inevitably try to "fix" this by putting them back in order. Could you please add a comment explaining why we're putting this in the middle?

test/e2e/e2e.sh
```diff
@@ -473,34 +492,6 @@ then
   exit 1
 fi
 
-# # Webhook should reject invalid v1beta1 resources
-if ! kubectl create --dry-run=server -f- <<'EOF' 2>&1 | grep -qF 'unknown time-slice interval'
-apiVersion: resource.k8s.io/v1beta1
```
nojnhuh, line 478: I'd still like to test this with the v1beta1 API version enabled.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -269,6 +274,20 @@ func (s *DeviceState) unprepareDevices(claimUID string, devices profiles.Prepare
...
+	hasAdminAccess := false
+	if claim != nil && claim.Status.Allocation != nil {
+		for _, result := range claim.Status.Allocation.Devices.Results {
+			if result.AdminAccess != nil && *result.AdminAccess {
+				hasAdminAccess = true
+				break
+			}
+		}
+	}
+	return hasAdminAccess
```
nojnhuh: nit to simplify this just a bit:
```suggestion
	if claim != nil && claim.Status.Allocation != nil {
		for _, result := range claim.Status.Allocation.Devices.Results {
			if result.AdminAccess != nil && *result.AdminAccess {
				return true
			}
		}
	}
	return false
```

README.md
```diff
@@ -348,10 +348,58 @@ You can use the IDs of the GPUs as well as the GPU sharing settings set in
 these environment variables to verify that they were handed out in a way
 consistent with the semantics shown in the figure above.
 
+### Demo DRA Admin Access Feature
+This example driver includes support for the [DRA AdminAccess feature](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#admin-access), which allows administrators to gain privileged access to devices already in use by other users. This example demonstrates the end-to-end flow by setting the `DRA_ADMIN_ACCESS` environment variable and potentially exposing host hardware information.
```
nojnhuh: > and potentially exposing host hardware information

Since the example driver doesn't do this, could we phrase this as something closer to "a driver managing real devices could use this to expose host hardware information"?

README.md
```diff
@@ -348,10 +348,58 @@ You can use the IDs of the GPUs as well as the GPU sharing settings set in
...
+    devices:
+      requests:
+      - name: admin-gpu
+        exactly:
+          deviceClassName: gpu.example.com
+          allocationMode: All
+          adminAccess: true
+```
+
+3. **Container**: Will receive the admin access indicator via environment variables
```
nojnhuh: Does phrasing it like this clarify the flow?

```suggestion
3. **Container**: Will receive elevated privileges from the driver, represented here as environment variables
```

Instead of the container being given "DRA admin access" it's given "elevated privileges through container edits." The container itself shouldn't have to care about DRA or the admin access feature, it should only care about its local environment (files, devices, env vars, etc.) which happens to be modified via DRA.

README.md
```diff
@@ -348,10 +348,58 @@ You can use the IDs of the GPUs as well as the GPU sharing settings set in
 these environment variables to verify that they were handed out in a way
 consistent with the semantics shown in the figure above.
 
+### Demo DRA Admin Access Feature
```
nojnhuh, line 351: It would be great to make docs like this more local to the example itself (like a README just for the example). The root README is the best place for it now but this is one thing I hope to address in #144.

nojnhuh (commented):
Thanks for the updates so far! I added a few other lightweight comments.

deployments/helm/dra-example-driver/values.yaml
ritazh: hmm I see pros and cons with this...running the example locally failed for me since the config didnt have enough devices. this works for e2e because numDevices is overridden there.

Issue #107, opened 2025-06-25
Implement seamless driver upgrades


nojnhuh:
https://github.com/kubernetes/kubernetes/pull/129832 from Kubernetes 1.33 allows DRA kubelet plugins to perform rolling upgrades without any driver downtime. [Docs added by that PR](https://github.com/kubernetes/kubernetes/blob/c1afec6a0b15ca1ed853c1321ac2c972488bf5b8/pkg/kubelet/pluginmanager/pluginwatcher/README.md#seamless-upgrade) outline the necessary changes for drivers.

There is at least one extra assumption added by #104 that needs to be revisited when implementing seamless upgrades: https://github.com/kubernetes-sigs/dra-example-driver/blob/103828c9b844b058a27d86a34744537c8242381f/cmd/dra-example-kubeletplugin/health.go#L66-L67

nojnhuh:
This was done in #113.

/close

PR #113, opened 2025-07-29
fix: support seamless upgrades with pod uid in socket path names + default helm chart values

alimaazamat:
Fixes issue #107
Socket path names in the format {baseName}{suffix} works fine for legacy deployments but if we have seamless upgrades (max surge 1 maxUnavailable 0) then it doesn't work because the socket path names include uid {baseName}-{uid}{suffix}.
Without UIDs in socket names: Two pods can't run simultaneously because they'd conflict trying to create the same socket file names. 

Solution:
Include pod UID in socket paths. Define default seamless upgrade values. Chart now defaults to seamless upgrades (max surge 1 maxUnavailable 0) which users can override by setting `kubeletPlugin.updateStrategy.rollingUpdate` values directly.

How to test:
```bash
# Build the kubelet plugin binary with the changes
make cmd-dra-example-kubeletplugin
# in demo dir and build/load image
cd demo
./scripts/build-driver-image.sh
./scripts/load-driver-image-into-kind.sh
cd ..
# made a new namespace
kubectl create namespace dra-example-driver

# Test Default Config where seamless upgrades disabled
helm install dra-example-driver deployments/helm/dra-example-driver --namespace dra-example-driver
# Check deployment
kubectl get pods --namespace dra-example-driver
# I verified DaemonSet update strategy correctly has **maxSurge: 1, maxUnavailable: 0**
# Check logs for legacy socket paths
kubectl logs -n dra-example-driver $POD_NAME | grep -E "(socket|endpoint)"
 
# Upgrade deployment triggering rolling update to enable seamless upgrades
kubectl patch daemonset dra-example-driver-kubeletplugin -n dra-example-driver \
  -p '{"spec":{"template":{"metadata":{"annotations":{"test-rolling-update":"'$(date +%s)'"}}}}}'
# verify uid socket paths in new pods
NEW_POD_NAME=$(kubectl get pods -n dra-example-driver -o jsonpath='{.items[0].metadata.name}')
kubectl logs -n dra-example-driver $NEW_POD_NAME | grep -E "(socket|endpoint)" | head -5

# Test non-seamless (overriding defaults)
helm upgrade dra-example-driver deployments/helm/dra-example-driver \
  --set kubeletPlugin.updateStrategy.rollingUpdate.maxSurge=0 \
  --set kubeletPlugin.updateStrategy.rollingUpdate.maxUnavailable=1

# Verify old rolling update strategy {"maxSurge": 0, "maxUnavailable": 1}
kubectl get daemonset dra-example-driver-kubeletplugin -n dra-example-driver \
  -o jsonpath='{.spec.updateStrategy.rollingUpdate}' | jq .
```
Results:
Seamless Upgrades Enabled default behavior:
maxSurge: 1, maxUnavailable: 0
```
I0422 23:21:04.522190 1 health.go:71] "connecting to registration socket" path="unix:///var/lib/kubelet/plugins_registry/gpu.example.com-<POD_UID>-reg.sock"
I0422 23:21:04.522536 1 health.go:89] "connecting to DRA socket" path="unix:///var/lib/kubelet/plugins/gpu.example.com/dra-<POD_UID>.sock"
```

Seamless Upgrades Disabled by user overriding helm chart:
maxSurge: 0, maxUnavailable: 1
- socket paths still include pod uid
```
I0422 23:18:56.245888 1 health.go:71] "connecting to registration socket" path="unix:///var/lib/kubelet/plugins_registry/gpu.example.com-<POD_UID>-reg.sock"
I0422 23:18:56.246339 1 health.go:89] "connecting to DRA socket" path="unix:///var/lib/kubelet/plugins/gpu.example.com/dra-<POD_UID>.sock"
```

### DIFF
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -49,14 +49,13 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
 	}
 	driver.state = state
 
-	helper, err := kubeletplugin.Start(
-		ctx,
-		driver,
+	helper, err := kubeletplugin.Start(ctx, driver,
 		kubeletplugin.KubeClient(config.coreclient),
 		kubeletplugin.NodeName(config.flags.nodeName),
 		kubeletplugin.DriverName(config.flags.driverName),
 		kubeletplugin.RegistrarDirectoryPath(config.flags.kubeletRegistrarDirectoryPath),
 		kubeletplugin.PluginDataDirectoryPath(config.DriverPluginPath()),
+		kubeletplugin.RollingUpdate(types.UID(config.flags.podUID)),
 	)
 	if err != nil {
 		return nil, err
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -52,6 +52,7 @@ type Flags struct {
 	healthcheckPort               int
 	profile                       string
 	driverName                    string
+	podUID                        string
 }
 
 type Config struct {
@@ -146,6 +147,12 @@ func newApp() *cli.App {
 			Usage:       "Name of the DRA driver. Its default is derived from the device profile.",
 			Destination: &flags.driverName,
 			EnvVars:     []string{"DRIVER_NAME"},
+		},
+		&cli.StringFlag{
+			Name:        "pod-uid",
+			Usage:       "UID of the pod (used for seamless upgrades to create unique socket names).",
+			Destination: &flags.podUID,
+			EnvVars:     []string{"POD_UID"},
 		},
 	}
 	cliFlags = append(cliFlags, flags.kubeClientConfig.Flags()...)
--- a/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
+++ b/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
@@ -84,6 +84,10 @@ spec:
         - name: HEALTHCHECK_PORT
           value: {{ .Values.kubeletPlugin.containers.plugin.healthcheckPort | quote }}
         {{- end }}
+        - name: POD_UID
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.uid
         volumeMounts:
         - name: plugins-registry
           mountPath: {{ .Values.kubeletPlugin.kubeletRegistrarDirectoryPath | quote }}
--- a/deployments/helm/dra-example-driver/values.yaml
+++ b/deployments/helm/dra-example-driver/values.yaml
@@ -41,6 +41,11 @@ kubeletPlugin:
   priorityClassName: "system-node-critical"
   updateStrategy:
     type: RollingUpdate
+    # for seamless rolling updates, maxSurge can be greater than 0.
+    # See: https://pkg.go.dev/k8s.io/dynamic-resource-allocation/kubeletplugin#RollingUpdate
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
   podAnnotations: {}
   podSecurityContext: {}
   nodeSelector: {}
--- a/cmd/dra-example-kubeletplugin/health.go
+++ b/cmd/dra-example-kubeletplugin/health.go
@@ -61,9 +61,12 @@ func startHealthcheck(ctx context.Context, config *Config) (*healthcheck, error)
 
 	regSockPath := (&url.URL{
 		Scheme: "unix",
-		// TODO: this needs to adapt when seamless upgrades
-		// are enabled and the filename includes a uid.
-		Path: path.Join(config.flags.kubeletRegistrarDirectoryPath, config.flags.driverName+"-reg.sock"),
+		Path: func() string {
+			if config.flags.podUID != "" {
+				return path.Join(config.flags.kubeletRegistrarDirectoryPath, config.flags.driverName+"-"+config.flags.podUID+"-reg.sock")
+			}
+			return path.Join(config.flags.kubeletRegistrarDirectoryPath, config.flags.driverName+"-reg.sock")
+		}(),
 	}).String()
 	log.Info("connecting to registration socket", "path", regSockPath)
 	regConn, err := grpc.NewClient(
@@ -76,7 +79,12 @@ func startHealthcheck(ctx context.Context, config *Config) (*healthcheck, error)
 
 	draSockPath := (&url.URL{
 		Scheme: "unix",
-		Path:   path.Join(config.DriverPluginPath(), "dra.sock"),
+		Path: func() string {
+			if config.flags.podUID != "" {
+				return path.Join(config.DriverPluginPath(), "dra-"+config.flags.podUID+".sock")
+			}
+			return path.Join(config.DriverPluginPath(), "dra.sock")
+		}(),
 	}).String()
 	log.Info("connecting to DRA socket", "path", draSockPath)
 	draConn, err := grpc.NewClient(
--- a/deployments/helm/dra-example-driver/Chart.yaml
+++ b/deployments/helm/dra-example-driver/Chart.yaml
@@ -25,4 +25,4 @@ version: 0.0.0-dev
 # It is recommended to use it with quotes.
 appVersion: "v0.2.1"
 
-kubeVersion: ">=1.32.0-0"
+kubeVersion: ">=1.33.0-0"


cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -49,15 +49,22 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
+	kubeletPluginOptions := []kubeletplugin.Option{
 		kubeletplugin.KubeClient(config.coreclient),
 		kubeletplugin.NodeName(config.flags.nodeName),
 		kubeletplugin.DriverName(consts.DriverName),
 		kubeletplugin.RegistrarDirectoryPath(config.flags.kubeletRegistrarDirectoryPath),
 		kubeletplugin.PluginDataDirectoryPath(config.DriverPluginPath()),
-	)
+	}
+
+	// Enable seamless upgrades when POD_UID is available (which gets set by Helm when seamlessUpgrades.enabled=true)
```
nojnhuh: There might be some other reason down the road that the pod UID is useful, so I think I would prefer to add a separate toggle to the config (exposed by a command line flag) that only sets the `RollingUpdate` option.

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
```diff
@@ -12,10 +12,18 @@ spec:
     matchLabels:
       {{- include "dra-example-driver.selectorLabels" . | nindent 6 }}
       app.kubernetes.io/component: kubeletplugin
+  {{- if .Values.kubeletPlugin.seamlessUpgrades.enabled }}
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 0
+  {{- else }}
```
nojnhuh: Setting `kubeletPlugin.seamlessUpgrades.enabled` will silently ignore the `kubeletPlugin.updateStrategy` value which isn't ideal.

Maybe we could keep the existing `updateStrategy` template block and use a different default for `maxSurge` and `maxUnavailable` when seamless upgrades are enabled? Then `kubeletPlugin.updateStrategy` is still authoritative if users express an opinion on that, but only setting `kubeletPlugin.seamlessUpgrades.enabled` is enough to set them to values that actually exercise seamless upgrades if they don't.

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
```diff
@@ -80,6 +88,12 @@ spec:
         - name: HEALTHCHECK_PORT
           value: {{ .Values.kubeletPlugin.containers.plugin.healthcheckPort | quote }}
         {{- end }}
+        {{- if .Values.kubeletPlugin.seamlessUpgrades.enabled }}
```
nojnhuh: If we don't enable seamless upgrades based on whether the pod UID is available, then I think it's reasonable to always unconditionally set this environment variable for the pods.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -46,6 +46,8 @@ controller:
 
 kubeletPlugin:
   priorityClassName: "system-node-critical"
+  seamlessUpgrades:
+    enabled: false
```
nojnhuh: We could enable this for e2e tests in case that happens to flag any issues.

https://github.com/kubernetes-sigs/dra-example-driver/blob/9ba4490762833a8ccffff5bf6f4c42d59eb15279/test/e2e/setup-e2e.sh#L36

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
alimaazamat: With my changes the user choice always wins so if we have seamless upgrades enabled and maxSurge/maxUnavailable specified then it would honor that, which wouldn't work with seamlessUpgrades

cmd/dra-example-kubeletplugin/driver.go
alimaazamat: should we instead have seamless upgrades flag instead? like isnt the user choice seamless upgrades which is done with rollingUpdate?

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
alimaazamat: Just to confirm, if user provides updateStrategy but seamlessUpgrades is enabled, then overwrite just the maxSurge and maxUnavailable of the given updateStrategy?

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
alimaazamat: Okay so POD_UID unconditionally set and then I added SEAMLESS_UPGRADES to yaml when enabled so that way it can be a CLI flag.

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
alimaazamat: If user specifies `updateStrategy.rollingUpdate` then use it
Otherwise
If `seamlessUpgrades.enabled = true `→ maxSurge: 1, maxUnavailable: 0
If s`eamlessUpgrades.enabled = false` → maxSurge: 0, maxUnavailable: 1

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -49,15 +49,22 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
 		kubeletplugin.DriverName(config.flags.driverName),
 		kubeletplugin.RegistrarDirectoryPath(config.flags.kubeletRegistrarDirectoryPath),
 		kubeletplugin.PluginDataDirectoryPath(config.DriverPluginPath()),
-	)
+	}
+
+	// Enable seamless upgrades when both seamless upgrades are enabled and POD_UID is available
+	if config.flags.seamlessUpgrades && config.flags.podUID != "" {
+		kubeletPluginOptions = append(kubeletPluginOptions,
+			kubeletplugin.RollingUpdate(types.UID(config.flags.podUID)),
```
nojnhuh: Since this depends on Kubernetes 1.33+, let's record that requirement in the chart: https://github.com/kubernetes-sigs/dra-example-driver/blob/b5001fcbe0427278502782395a289ceb9de6116e/deployments/helm/dra-example-driver/Chart.yaml#L28

1.33 is EOL, so I think we can ignore older versions for now. We can wait for someone to ask about running this on LTS versions 😃

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -147,6 +149,18 @@ func newApp() *cli.App {
...
 			EnvVars:     []string{"DRIVER_NAME"},
 		},
+		&cli.StringFlag{
+			Name:        "pod-uid",
+			Usage:       "UID of the pod (used for seamless upgrades to create unique socket names).",
+			Destination: &flags.podUID,
+			EnvVars:     []string{"POD_UID"},
+		},
+		&cli.BoolFlag{
+			Name:        "seamless-upgrades",
```
nojnhuh: I wonder if we can skip a flag for this and keep it enabled, then require the Pod UID to be set. Can you try that out and look for any obvious issues with the `updateStrategy` params for seamless and seamful(?) upgrades?

A flag would be required if we want to support pre-1.33 Kubernetes, but that's old enough now that I think we have an excuse to stick to newer versions. I suppose that may break users running this binary outside Kubernetes altogether, but I'd be surprised if anyone is actually doing that.

alimaazamat: Tested this behavior and works as intended:
If user provides `updateStrategy.rollingUpdate` we ignore the helm chart `seamlessUpgrades.enabled` value and the given values are picked up.
Otherwise if `seamlessUpgrades.enabled` (which by default it is) maxSurge 1 maxUnavailable 0
Otherwise if `seamlessUpgrades.disabled` maxSurge 0 maxUnavailable 1

There is the edge case if a user gives only maxSurge or only maxUnavailable . Then it would pick up only the given value and use k8s defaults (maxSurge 0 maxUnavailable 1) regardless of it seamlessUpgrades is enabled/disabled:
- If user specifies maxSurge 5 then we get default maxUnavailable 1 
- If user specifies maxUnavailable 3 then we get default maxSurge 0

Sometimes this can go against the intended behavior depending on seamlessUpgrades.enabled/disabled but like you said the user should just be able to configure this correctly.

alimaazamat: Commited [1cc6fff](https://github.com/kubernetes-sigs/dra-example-driver/pull/113/commits/1cc6fff926a041bb2d8d94d8c09c87b920b5dfeb) where we default maxSurge 1 maxUnavailable 0 since we have seamlessUpgrades.enabled. It was hard to read so I did default vars in the template. Lmk if that sounds good.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -49,15 +49,22 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
 		kubeletplugin.DriverName(config.flags.driverName),
 		kubeletplugin.RegistrarDirectoryPath(config.flags.kubeletRegistrarDirectoryPath),
 		kubeletplugin.PluginDataDirectoryPath(config.DriverPluginPath()),
-	)
+	}
+
+	// Enable seamless upgrades when POD_UID is available
+	if config.flags.podUID != "" {
+		kubeletPluginOptions = append(kubeletPluginOptions,
+			kubeletplugin.RollingUpdate(types.UID(config.flags.podUID)),
```
nojnhuh: nit: Checking the Pod UID here is redundant. Passing an empty Pod UID to `kubeletplugin.RollingUpdate` is equivalent to not calling it at all.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -39,6 +39,8 @@ kubeletPlugin:
   # deviceProfile is used. Not relevant for other profiles.
   numDevices: 9
   priorityClassName: "system-node-critical"
+  seamlessUpgrades:
+    enabled: true
```
nojnhuh: Can we remove this and let users tweak the `updateStrategy` directly?

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -39,6 +39,8 @@ kubeletPlugin:
...
   updateStrategy:
     type: RollingUpdate
```
nojnhuh, line 43: Can we add `maxSurge` and `maxUnavailable` inline here and then add the literal value of `.Values.kubeletPlugin.updateStrategy` to the DaemonSet like the template did before?

And can we include a comment explaining that `maxSurge` can be greater than 0 because of https://pkg.go.dev/k8s.io/dynamic-resource-allocation/kubeletplugin#RollingUpdate?

cmd/dra-example-kubeletplugin/driver.go
alimaazamat: Done in https://github.com/kubernetes-sigs/dra-example-driver/pull/113/commits/47a2e9b9aa8582130ebc1b99a8f8c6357721452e

deployments/helm/dra-example-driver/values.yaml
alimaazamat: Done in https://github.com/kubernetes-sigs/dra-example-driver/pull/113/commits/47a2e9b9aa8582130ebc1b99a8f8c6357721452e

deployments/helm/dra-example-driver/values.yaml
alimaazamat: Done in [47a2e9b](https://github.com/kubernetes-sigs/dra-example-driver/commit/47a2e9b9aa8582130ebc1b99a8f8c6357721452e)

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -49,15 +49,19 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
+	kubeletPluginOptions := []kubeletplugin.Option{
 		kubeletplugin.KubeClient(config.coreclient),
 		kubeletplugin.NodeName(config.flags.nodeName),
 		kubeletplugin.DriverName(config.flags.driverName),
 		kubeletplugin.RegistrarDirectoryPath(config.flags.kubeletRegistrarDirectoryPath),
 		kubeletplugin.PluginDataDirectoryPath(config.DriverPluginPath()),
+	}
+
+	kubeletPluginOptions = append(kubeletPluginOptions,
+		kubeletplugin.RollingUpdate(types.UID(config.flags.podUID)),
```
nojnhuh, line 58: One more nit: can we define this inline with the other options? Then we can get rid of the `kubeletPluginOptions` variable and pass everything directly to `kubeletplugin.Start` like before.

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
```diff
@@ -12,10 +12,12 @@ spec:
...
       app.kubernetes.io/component: kubeletplugin
-  {{- with .Values.kubeletPlugin.updateStrategy }}
   updateStrategy:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
+    {{- if .Values.kubeletPlugin.updateStrategy }}
+    {{- toYaml .Values.kubeletPlugin.updateStrategy | nindent 4 }}
+    {{- else }}
+    type: RollingUpdate
+    {{- end }}
```
nojnhuh: I think we can keep this exactly as it was before in case users really want to leave the `updateStrategy` completely undefined.

cmd/dra-example-kubeletplugin/driver.go
alimaazamat: Done in https://github.com/kubernetes-sigs/dra-example-driver/pull/113/commits/c9cbe693cf8a2c2576c1ff9d51e1ae9e3fa28cd4

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
alimaazamat: When you say keep it as it was before, you mean with `{{- with .Values.kubeletPlugin.updateStrategy }}`? Or a previous commit I had?
I reverted it to `{{- with .Values.kubeletPlugin.updateStrategy }}` with https://github.com/kubernetes-sigs/dra-example-driver/pull/113/commits/5e788f89c045aa008a1f549439e7e0d1ff50f679

nojnhuh (commented):
/lgtm

Thanks @alimaazamat! Could you please squash down to one commit?

nojnhuh (approved):
/approve

PR #116, opened 2025-08-04
Updated DRA APIs to v1 

### LINKED ISSUE #115: Update DRA APIs to v1
Kubernetes 1.34 introduces the `resource.k8s.io/v1` version for the DRA APIs. The example driver should demonstrate how to implement and use the latest and greatest API. These are the rough steps to do that:

1. Update all of the k8s.io dependencies to the latest 1.34 pre-release version (currently [v1.34.0-rc.1](https://github.com/kubernetes/kubernetes/releases/tag/v1.34.0-rc.1)). See #95 for an example (which also includes some changes specific to that upgrade that won't apply here).

1. Importing the new k8s.io dependencies will require updating the driver itself to operate against the new API version. Imports like this should all be updated to use v1 instead of v1beta1:
https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/cmd/dra-example-kubeletplugin/driver.go#L24

1. Update the validating webhook to use v1:
https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/cmd/dra-example-webhook/main.go#L30

1. Update the kind config to remove the config [enabling API versions](https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/demo/scripts/kind-cluster-config.yaml#L17-L18) and the [feature gate](https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/demo/scripts/kind-cluster-config.yaml#L3-L4). The v1 API and the DRA feature gate are enabled by default and I think it's valuable to exercise that.

1. Update YAML manifests to the new version, e.g. in [examples](https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/demo/gpu-test1.yaml#L11), [the Helm chart](https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/deployments/helm/dra-example-driver/templates/deviceclass.yaml#L2), and [sample output](https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/README.md?plain=1#L111).

1. Call out the new minimum required Kubernetes version in the Helm chart as 1.34:
https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/deployments/helm/dra-example-driver/Chart.yaml#L28

The differences between the APIs are fairly straightforward, e.g. from https://github.com/kubernetes/kubernetes/commit/5c4f81743c2ee73fae71d336304e491b2539b6e3:
- ResourceSlice no longer contains the `basic` level of indirection:
  ```diff
  diff --git a/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/allocatortesting/allocator_testing.go b/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/allocatortesting/allocator_testing.go
  index ebd22fc217eeb..3d8c81b1fde01 100644
  --- a/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/allocatortesting/allocator_testing.go
  +++ b/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/allocatortesting/allocator_testing.go
  @@ -257,16 +259,14 @@ const (
   // generate a Device object with the given name, capacity and attributes.
   func device(name string, capacity any, attributes map[resourceapi.QualifiedName]resourceapi.DeviceAttribute) wrapDevice {
    device := resourceapi.Device{
  -		Name: name,
  -		Basic: &resourceapi.BasicDevice{
  -			Attributes: attributes,
  -		},
  +		Name:       name,
  +		Attributes: attributes,
    }
  ```
- ResourceClaims now have an additional `exactly` level of indirection:
  ```diff
  diff --git a/test/e2e/dra/test-driver/deploy/example/pod-inline-multiple.yaml b/test/e2e/dra/test-driver/deploy/example/pod-inline-multiple.yaml
  index 520935c59265a..59cd7b63b59f5 100644
  --- a/test/e2e/dra/test-driver/deploy/example/pod-inline-multiple.yaml
  +++ b/test/e2e/dra/test-driver/deploy/example/pod-inline-multiple.yaml
  @@ -1,4 +1,4 @@
  -apiVersion: resource.k8s.io/v1beta1
  +apiVersion: resource.k8s.io/v1
   kind: ResourceClaimTemplate
   metadata:
     name: pause-template
  @@ -11,7 +11,8 @@ spec:
       devices:
         requests:
         - name: my-device
  -        deviceClassName: example
  +        exactly:
  +          deviceClassName: example
         config:
         - opaque:
             driver: test-driver.cdi.k8s.io
  ```

For reference, the v1beta2 APIs are documented here, which are equivalent to v1: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/

/assign @alimaazamat

alimaazamat:
fixes #115


klueska:
Please see the following for how to gt the DRA driver to work with all of v1, v1alpha1, and v1alpha2: https://github.com/kubernetes-sigs/dra-example-driver/issues/115#issuecomment-3201632223

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -149,3 +152,12 @@ func (d *driver) unprepareResourceClaim(_ context.Context, claim kubeletplugin.N
...
 }
+
+func (d *driver) HandleError(ctx context.Context, err error, msg string) {
+	utilruntime.HandleErrorWithContext(ctx, err, msg)
+	if !errors.Is(err, kubeletplugin.ErrRecoverable) {
+		// Fatal error - exit the process
+		klog.FromContext(ctx).Error(err, "Fatal error encountered, exiting process")
+		os.Exit(1)
+	}
+}
```
alimaazamat, line 162: HandleError is mandatory as of DRA v1 API 
https://github.com/kubernetes/kubernetes/blob/release-1.34/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L124
Example: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dra/test-driver/app/kubeletplugin.go#L244

deployments/helm/dra-example-driver/templates/deviceclass.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: resource.k8s.io/v1beta1
+apiVersion: resource.k8s.io/v1
```
klueska: If you still want this to work on k8s 1.32 (as indicated in the `kubeVersion` of the helm chart) then this can't be changed.

deployments/helm/dra-example-driver/templates/validatingwebhookconfiguration.yaml
```diff
@@ -12,7 +12,7 @@ webhooks:
 - name: "dra.example.com"
   rules:
   - apiGroups:   ["resource.k8s.io"]
-    apiVersions: ["v1beta1"]
+    apiVersions: ["v1"]
```
klueska: If you still want this to work on k8s 1.32 (as indicated in the `kubeVersion` of the helm chart) then this needs to be:
```
apiVersions: ["v1", "v1beta1", "v1beta2"]
```

deployments/helm/dra-example-driver/templates/deviceclass.yaml
alimaazamat: Thanks so much, fixed!

deployments/helm/dra-example-driver/templates/validatingwebhookconfiguration.yaml
alimaazamat: Done thank you!

lauralorenz:
I am using the example driver in some kubernetes documentation, and want to queue up a PR that references the tag for the image that will be released with this (in https://github.com/kubernetes/website/pull/51979). Will the new image released be `registry.k8s.io/dra-example-driver/dra-example-driver:v0.2.0`?

Also for what it's worth, I did run through that tutorial with a local build of the example driver from this branch and it worked great so, I can confirm that it works with k/k 1.34 and the v1 APIs haha 😇

nojnhuh:
> I am using the example driver in some kubernetes documentation, and want to queue up a PR that references the tag for the image that will be released with this (in [kubernetes/website#51979](https://github.com/kubernetes/website/pull/51979)). Will the new image released be `registry.k8s.io/dra-example-driver/dra-example-driver:v0.2.0`?

Yes, I'd like to publish a v0.2.0 image and chart soon after this merges.

> Also for what it's worth, I did run through that tutorial with a local build of the example driver from this branch and it worked great so, I can confirm that it works with k/k 1.34 and the v1 APIs haha 😇

Thanks for checking!

demo/scripts/kind-cluster-config.yaml
```diff
@@ -14,8 +12,8 @@ nodes:
   - |
     kind: ClusterConfiguration
     apiServer:
-        extraArgs:
-          runtime-config: "resource.k8s.io/v1beta1=true"
+      extraArgs:
+        runtime-config: "resource.k8s.io/v1beta1=true"
```
klueska, line 18: If you expect this to work with kind clusters >= 1.32.0, then you probably want
```suggestion
      extraArgs:
        runtime-config: "resource.k8s.io/all=true"
```

demo/scripts/kind-cluster-config.yaml
```diff
@@ -1,7 +1,5 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
-featureGates:
-  DynamicResourceAllocation: true
```
klueska: If you expect this to work with kind clusters >= 1.32.0, then you probably want to keep this.

nojnhuh: I suggested removing this if we only intend to test on 1.34+ going forward where the feature gate is enabled by default. That requirement isn't enforced anywhere though for someone who might be trying to run this against an older cluster though, so keeping the feature gate makes sense.

Eventually we'll come to a point though where the feature gate will get removed and clusters will fail to start because of the invalid feature gate, but we can cross that bridge when we get there.

klueska:
I know I saw the changes from https://github.com/NVIDIA/k8s-dra-driver-gpu/commit/920a2871da917b2630425144011885d4dd3bb95d integrated in the PR at some point, but they seem to have been dropped...

alimaazamat:
Also having `runtime-config: "resource.k8s.io/all=true"` causes e2e tests to fail (regardless of including/excluding changes from [NVIDIA/k8s-dra-driver-gpu@920a287](https://github.com/NVIDIA/k8s-dra-driver-gpu/commit/920a2871da917b2630425144011885d4dd3bb95d))
`error: error execution phase wait-control-plane: failed while waiting for the control plane to start: kube-apiserver check failed at https://172.18.0.2:6443/livez: client rate limiter Wait returned an error: context deadline exceeded`

When excluding changes from [NVIDIA/k8s-dra-driver-gpu@920a287](https://github.com/NVIDIA/k8s-dra-driver-gpu/commit/920a2871da917b2630425144011885d4dd3bb95d) and with `runtime-config: "resource.k8s.io/all=v1beta1"` we pass

alimaazamat:
> > Also having `runtime-config: "resource.k8s.io/all=true"` causes e2e tests to fail (regardless of including/excluding changes from [NVIDIA/k8s-dra-driver-gpu@920a287](https://github.com/NVIDIA/k8s-dra-driver-gpu/commit/920a2871da917b2630425144011885d4dd3bb95d))
> 
> It looks like `all` only works as `api/all`, but not for a particular API group. Here we probably have to enable each one individually. Let's enable both v1beta1 and v1beta2 (leaving v1 enabled by default). https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#:~:text=%2D%2Druntime%2Dconfig%20%3Ccomma%2Dseparated%20%27key%3Dvalue%27%20pairs%3E

`resource.k8s.io/v1beta1=true` with cluster 1.32 errors: 
```Shell
Waiting for webhook to be available
error: resource mapping not found for name: "webhook-test" namespace: "" from "STDIN": no matches for kind "ResourceClaim" in version "resource.k8s.io/v1"
ensure CRDs are installed first
```
 
`resource.k8s.io/v1=true,resource.k8s.io/v1beta1=true` with cluster 1.32 errors:
```Shell
[api-check] The API server is not healthy after 4m0.002513276s
 ```

`resource.k8s.io/v1=true,resource.k8s.io/v1beta1=true`with cluster 1.32 with conditional statements in deviceclass.yaml errors:
```Shell
[api-check] The API server is not healthy after 4m0.00208332s
  ```

`resource.k8s.io/v1beta1=true` with cluster 1.34.0-rc.1 and no conditionals in deviceclass.yaml we pass (aka the base case on PR e2e test results)

**Comment by alimaazamat** [comment]
> > Also having `runtime-config: "resource.k8s.io/all=true"` causes e2e tests to fail (regardless of including/excluding changes from [NVIDIA/k8s-dra-driver-gpu@920a287](https://github.com/NVIDIA/k8s-dra-driver-gpu/commit/920a2871da917b2630425144011885d4dd3bb95d))
> 
> It looks like `all` only works as `api/all`, but not for a particular API group. Here we probably have to enable each one individually. Let's enable both v1beta1 and v1beta2 (leaving v1 enabled by default). https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#:~:text=%2D%2Druntime%2Dconfig%20%3Ccomma%2Dseparated%20%27key%3Dvalue%27%20pairs%3E

With `runtime-config: "resource.k8s.io/v1beta1=true,resource.k8s.io/v1beta2=true"` I tested this with:
 1.32 cluster (errors out ```
[api-check] The API server is not healthy after 4m0.001581692s

Unfortunately, an error has occurred:
        context deadline exceeded

This error is likely caused by:
        - The kubelet is not running
        - The kubelet is unhealthy due to a misconfiguration of the node in some way (required cgroups disabled)
```)
 1.33 cluster (correctly uses v1beta2)
 1.34 cluster (correctly uses v1)

To recap:
1.32 allows v1beta1
1.33 allows v1beta1, v1beta2
1.34 allows v1beta1, v1beta2, v1

File: demo/scripts/common.sh [hunk 6]
```diff
@@ -49,6 +49,32 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
+    fi
+}
+
+# Function to update kind-cluster-config.yaml with appropriate runtime-config
+update_kind_config_runtime() {
+    local config_file="${KIND_CLUSTER_CONFIG_PATH}"
+    local runtime_config="$(get_dra_runtime_config)"
+    
+    echo "Setting runtime-config for ${KIND_K8S_TAG}: ${runtime_config}"
+    sed "s|runtime-config: \".*\"|runtime-config: \"${runtime_config}\"|" "${config_file}" > "${config_file}.tmp"
```
nojnhuh: I think I'd prefer to stick with only explicitly enabling v1beta1 in the static config. That should produce a working cluster for all versions we care about and still let us use v1 for 1.34 and beyond. I don't think missing v1beta2 for 1.33 is a big deal.

deployments/helm/dra-example-driver/templates/validatingwebhookconfiguration.yaml
nojnhuh: We should only advertise in the webhook config here what the webhook can actually validate, and with the changes here it only knows about v1. By omitting the other versions here, the API server knows it should convert between the other versions so users can still submit v1beta1 or v1beta2 objects. That might not be optimal for performance, but we can assess and follow up later if we find that's a concern.

nojnhuh: It might be worth making the webhook part of the e2e test use at least one v1beta1 object to demonstrate that.

nojnhuh: Maybe the easier thing to do for now is revert the webhook changes and only validate v1beta1 so we can keep compatibility with 1.32.

demo/scripts/common.sh
alimaazamat: Yeah I agree with this!

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -149,3 +152,12 @@ func (d *driver) unprepareResourceClaim(_ context.Context, claim kubeletplugin.N
 
 	return nil
 }
+
+func (d *driver) HandleError(ctx context.Context, err error, msg string) {
+	utilruntime.HandleErrorWithContext(ctx, err, msg)
+	if !errors.Is(err, kubeletplugin.ErrRecoverable) {
+		// Fatal error - exit the process
+		klog.FromContext(ctx).Error(err, "Fatal error encountered, exiting process")
+		os.Exit(1)
```
nojnhuh: Ideally we send this down the graceful shutdown route instead of aborting. Roughly following the example of the [k/k test driver](https://github.com/kubernetes/kubernetes/blob/091f87c10bc3532041b77a783a5f832de5506dc8/test/e2e/dra/test-driver/app/kubeletplugin.go#L244), I came up with this:

<details>

```diff
diff --git a/cmd/dra-example-kubeletplugin/driver.go b/cmd/dra-example-kubeletplugin/driver.go
index 02f9dba..0f6e224 100644
--- a/cmd/dra-example-kubeletplugin/driver.go
+++ b/cmd/dra-example-kubeletplugin/driver.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"maps"
-	"os"
 
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -39,11 +38,13 @@ type driver struct {
 	helper      *kubeletplugin.Helper
 	state       *DeviceState
 	healthcheck *healthcheck
+	cancelCtx   func(error)
 }
 
 func NewDriver(ctx context.Context, config *Config) (*driver, error) {
 	driver := &driver{
-		client: config.coreclient,
+		client:    config.coreclient,
+		cancelCtx: config.cancelMainCtx,
 	}
 
 	state, err := NewDeviceState(config)
@@ -155,9 +156,7 @@ func (d *driver) unprepareResourceClaim(_ context.Context, claim kubeletplugin.N
 
 func (d *driver) HandleError(ctx context.Context, err error, msg string) {
 	utilruntime.HandleErrorWithContext(ctx, err, msg)
-	if !errors.Is(err, kubeletplugin.ErrRecoverable) {
-		// Fatal error - exit the process
-		klog.FromContext(ctx).Error(err, "Fatal error encountered, exiting process")
-		os.Exit(1)
+	if !errors.Is(err, kubeletplugin.ErrRecoverable) && d.cancelCtx != nil {
+		d.cancelCtx(fmt.Errorf("fatal background error: %w", err))
 	}
 }
diff --git a/cmd/dra-example-kubeletplugin/main.go b/cmd/dra-example-kubeletplugin/main.go
index a217b4e..e0cfaf9 100644
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -52,8 +52,9 @@ type Flags struct {
 }
 
 type Config struct {
-	flags      *Flags
-	coreclient coreclientset.Interface
+	flags         *Flags
+	coreclient    coreclientset.Interface
+	cancelMainCtx func(error)
 }
 
 func (c Config) DriverPluginPath() string {
@@ -172,6 +173,8 @@ func RunPlugin(ctx context.Context, config *Config) error {
 
 	ctx, stop := signal.NotifyContext(ctx, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
 	defer stop()
+	ctx, cancel := context.WithCancelCause(ctx)
+	config.cancelMainCtx = cancel
 
 	driver, err := NewDriver(ctx, config)
 	if err != nil {
@@ -182,7 +185,7 @@ func RunPlugin(ctx context.Context, config *Config) error {
 	// restore default signal behavior as soon as possible in case graceful
 	// shutdown gets stuck.
 	stop()
-	if err := ctx.Err(); err != nil && !errors.Is(err, context.Canceled) {
+	if err := context.Cause(ctx); err != nil && !errors.Is(err, context.Canceled) {
 		// A canceled context is the normal case here when the process receives
 		// a signal. Only log the error for more interesting cases.
 		logger.Error(err, "error from context")
```

</details>

demo/scripts/common.sh
```diff
@@ -36,12 +36,12 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 
 # The kubernetes tag to build the kind cluster from
 # From ${KIND_K8S_REPO}/tags
-: ${KIND_K8S_TAG:="v1.33.0"}
+: ${KIND_K8S_TAG:="v1.34.0-rc.1"}
```
nojnhuh: Reminder that we should come back to this to update it to the official release once the kind image is available.

demo/scripts/common.sh
```diff
@@ -36,12 +36,12 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
 
 # At present, kind has a new enough node image that we don't need to build our
 # own. This won't always be true and we may need to set the variable below to
 # 'true' from time to time as things change.
-: ${BUILD_KIND_IMAGE:="false"}
+: ${BUILD_KIND_IMAGE:="true"}
```
nojnhuh: Likewise, we should revert this once we can use a prebuilt kind image.

deployments/helm/dra-example-driver/templates/validatingwebhookconfiguration.yaml
```diff
@@ -12,7 +12,7 @@ webhooks:
 - name: "dra.example.com"
   rules:
   - apiGroups:   ["resource.k8s.io"]
-    apiVersions: ["v1beta1"]
+    apiVersions: ["v1", "v1beta1", "v1beta2"]
```
nojnhuh: ICYMI I added onto a resolved thread to add a comment about the webhook: https://github.com/kubernetes-sigs/dra-example-driver/pull/116#discussion_r2296773597

go.mod
```diff
@@ -6,19 +6,19 @@ toolchain go1.24.2
...
 	github.com/stretchr/testify v1.10.0
 	github.com/urfave/cli/v2 v2.25.3
-	google.golang.org/grpc v1.68.1
-	k8s.io/api v0.33.0
-	k8s.io/apimachinery v0.33.0
-	k8s.io/client-go v0.33.0
-	k8s.io/component-base v0.33.0
-	k8s.io/dynamic-resource-allocation v0.33.0
+	google.golang.org/grpc v1.72.1
+	k8s.io/api v0.34.0-rc.1
```
nojnhuh: Another reminder to bump these dependencies to the real release once they're available.

deployments/helm/dra-example-driver/templates/validatingwebhookconfiguration.yaml
klueska: My assumption was that you were (additionally) planning to update the webhook to follow what we did here: 
* https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/main/cmd/webhook/main.go#L201
* https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/main/cmd/webhook/resource.go

With this approach we have validated on 1.32, 1.33, and (a pre-build of) 1.34

nojnhuh: Eventually we'll need to support v1 directly in the webhooks once v1beta1 is removed. It's not strictly necessary to do right now though AFAICT. I've tested the v1beta1-only webhook with 1.32-1.34 and it works as expected to validate v1beta1, v1beta2, and v1 manifests. The webhook being able to validate each version on its own is probably more ideal overall, but I'd like to trim scope here where we can so we can cut a release of the driver as soon as possible after 1.34.0 is cut.

test/e2e/e2e.sh
```diff
@@ -440,19 +441,21 @@ kubectl delete -f demo/gpu-test4.yaml --timeout=25s
 kubectl delete -f demo/gpu-test5.yaml --timeout=25s
 kubectl delete -f demo/gpu-test6.yaml --timeout=25s
 
-# Webhook should reject invalid resources
+# Webhook should reject invalid resources (v1 version)
```
nojnhuh: At first glance this kind of reads to me like it's because the resource is v1 that it's rejected, which isn't true. This might be a bit clearer? And similarly below for the v1beta1 ResourceClaim.

```suggestion
# Webhook should reject invalid v1 resources
```

test/e2e/e2e.sh
```diff
@@ -470,8 +473,37 @@ then
...
 apiVersion: resource.k8s.io/v1beta1
+kind: ResourceClaim
+metadata:
+  name: webhook-test-v1beta1
+spec:
+  devices:
+    requests:
+    - name: ts-gpu
+      exactly:
+        deviceClassName: gpu.example.com
```
nojnhuh: I think e2e is failing because this isn't an otherwise valid v1beta1 ResourceClaim.

```suggestion
      deviceClassName: gpu.example.com
```

test/e2e/e2e.sh
```diff
@@ -470,8 +473,37 @@ then
...
+        parameters:
+          apiVersion: gpu.resource.example.com/v1alpha1
+          kind: GpuConfig
+          sharing:
+            strategy: TimeSlicing
+            timeSlicingConfig:
+              interval: InvalidInterval
+EOF
+then
+  echo "Webhook did not reject v1beta1 ResourceClaim invalid GpuConfig with the expected message"
```
nojnhuh, line 500: Could we also mention the API version of the ResourceClaim in the message for the v1 test below like we do here?

test/e2e/e2e.sh
alimaazamat: Yes just figured that out locally fixing now thank you

nojnhuh:
I ran through all the valid combinations of v1beta1/v1beta2/v1 and 1.32/1.33/1.34 and things are working as I expect, so overall LGTM.

@alimaazamat As soon as the [k8s.io dependencies](https://pkg.go.dev/k8s.io/dynamic-resource-allocation?tab=versions) and [kind image](https://hub.docker.com/r/kindest/node/tags) for the final 1.34.0 release are available let's get a clean CI run before squashing and merging this. Feel free to squash earlier too if that's easier.

deployments/helm/dra-example-driver/templates/validatingwebhookconfiguration.yaml
nojnhuh: One point I hadn't considered here is that implementing the webhook only for v1beta1 means that API version has to be explicitly enabled for the API server, which isn't ideal in 1.34 when v1 is enabled by default. If the webhook can natively validate both v1beta1 and v1, then we can both keep back-compat to 1.32 and not require any extra API versions to be enabled for 1.34+.

I think I'd still like to cut v0.2.0 here as soon as we can after 1.34 is published though to unblock @lauralorenz's docs in https://github.com/kubernetes/website/pull/51979. Those docs define a separate DeviceClass and don't configure the webhook, so following those docs shouldn't require v1beta1 to be enabled.

Overall I'm slightly more inclined to keep things as-is so we're able to cut a v0.2.0 here ASAP and queue up that change for the next release. @alimaazamat If you're able to implement and test that before the 1.34 release though that would be awesome! But not a deal breaker if we don't get it in IMO considering users of this chart already have to enable the v1beta1 APIs.

demo/scripts/common.sh
nojnhuh: I'm thinking we don't need to block merging this PR on a prebuilt kind image. If an image we build ourselves for the 1.34.0 release works, then that's a good enough signal to me that the prebuilt one will work. Still worth changing this value backwhen the kind image is available though.

Also if we get to it, updating the comment here to agree with the current value would be worthwhile.

nojnhuh: Since there's already a v1.34.0, let's go ahead and refer to that, and default to `false` here again. https://hub.docker.com/layers/kindest/node/v1.34.0/images/sha256-ede5bb6984830375d264e41b7169d03643f586d3e35316bf5229eaeb12bb18e0

nojnhuh (approved):
/lgtm
/approve

Issue #119, opened 2025-08-26
Cut v0.2.0


nojnhuh:
Per [RELEASE.md](https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/RELEASE.md), I'm requesting that we cut a v0.2.0 of the dra-example-driver container image and 0.2.0 of the Helm chart. I can drive cutting the release with consensus from at least one other maintainer that a release is warranted. FYI @pohly @klueska @bart0sh @elezar @byako 

Here are the release notes I'm proposing:

```
## Notable Changes since v0.1.0

* @nojnhuh is now a maintainer (by @jackfrancis in #99)
* New Helm chart values may be set to define where kubelet plugin handles are created: `kubeletPlugin.kubeletRegistrarDirectoryPath`, `kubeletPlugin.kubeletRegistrarDirectoryPath` (by @nojnhuh in #98)
* A new example shows how to share devices between an `initContainer` and `container` in the same Pod (by @jackfrancis #106)
* The kubelet plugin now includes a livenessProbe (by @nojnhuh in #104)
* The number of simulated devices on each Node is now configurable via the `kubeletPlugin.numDevices` Helm chart value (by @Neha-dot-Yadav in #111)
* The driver now supports Kubernetes 1.34 (backwards compatible to 1.32). Examples have been updated to use the `resource.k8s.io/v1` API version. Enabling the `resource.k8s.io/v1beta1` API version for `kube-apiserver` is still required. (by @alimaazamat in #116)
```

The release notes mention #116 which hasn't merged yet but I intend to include in the release.

/assign

PR #120, opened 2025-08-26
Cut v0.2.0

nojnhuh:
This PR fulfills step 3 of the release guide:
https://github.com/kubernetes-sigs/dra-example-driver/blob/fc2d0ba1d0754ab379ced5b23d98bf840435e278/RELEASE.md?plain=1#L17-L18

ref #119

/hold for #116

### DIFF
--- a/deployments/helm/dra-example-driver/Chart.yaml
+++ b/deployments/helm/dra-example-driver/Chart.yaml
@@ -23,6 +23,6 @@ version: 0.0.0-dev
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "v0.1.0"
+appVersion: "v0.2.0"
 
 kubeVersion: ">=1.32.0-0"
--- a/common.mk
+++ b/common.mk
@@ -17,7 +17,7 @@ GOLANG_VERSION ?= 1.24.2
 DRIVER_NAME := dra-example-driver
 MODULE := sigs.k8s.io/$(DRIVER_NAME)
 
-VERSION  ?= v0.1.0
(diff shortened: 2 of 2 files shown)


nojnhuh:
#116 merged!

/hold cancel
/assign @pohly @klueska

deployments/helm/dra-example-driver/Chart.yaml
```diff
@@ -23,6 +23,6 @@ version: 0.0.0-dev
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "v0.1.0"
+appVersion: "v0.2.0"
```
bart0sh, line 26: I can also see` v0.1.0` in the `common.mk` file. Should it be updated there as well?

nojnhuh: That doesn't need to be updated since that variable gets overwritten by the tag value:
https://github.com/kubernetes-sigs/dra-example-driver/blob/7fbfdca89d26702c8b85bad81494fd52ec7cf676/demo/scripts/build-driver-image.sh#L43

nojnhuh: And when I tag this commit locally and run `make push-release-artifacts` I see this snipped from `docker inspect`-ing the resulting image:
```
            "Labels": {
                ...
                "version": "v0.2.0"
            },
```

bart0sh: If that value is not used, should we remove it? It's still a little bit confusing to see it there.

nojnhuh: Removed it.

PR #123, opened 2025-09-26
Update DRA API version to v1

enoodle:
Bump the API version from v1beta1 to v1 in all relevant files, including Go imports, Helm charts, and test cases. This ensures compatibility with the latest Kubernetes resource APIs.


cmd/dra-example-kubeletplugin/state.go
```diff
@@ -23,7 +23,7 @@ import (
 
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1beta1"
```
nojnhuh, line 26: I was originally thinking we should keep the same version here since this could affect how checkpoints are represented on disk. In practice though I don't think there's any change since the JSON representations of the types aren't changed between v1beta1 and v1.

cmd/dra-example-webhook/main.go
```diff
@@ -219,13 +210,10 @@ func admitResourceClaimParameters(ar admissionv1.AdmissionReview) *admissionv1.A
 	var deviceConfigs []resourceapi.DeviceClaimConfiguration
 	var specPath string
 
-	raw := ar.Request.Object.Raw
-	deserializer := codecs.UniversalDeserializer()
-
 	switch ar.Request.Resource {
-	case resourceClaimResource:
```
nojnhuh, line 226: Could you replace the last remaining uses of this variable (and the claim template one) with the new variables defined in resource.go?

cmd/dra-example-webhook/resource.go
nojnhuh, line 1: Thank you @klueska for blazing this trail in the NVIDIA driver!

cmd/dra-example-webhook/main.go
enoodle: I removed this variable to only use the new ones and added tests cases for the conversion and unknown types.

cmd/dra-example-kubeletplugin/state.go
enoodle: I am not sure about that.. I see that the Device struct is a bit different between those two versions. Do you want to keep using the v1beta1 version?

nojnhuh: Which differences do you see? This command from the k/k release-1.34 branch doesn't show any differences between v1beta1 and v1 that would affect the JSON representation of a `Device` on disk AFAICT:
```
git diff --no-index staging/src/k8s.io/kubelet/pkg/apis/dra/v1{beta1,}
```

Though with changes like https://github.com/kubernetes/kubernetes/pull/134520 potentially coming up, we should figure out a way to decouple the example driver's disk format and its dependencies. I think @pohly brought that up a while ago but I'm not having any luck finding that conversation.

I'm ok with leaving this version at v1beta1 for now so we can address this particular issue separately.

enoodle: I am not sure how it translates exaclty, but I see here that the structs are different:
https://github.com/kubernetes/api/blob/master/resource/v1/types.go#L260
https://github.com/kubernetes/api/blob/master/resource/v1beta1/types.go#L264

I updated the PR to keep using the previous v1beta1 version.

nojnhuh: Those are the Kubernetes API types, but the code here is referring to the kubelet API types:
- https://github.com/kubernetes/kubernetes/blob/183e01c6f2098710540bcc9cfa40bc8055eb31a6/staging/src/k8s.io/kubelet/pkg/apis/dra/v1beta1/api.pb.go#L210
- https://github.com/kubernetes/kubernetes/blob/183e01c6f2098710540bcc9cfa40bc8055eb31a6/staging/src/k8s.io/kubelet/pkg/apis/dra/v1/api.pb.go#L210

Those types serialize to the same JSON AFAICT.

We can address this later, so let's stick with v1beta here for now.

cmd/dra-example-kubeletplugin/state_test.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1beta1"
+	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1"
```
nojnhuh: I see unit tests are failing locally for me but changing this back to v1beta1 should fix things.

cmd/dra-example-webhook/resource.go
```diff
@@ -0,0 +1,154 @@
...
+	admissionv1 "k8s.io/api/admission/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	resourcev1beta1 "k8s.io/api/resource/v1beta1"
+	resourcev1beta2 "k8s.io/api/resource/v1beta2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	// This will register the conversions between api versions
```
nojnhuh: I think the linter will also flag this in CI:
```
cmd/dra-example-webhook/resource.go:31:2: Comment should end in a period (godot)
	// This will register the conversions between api versions
	^
```

```suggestion
	// This will register the conversions between api versions.
```

nojnhuh (approved):
/lgtm
/approve

Thanks @enoodle!

PR #125, opened 2025-09-29
Add mechanism to pass device attributes

dharmjit:
This PR adds a mechanism to pass device attributes to emulate the real devices in local test setups.
- Add --device-attributes CLI flag and DEVICE_ATTRIBUTES env var
- Implement automatic type detection for int, bool, version, string values
- Add Helm template support for device attributes configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,55 @@ Finally, you can run the following to cleanup your environment and delete the
 
 ## Anatomy of a DRA resource driver
 
+### Configuration
+
+The DRA example driver supports several configuration options that can be set via command-line flags, environment variables, or Helm values:
+
+#### Device Configuration
+
+- `--num-devices` / `NUM_DEVICES`: The number of mock GPU devices to create (default: 8)
+- `--device-attributes` / `DEVICE_ATTRIBUTES`: Additional device attributes to be added to resource slices in `key=value` format, separated by commas (default: empty)
+
+The driver automatically detects the value type based on the input:
+- **String**: `productName=NVIDIA GeForce RTX 5090`, `architecture=Blackwell` (default)
+
+Example usage:
+```bash
+# Via command line
+./dra-example-kubeletplugin --num-devices=4 --device-attributes="productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
+
+# Via environment variable
+export DEVICE_ATTRIBUTES="productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
+./dra-example-kubeletplugin --num-devices=4
+
+# Via Helm values.yaml
+kubeletPlugin:
+  numDevices: 4
+  deviceAttributes: "productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
+```
+
+The device attributes will be added to each device in the resource slice alongside the default attributes (index, uuid, model, driverVersion). The resulting resource slice will have properly typed attributes:
+
+```yaml
+- attributes:
+    index:
+      int: 0
+    uuid:
+      string: gpu-18db0e85-99e9-c746-8531-ffeb86328b39
+    model:
+      string: LATEST-GPU-MODEL
+    driverVersion:
+      version: 1.0.0
+    productName:                    # <- Custom string attribute
+      string: NVIDIA GeForce RTX 5090
+    architecture:                   # <- Custom string attribute
+      string: Blackwell
+  capacity:
+    memory:
+      value: 80Gi
+  name: gpu-0
+```
+
 TBD
 
 ## Code Organization
--- a/README.md
+++ b/README.md
@@ -377,56 +377,9 @@ Finally, you can run the following to cleanup your environment and delete the
 
 ## Anatomy of a DRA resource driver
 
-### Configuration
-
-The DRA example driver supports several configuration options that can be set via command-line flags, environment variables, or Helm values:
-
-#### Device Configuration
-
-- `--num-devices` / `NUM_DEVICES`: The number of mock GPU devices to create (default: 8)
-- `--device-attributes` / `DEVICE_ATTRIBUTES`: Additional device attributes to be added to resource slices in `key=value` format, separated by commas (default: empty)
-
-The driver automatically detects the value type based on the input:
-- **String**: `productName=NVIDIA GeForce RTX 5090`, `architecture=Blackwell` (default)
-
-Example usage:
-```bash
-# Via command line
-./dra-example-kubeletplugin --num-devices=4 --device-attributes="productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
-
-# Via environment variable
-export DEVICE_ATTRIBUTES="productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
-./dra-example-kubeletplugin --num-devices=4
-
-# Via Helm values.yaml
-kubeletPlugin:
-  numDevices: 4
-  deviceAttributes: "productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
-```
-
-The device attributes will be added to each device in the resource slice alongside the default attributes (index, uuid, model, driverVersion). The resulting resource slice will have properly typed attributes:
-
-```yaml
-- attributes:
-    index:
-      int: 0
-    uuid:
-      string: gpu-18db0e85-99e9-c746-8531-ffeb86328b39
-    model:
-      string: LATEST-GPU-MODEL
-    driverVersion:
-      version: 1.0.0
-    productName:                    # <- Custom string attribute
-      string: NVIDIA GeForce RTX 5090
-    architecture:                   # <- Custom string attribute
-      string: Blackwell
-  capacity:
-    memory:
-      value: 80Gi
-  name: gpu-0
-```
-
-TBD
+For usage and configuration options, prefer:
+- CLI help: run `./dra-example-kubeletplugin --help` for flags and examples
+- Helm values: consult `deployments/helm/dra-example-driver/values.yaml` for configurable settings and inline docs
 
 ## Code Organization
 
--- a/cmd/dra-example-kubeletplugin/discovery.go
+++ b/cmd/dra-example-kubeletplugin/discovery.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"strconv"
+	"strings"
 
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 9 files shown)


cmd/dra-example-kubeletplugin/main.go
```diff
@@ -94,6 +95,13 @@ func newApp() *cli.App {
 			Destination: &flags.numDevices,
 			EnvVars:     []string{"NUM_DEVICES"},
 		},
+		&cli.StringFlag{
+			Name:        "device-attributes",
+			Usage:       "Additional device attributes to be added to resource slices in key=value format, separated by commas. Example: productName=NVIDIA GeForce RTX 5090,architecture=Blackwell",
```
nojnhuh: Is there a way we could encode the type in here too to eliminate the need to guess? Maybe like `key=type=value`? Or encoding these as JSON or some other format that already knows how to distinguish types? Or maybe take inspiration from Helm's `--set` and `--set-string`? Telling the difference between a string and a version with the latter two options might still be tricky though unless we make the user specify that explicitly like in the first option.

Something like that would also allow setting `true` and `false` as strings instead of always being interpreted as booleans. And likewise allowing strings that look like numbers or versions.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -46,6 +46,7 @@ controller:
 
 kubeletPlugin:
   numDevices: 8
+  deviceAttributes: ""
```
nojnhuh, line 55: Since this isn't obviously tied to the command line flag only from looking at this file, could we include the format and one example here? That might save others from digging around too much to see how to use this.

README.md
```diff
@@ -377,6 +377,55 @@ Finally, you can run the following to cleanup your environment and delete the
...
+      string: NVIDIA GeForce RTX 5090
+    architecture:                   # <- Custom string attribute
+      string: Blackwell
+  capacity:
+    memory:
+      value: 80Gi
+  name: gpu-0
+```
+
 TBD
```
nojnhuh: Maybe @klueska can clarify the original intent, but I envisioned this "Anatomy of a DRA resource driver" section to describe how drivers work in general, not the specific settings of this particular driver that might not apply to at least most other drivers.

If I'm interpreting that right, could we make `### Configuration` an H2 sibling right above "Anatomy of a DRA resource driver"? Or maybe we can take it out entirely if we can port all of this content elsewhere like I suggested might be possible in my comment toward the top of this section.

README.md
```diff
@@ -377,6 +377,55 @@ Finally, you can run the following to cleanup your environment and delete the
 
 ## Anatomy of a DRA resource driver
 
+### Configuration
+
+The DRA example driver supports several configuration options that can be set via command-line flags, environment variables, or Helm values:
```
nojnhuh: IMO this kind of documentation is best kept as close to where these are defined as possible. For command line flags, I'd prefer good descriptions/examples that show up in the generated `--help` (like you've already added in main.go) and for Helm values I'd like to see comments in values.yaml.

Those places are sufficient for basic usage docs like this, easier to maintain as we continue to add more config options. And those places drive where I tend to first look for usage docs (`--help` and `helm show values`).

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -61,6 +77,112 @@ func enumerateAllPossibleDevices(numGPUs int) (AllocatableDevices, error) {
...
 }
 
+// parseDeviceAttributes parses a comma-separated string of key=value pairs
+// and returns a map of device attributes with automatic type detection.
+// Supported value types:
+// - int: integer values (e.g., "count=5")
+// - bool: boolean values (e.g., "enabled=true", "disabled=false")
+// - version: semantic version values (e.g., "driver_version=1.2.3")
+// - string: any other value (e.g., "productName=NVIDIA GeForce RTX 5090", "architecture=Blackwell")
+func parseDeviceAttributes(deviceAttributes string) (map[resourceapi.QualifiedName]resourceapi.DeviceAttribute, error) {
```
nojnhuh: This function seems like a good candidate for some unit tests.

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -61,6 +77,112 @@ func enumerateAllPossibleDevices(numGPUs int) (AllocatableDevices, error) {
...
+	attributes := make(map[resourceapi.QualifiedName]resourceapi.DeviceAttribute)
+
+	if deviceAttributes == "" {
+		return attributes, nil
+	}
+
+	pairs := strings.Split(deviceAttributes, ",")
```
nojnhuh: Is it possible to specify a custom string attribute that contains a comma?

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -61,6 +77,112 @@ func enumerateAllPossibleDevices(numGPUs int) (AllocatableDevices, error) {
...
+		value := strings.TrimSpace(parts[1])
+
+		if key == "" {
+			return nil, fmt.Errorf("device attribute key cannot be empty")
+		}
+
+		// Detect value type and create appropriate DeviceAttribute
+		attr, err := createDeviceAttribute(value)
+		if err != nil {
+			return nil, fmt.Errorf("invalid value for attribute %s: %v", key, err)
```
nojnhuh:
```suggestion
			return nil, fmt.Errorf("invalid value for attribute %s: %w", key, err)
```

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -94,6 +95,13 @@ func newApp() *cli.App {
 			Destination: &flags.numDevices,
 			EnvVars:     []string{"NUM_DEVICES"},
 		},
+		&cli.StringFlag{
```
nojnhuh: Would making this a `StringSliceFlag` let us offload splitting a set of multiple attributes to the arg parser so we don't need a hand-rolled CSV-like thing?

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -61,6 +77,112 @@ func enumerateAllPossibleDevices(numGPUs int) (AllocatableDevices, error) {
...
+	if len(parts) < 3 {
+		return false
+	}
+
+	// Check if first three parts are numeric
+	for i := 0; i < 3; i++ {
+		if _, err := strconv.Atoi(parts[i]); err != nil {
+			return false
+		}
+	}
```
nojnhuh: At least according to the regex from https://semver.org, a value like `1.2.3.4` is not valid but would be interpreted as a version here.

And does this handle values like `1.0.0-beta` where the `-beta` shouldn't be expected to necessarily be a number?

Overall would it be easier to use a library like [github.com/Masterminds/semver/v3](https://pkg.go.dev/github.com/Masterminds/semver/v3) to see if a string is a valid semver?

README.md
dharmjit: sure @nojnhuh, we can add configuration docs in a followup PR if required. I will remove these changes from this PR.

README.md
dharmjit: make sense, I will remove these changes.

cmd/dra-example-kubeletplugin/discovery.go
dharmjit: I've added units tests for this function.

cmd/dra-example-kubeletplugin/discovery.go
dharmjit: CLI: with string slice flag, each `--device-attributes` is treated as complete key=value pair and we do not split by commas now, so its okay to pass values with comma as below
```
--device-attributes 'productName=NVIDIA GeForce RTX 5090' --device-attributes 'notes=foo,bar'
```

Env Vars: is parsed as a comma-separated string (e.g., k=v,k=v), so commas inside values would be misinterpreted as separators

So If someone need commas, don’t use the env var. Use repeated CLI flags instead, or configure your deployment to pass args rather than env vars.

cmd/dra-example-kubeletplugin/discovery.go
dharmjit: makes sense to use a library. I will add this in the subsequent commit.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -46,6 +46,13 @@ controller:
 
 kubeletPlugin:
   numDevices: 8
+  # Additional device attributes to be added to resource slices.
+  # When setting via env var (DEVICE_ATTRIBUTES), provide a comma-separated list of key=value entries:
+  #   DEVICE_ATTRIBUTES: "productName=NVIDIA GeForce RTX 5090,architecture=Blackwell"
+  # Values containing commas are not supported via env var. Prefer repeated CLI flags, e.g.:
+  #   --device-attributes productName=NVIDIA GeForce RTX 5090 \
+  #   --device-attributes architecture=Blackwell
```
nojnhuh, line 54: Let's avoid mentioning exactly how this value is used within the chart since that's an implementation detail. The important details are:
1. What the value does (it sets attributes on the devices)
2. How it's formatted (comma-separated key=value pairs)

Leaving out the other details will make it easier to tweak this in the chart if we find a need to do that.

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -94,6 +95,12 @@ func newApp() *cli.App {
 			Destination: &flags.numDevices,
 			EnvVars:     []string{"NUM_DEVICES"},
 		},
+		&cli.StringSliceFlag{
+			Name:    "device-attributes",
```
nojnhuh, line 99: If each instance of the flag defines one attribute, could we rename this flag?

```suggestion
			Name:    "device-attribute",
```

README.md
```diff
@@ -377,7 +377,9 @@ Finally, you can run the following to cleanup your environment and delete the
 
 ## Anatomy of a DRA resource driver
 
-TBD
+For usage and configuration options, prefer:
+- CLI help: run `./dra-example-kubeletplugin --help` for flags and examples
+- Helm values: consult `deployments/helm/dra-example-driver/values.yaml` for configurable settings and inline docs
```
nojnhuh, line 382: Let's leave this section unchanged for now. There's a lot more we would want to put here and "TBD" is a good enough reminder of that.

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -20,36 +20,53 @@ import (
...
 	"math/rand"
 	"os"
+	"strconv"
+	"strings"
 
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/ptr"
 
+	semver "github.com/Masterminds/semver/v3"
```
nojnhuh, line 30: I see now that the exact library used for CEL expressions in Kubernetes is https://pkg.go.dev/github.com/blang/semver/v4@v4.0.0. I don't imagine we'd see any practical differences with this one, but it'd be nice not to have to worry about that.

cmd/dra-example-kubeletplugin/discovery.go
```diff
@@ -20,36 +20,53 @@ import (
...
 			},
+			"driverVersion": {
+				VersionValue: ptr.To("1.0.0"),
+			},
+		}
+
+		// Add additional attributes from the flag
+		for key, value := range additionalAttributes {
+			attributes[key] = value
+		}
```
nojnhuh, line 65: My editor offers this simplification:

```suggestion
		maps.Copy(attributes, additionalAttributes)
```

cmd/dra-example-kubeletplugin/main.go
nojnhuh: @dharmjit I'm still curious if addressing this issue is worthwhile? I wouldn't block the PR on this, but I think we should at least note the caveats for anyone else who might be looking to implement this in their own drivers.

One other potential issue I could see is that if new types are implemented upstream (like [lists](https://github.com/kubernetes/enhancements/issues/5491)), then implementing those here will be challenging without affecting existing users.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -46,6 +46,13 @@ controller:
...
+  deviceAttributes: ""
```
nojnhuh, line 55: One issue I ran into trying this out is that I'm not sure it's possible to set multiple attributes with Helm's `--set` flag.

e.g. with `--set kubeletPlugin.deviceAttributes="jon=was,here=9,jonagain=1.1,jon3=1.1.1,JON=true,jOn=false"`:
```
% helm get values -n dra-example-driver dra-example-driver
USER-SUPPLIED VALUES:
JON: "true"
here: "9"
jOn: "false"
jon3: 1.1.1
jonagain: "1.1"
kubeletPlugin:
  deviceAttributes: jon=was
```

Maybe we can make the Helm value an array and then build up the comma-separated string in the template where we define `DEVICE_ATTRIBUTES`? That would enable the following (ugly, but serviceable):
```
  --set "kubeletPlugin.deviceAttributes[0]=jon=was" \
  --set "kubeletPlugin.deviceAttributes[1]=here=9" \
  --set "kubeletPlugin.deviceAttributes[2]=jonagain=1.1" \
  --set "kubeletPlugin.deviceAttributes[3]=jon3=1.1.1" \
  --set "kubeletPlugin.deviceAttributes[4]=JON=true" \
  --set "kubeletPlugin.deviceAttributes[5]=jOn=false"
```

This template snippet seems to work:
```yaml
        # Additional device attributes to be added to resource slices.
        - name: DEVICE_ATTRIBUTES
          value: {{ join "," .Values.kubeletPlugin.deviceAttributes | quote }}
```

pohly:
Is this PR still being worked on?

vishalanarase (commented):
needs clean rebase before review it's pending long time

PR #127, opened 2025-11-11
Add KEP-4680 device health reporting to the example driver

harche:
## Summary

Adds KEP-4680 device health reporting to the example driver using the version neutral WatchHealthStatus API from the kubeletplugin helper (which serves both v1 and v1alpha1 DRAResourceHealth). Allocated device health surfaces in `pod.status.containerStatuses[].allocatedResourcesStatus`.

A small HealthSimulator stands in for real device telemetry. Devices report healthy by default, and the periodic poll re-sends the report as the kubelet's freshness heartbeat.

For demos and e2e, annotating the driver pod with `health.example.com/<device>=healthy|unhealthy|unknown` pins a device's health.

Health reporting is on by default. It can be disabled by setting `HEALTH_SERVICE=false` (Helm: `kubeletPlugin.deviceHealth=false`), which makes WatchHealthStatus return ErrHealthNotSupported. This shows how a driver opts out.

Requires Kubernetes 1.36+ (ResourceHealthStatus and ResourceHealthStatusMessage, Beta and on by default). 

AI Usage : AI was used in testing and reviewing the code.


harche:
/hold for the implementation PRs to merge, 

https://github.com/kubernetes/enhancements/issues/4680#issuecomment-3517731091

harche:
> I will update once 1.36 beta is cut

/hold cancel

harche:
> /cc @SergeyKanzhelev @nojnhuh

/hold I am adding a feature to trigger the health status update in this example driver.

harche:
Putting this on hold until #201 merges. Both PRs modify the e2e test infrastructure (`e2e_setup_test.go`, `e2e_test.go`) and will conflict. Will rebase after #201 lands.

/hold

pohly:
As mentioned before, I don't like that DRA drivers have to interact with the gRPC interface (drahealthv1alpha1 currently, despite the beta graduation of the feature?!). Can we first improve the kubeletplugin support for this feature (abstract from underlying gRPC interface similar to how it's done for the main DRA kubelet interace, better documentation) and only then make the DRA example driver use that improved support?

harche:
I tested both DRAResourceHealth API generations end-to-end on kind by temporarily redirecting the `k8s.io` staging modules in `go.mod` to kubernetes master pseudo-versions (which contain the merged `WatchHealthStatus` API from kubernetes/kubernetes#139477):

- **v1alpha1 path**: against a released `kindest/node:v1.36.1` kubelet (pre-v1), the health e2e passes via the helper's v1alpha1 compatibility serving — initial `Healthy`, annotation-forced `Unhealthy`, and recovery back to `Healthy` all surface in `pod.status.containerStatuses[].allocatedResourcesStatus`.
- **v1 path**: against a kubelet built from kubernetes master (`v1.37.0-beta.0.446+4b1379a4a44`), the same e2e passes, and the driver logs confirm the kubelet opened the v1 stream (`/v1.DRAResourceHealth/NodeWatchResources`) rather than falling back to v1alpha1.

The `replace` block in `go.mod` is temporary: once `WatchHealthStatus` ships in a tagged `k8s.io/dynamic-resource-allocation` release, I'll drop the redirect and bump to the release version.

harche:
> As mentioned before, I don't like that DRA drivers have to interact with the gRPC interface (drahealthv1alpha1 currently, despite the beta graduation of the feature?!). Can we first improve the kubeletplugin support for this feature (abstract from underlying gRPC interface similar to how it's done for the main DRA kubelet interace, better documentation) and only then make the DRA example driver use that improved support?

@pohly @nojnhuh this PR is now ready for review. Thanks.

cmd/dra-example-kubeletplugin/health_simulator.go
```diff
@@ -0,0 +1,324 @@
...
+		switch {
+		case recoveryDuration < 30*time.Second:
+			return kubeletplugin.HealthStatusUnhealthy,
+				fmt.Sprintf("%s initiating recovery sequence (diagnostics in progress)", deviceName)
+		case recoveryDuration < 1*time.Minute:
+			return kubeletplugin.HealthStatusUnhealthy,
+				fmt.Sprintf("%s recovery in progress (running self-tests)", deviceName)
+		default:
+			return kubeletplugin.HealthStatusHealthy,
+				fmt.Sprintf("%s recovered successfully, all diagnostics passed", deviceName)
```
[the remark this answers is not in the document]
harche: Thanks, aligned the recovery thresholds and added boundary tests.

deployments/helm/dra-example-driver/templates/role.yaml
```diff
@@ -0,0 +1,14 @@
...
+# namespace, so it is a namespaced Role rather than a cluster-wide grant.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "dra-example-driver.fullname" . }}-pods-role
+  namespace: {{ include "dra-example-driver.namespace" . }}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["watch"]
```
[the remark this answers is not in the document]
harche: Thanks, both the Role and RoleBinding are now conditional on device health being enabled.

demo/examples/health-reporting/health-reporting.yaml
```diff
@@ -0,0 +1,89 @@
...
+# make this example driver's health reporting easy to exercise by hand. It lets
+# anyone who can annotate the driver pod declare any device healthy/unhealthy/
+# unknown, which is a health-control capability. A real driver MUST derive health
+# from actual device telemetry and must NOT ship an annotation-driven override.
+#
+# To force a specific health state, annotate the driver pod:
+#
+#   # Find the driver pod
+#   DRIVER_POD=$(kubectl get pod -n dra-example-driver \
+#     -l app.kubernetes.io/component=kubeletplugin -o name | head -1)
```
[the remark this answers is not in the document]
harche: Thanks, updated the commands to select the driver pod on the workload's node and use the actual allocated device.

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -178,6 +182,31 @@ func newApp() *cli.App {
...
+			Usage:       "Namespace of this driver pod. Used together with pod-name to watch for health override annotations.",
+			Destination: &flags.namespace,
+			EnvVars:     []string{"NAMESPACE"},
+		},
+		&cli.BoolFlag{
+			Name:        "device-health",
+			Usage:       "Enable device health reporting (KEP-4680). When false, the driver advertises no health service to the kubelet and WatchHealthStatus returns ErrHealthNotSupported - the way a driver with no health source opts out.",
+			Value:       true,
+			Destination: &flags.deviceHealth,
+			EnvVars:     []string{"HEALTH_SERVICE"},
```
bart0sh: nit: inconsistency between env var name and option name. I'd suggest to rename the option to --health-service or the env to DEVICE_HEALTH.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -74,13 +154,30 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
+
+	logger.Info("Stopping device health monitoring")
+	// Closing stopHealthCh unblocks the deviceHealthLoop select, tells every
+	// in-flight WatchHealthStatus call to return, and (via the watcher-cancel
+	// goroutine in watchHealthOverrides) closes the pod watch so that loop
+	// unblocks too. Each WatchHealthStatus call deregisters its own subscriber
+	// channel as it exits.
+	close(d.stopHealthCh)
+
+	d.healthWg.Wait()
```
bart0sh: Ths is called unconditionally, but `NewDriver:heathWg.Add`  is called only when `driver.deviceHealth` is true.

bart0sh: Can we make shutdown and setup health-related actions look consistent? Currently setup is:
```
if driver.deviceHealth {
	driver.healthWg.Add(2)
	go driver.deviceHealthLoop(ctx)
	go driver.watchHealthOverrides(ctx)
}
```

and shutdown looks totally different and hard to match.

cmd/dra-example-kubeletplugin/main.go
harche: Thanks, renamed the env to `DEVICE_HEALTH` and added a Helm test assertion so they stay in sync.

cmd/dra-example-kubeletplugin/driver.go
harche: Thanks for looking closely at this. It was safe as written since `Wait()` on a WaitGroup that never had `Add()` called returns immediately, and `stopHealthCh` is always created in `NewDriver`. That said, you're right that it reads oddly next to the guarded startup, and the "Stopping device health monitoring" log line fired even when health reporting was disabled. I've wrapped the log, the channel close, and the `Wait()` in `if d.deviceHealth` so `Shutdown` now mirrors `NewDriver`.

internal/api/checkpoint/v1/zz_generated.conversion.go
```diff
@@ -81,7 +80,7 @@ func Convert_checkpoint_Checkpoint_To_v1_Checkpoint(in *checkpoint.Checkpoint, o
 }
 
 func autoConvert_v1_PreparedClaim_To_checkpoint_PreparedClaim(in *PreparedClaim, out *checkpoint.PreparedClaim, s conversion.Scope) error {
-	out.UID = types.UID(in.UID)
+	*out = *(*checkpoint.PreparedClaim)(unsafe.Pointer(in))
```
bart0sh, line 83: How this change is related to this PR?

harche: @bart0sh Good catch, and thanks for asking. This isn't a hand edit. It's the output of running `make generate` with `conversion-gen` v0.37.0, matching the k8s 1.37 dependency bump from #262. Between code-generator v0.36 and v0.37, conversion-gen started emitting an `unsafe.Pointer` cast for structurally identical types instead of the field-by-field assignment, so the file on `main` is simply stale. I verified this on a fresh clone of `main`: v0.33 through v0.36 reproduce the current file, and v0.37.0 produces exactly this diff. Regenerating on this branch produces no further changes.

Since this is an example driver, I'd lean toward keeping the regenerated file in this PR so `make generate` stays clean on the branch. That said, if you'd prefer to keep this PR focused, I'm happy to drop the change here, send the regeneration as a separate PR, and hold this one until that merges. Just let me know which you prefer.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -52,13 +109,36 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
+	// ErrHealthNotSupported and the kubelet never subscribes.
+	driver.deviceHealth = config.flags.deviceHealth
+	if driver.deviceHealth {
+		for deviceName := range state.allocatable {
+			driver.devices = append(driver.devices, deviceName)
+		}
+		sort.Strings(driver.devices)
+
+		driver.simulator = NewHealthSimulator(driver.devices, config.flags.simulateHealthChanges)
+		klog.Infof("Device health reporting enabled for %d devices (auto-simulation: %t)",
```
bart0sh: Can you use contextual logging here? ctx is available in this function.

cmd/dra-example-kubeletplugin/health_simulator_test.go
```diff
@@ -0,0 +1,310 @@
...
+	sim := NewHealthSimulator(names, true)
+
+	var wg sync.WaitGroup
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for j := 0; j < 100; j++ {
+				health, msg := sim.GetDeviceHealth(names[j%len(names)])
+				require.True(t,
```
bart0sh: This will terminate go-routine, not the test. It would be safer to use `assert` here and below.

cmd/dra-example-kubeletplugin/driver.go
harche: Thanks, done. Switched both messages in `NewDriver` to `klog.FromContext(ctx)` with structured key/value pairs. Those were the only non-contextual calls in the PR; `make logcheck` is now clean across the repo.

cmd/dra-example-kubeletplugin/health_simulator_test.go
harche: Good catch, thanks. Switched both calls in `TestGetDeviceHealth_ConcurrentAccess` to `assert` since `FailNow` is only valid on the test goroutine. I checked the other goroutines in the PR's tests and none of them assert directly; they hand results back over channels and assert after `wg.Wait()`, so this was the only instance.

bart0sh:
@harche 

It would be nice to increase test coverage of the new code, currently it's not that impressive:

```
$ go tool cover -func=coverage.out | grep 'driver.go\|health.go'
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:94:              NewDriver                       0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:168:             Shutdown                        0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:190:             PrepareResourceClaims           0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:202:             prepareResourceClaim            0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:235:             UnprepareResourceClaims         0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:247:             unprepareResourceClaim          0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:260:             HandleError                     0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:270:             deviceHealthLoop                0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:308:             watchHealthOverrides            0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:378:             applyHealthOverrides            100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:437:             pollDeviceHealth                100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:464:             buildHealthReport               100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:484:             notifySubscribers               100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:499:             WatchHealthStatus               85.7%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:545:             sendReport                      60.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/health.go:48:              startHealthcheck                0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/health.go:118:             Stop                            0.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/health.go:127:             Check                           0.0%
```

harche:
> It would be nice to increase test coverage of the new code, currently it's not that impressive:

@bart0sh Thanks, added a few tests and the coverage is up now.

```
$ go tool cover -func=coverage.out | grep 'driver.go\|health.go'
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:99:              NewDriver                       95.1%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:176:             Shutdown                        100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:198:             PrepareResourceClaims           100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:210:             prepareResourceClaim            100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:243:             UnprepareResourceClaims         100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:255:             unprepareResourceClaim          100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:268:             HandleError                     100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:278:             deviceHealthLoop                100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:319:             watchHealthOverrides            91.7%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:389:             applyHealthOverrides            100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:448:             pollDeviceHealth                100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:475:             buildHealthReport               100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:495:             notifySubscribers               100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:510:             WatchHealthStatus               100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/driver.go:556:             sendReport                      100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/health.go:48:              startHealthcheck                89.4%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/health.go:118:             Stop                            100.0%
sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin/health.go:127:             Check                           93.1%
```

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -20,30 +20,95 @@ import (
...
+	// advertises no health service, starts no simulator or health goroutines,
+	// and WatchHealthStatus returns ErrHealthNotSupported.
+	deviceHealth bool
+	simulator    *HealthSimulator
+	// devices holds the device names, sorted, for stable report/log iteration.
+	// It is immutable after startup and so needs no lock.
+	devices []string
+
+	// healthMu guards the two device-health bookkeeping maps below.
+	healthMu sync.RWMutex
```
bart0sh: I don't see any usage of `RLock` with this. Should it be `sync.Mutex`?

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -74,13 +164,33 @@ func NewDriver(ctx context.Context, config *Config) (*driver, error) {
...
+	// Mirror the guarded startup in NewDriver: the health goroutines only
+	// exist when device health reporting is enabled.
+	if d.deviceHealth {
+		logger.Info("Stopping device health monitoring")
+		// Closing stopHealthCh unblocks the deviceHealthLoop select, tells every
+		// in-flight WatchHealthStatus call to return, and (via the watcher-cancel
+		// goroutine in watchHealthOverrides) closes the pod watch so that loop
+		// unblocks too. Each WatchHealthStatus call deregisters its own subscriber
+		// channel as it exits.
+		close(d.stopHealthCh)
```
bart0sh: nit: This makes Shutdown function non-idempotent. Second attempt to close this channel would panic. Up to you to fix it or leave as it is, just a nit.

cmd/dra-example-kubeletplugin/driver.go
```diff
@@ -168,3 +274,293 @@ func (d *driver) HandleError(ctx context.Context, err error, msg string) {
...
+				case <-d.stopHealthCh:
+					watcher.Stop()
+					return
+				default:
+				}
+
+				// Only additions and modifications carry the current annotation
+				// set. Ignore other event types (e.g. Bookmark, which carries an
+				// annotation-stripped object that would spuriously clear overrides).
+				if event.Type != watch.Added && event.Type != watch.Modified {
```
bart0sh, line 380: This filters out watch.Error, which probably should be somehow processed. watch.Deleted is also dropped, but I'm not sure if we want/need to process it.

bart0sh:
@harche lgtm. I've added some comments, but they're minor.

cmd/dra-example-kubeletplugin/driver.go
harche: Good catch, thanks. You're right, nothing in the driver takes a read lock; the only `RLock` callers were two test helpers. Switched it to `sync.Mutex` and updated the tests to match.

cmd/dra-example-kubeletplugin/driver.go
harche: Thanks, fixed.

cmd/dra-example-kubeletplugin/driver.go
harche: Thanks, agreed on `watch.Error`. It's now logged via `apierrors.FromObject` and the loop stops the watcher and reconnects through the existing backoff, rather than relying on the server closing the channel. For `watch.Deleted`, the watch is on the driver's own pod, so a delete means this pod is being torn down and any override state dies with the process. I left it ignored and added a comment explaining that.

harche:
> @harche lgtm. I've added some comments, but they're minor.

Thanks @bart0sh for your review. 


/cc @pohly @nojnhuh  for approval.

harche:
> /approve

Thanks @nojnhuh , 

@bart0sh I think the tide needs `lgtm` label.

bart0sh:
> @bart0sh I think the tide needs `lgtm` label.

I believe it asks to resolve conflict in go.mod

harche:
> > @bart0sh I think the tide needs `lgtm` label.
> 
> I believe it asks to resolve conflict in go.mod

Thanks @bart0sh , looks like a few changes over the weekend crossed it's path. Rebased and pushed.

Issue #101, opened 2025-05-19
Demonstrate how to update ResourceClaim status


nojnhuh:
The example driver should show how status can be updated with server-side apply, e.g. to show network device drivers how to record allocated IP addresses.

https://kubernetes.slack.com/archives/C0409NGC1TK/p1747668302074819

guptaNswati:
I will do this. did it for nvidia-dra-driver health check https://github.com/NVIDIA/k8s-dra-driver-gpu/pull/545#issuecomment-3272419830 following https://github.com/google/dranet/pull/78

nojnhuh:
Still WIP

/remove-lifecycle stale

PR #128, opened 2025-11-13
add device data in resourceclaim

guptaNswati:
Addressing https://github.com/kubernetes-sigs/dra-example-driver/issues/101 to add device attributes to demonstrate how to do resourceclaim status update. 

Test run
```
    devices:
    - conditions:
      - lastTransitionTime: "2025-11-13T22:22:56Z"
        message: ""
        reason: GPUDeviceReady
        status: "True"
        type: Ready
      data:
        driverVersion:
          version: 1.0.0
        uuid:
          string: gpu-18db0e85-99e9-c746-8531-ffeb86328b39
      device: gpu-0

  devices:
    - conditions:
      - lastTransitionTime: "2025-11-13T22:22:56Z"
        message: ""
        reason: GPUDeviceReady
        status: "True"
        type: Ready
      data:
        driverVersion:
          version: 1.0.0
        uuid:
          string: gpu-93d37703-997c-c46f-a531-755e3e0dc2ac
      device: gpu-1
```


cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
 
 	return resultConfigs, nil
 }
+
+func (s *DeviceState) buildDeviceStatus(res *resourceapi.DeviceRequestAllocationResult) *resourceapply.AllocatedDeviceStatusApplyConfiguration {
```
nojnhuh: Does `res` need to be a pointer, or can we save the extra syntax and theoretical nil-dereference by making this a plain value? Not a blocking issue, but I may propose refactoring this when I integrate with #129.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
...
+	dn := res.Device
+	deviceInfo := make(map[string]interface{})
```
nojnhuh: nit: could we make this more strongly typed?


```suggestion
	deviceInfo := make(map[string]resourceapi.DeviceAttribute)
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
...
+		WithReason("GPUDeviceReady").
+		WithMessage("GPUDeviceAllocated").
+		WithLastTransitionTime(metav1.Now())
+
+	return resourceapply.AllocatedDeviceStatus().
+		WithDevice(dn).
+		WithDriver(res.Driver).
+		WithPool(res.Pool).
+		WithConditions(cond).
+		WithData(data)
```
nojnhuh: For my own understanding, who or what generally consumes this data?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
...
+	if err != nil {
+		klog.Errorf("Failed to marshal device data: %v", err)
+		jsonBytes = []byte("{}")
+	}
+	data := runtime.RawExtension{
+		Raw: jsonBytes,
+	}
+	cond := metav1apply.Condition().
+		WithType("Ready").
+		WithStatus(metav1.ConditionTrue).
```
nojnhuh: Does the NVIDIA driver also only set this condition at the time the claim is allocated, or can it ever change after that? If it can change, I think the example driver should model how to update that. Even if it stays constant, scaffolding out how that can work with something that says "this is where the latest status is determined" would be helpful.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
...
+		WithData(data)
+}
+
+func (s *DeviceState) applyDeviceStatus(ctx context.Context, ns, name string, devices ...*resourceapply.AllocatedDeviceStatusApplyConfiguration) error {
+	claim := resourceapply.ResourceClaim(name, ns).
+		WithStatus(resourceapply.ResourceClaimStatus().WithDevices(devices...))
+
+	opts := metav1.ApplyOptions{FieldManager: consts.DriverName, Force: true}
+
+	_, err := s.config.coreclient.ResourceV1().ResourceClaims(ns).ApplyStatus(ctx, claim, opts)
```
nojnhuh: Is there some lightweight verification we can add to the e2e tests? At least something to verify that something like the condition or one of the attributes is set on even one of the examples would make sure this doesn't totally break later.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
...
+
+func (s *DeviceState) buildDeviceStatus(res *resourceapi.DeviceRequestAllocationResult) *resourceapply.AllocatedDeviceStatusApplyConfiguration {
+	dn := res.Device
+	deviceInfo := make(map[string]interface{})
+
+	if d, ok := s.allocatable[dn]; ok {
+		if d.Attributes != nil {
+			attributes := d.Attributes
+
+			if uuid, ok := attributes["uuid"]; ok {
```
nojnhuh: nit: indexing into a nil map doesn't panic, so we can skip this check. I at least don't see any issues skipping it if I remove all of the attributes from the devices and allocate them.

```suggestion
		if uuid, ok := d.Attributes["uuid"]; ok {
```

cmd/dra-example-kubeletplugin/state.go
guptaNswati: so for nvidia-dra-driver i was looking to use this as part of GPU health check status update. But we dint really see any value of showing this information as part of an allocated claim as it does not really change. Its more complex to show any real updates on an allocated claim.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: sure.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: To me, this seems useful for monitoring and debugging when admin want to know pod to gpu mapping, which GPU is being used by a particular pod. right now this info is not readily available.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +400,56 @@ func GetOpaqueDeviceConfigs(
...
+				deviceInfo["uuid"] = uuid
```
nojnhuh: Taking another look at this, I'm not sure I see the value in propagating attributes that already exist in the ResourceSlice. Is there a common use case where deriving these attributes from the ResourceSlice doesn't work?

Or is there some other data we can use here? Maybe like the time-slicing/space-partitioning config? I could see that being useful to model the opaque config being like a request, then the device status would represent the response in cases where a given config might result in several different outcomes. Kind of like the regular spec/status model.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: Makes sense. I think for setting device conditions, the interesting part isn't the API call to update the condition, but actually calculating that value of the condition asynchronously. I think I'd rather include that (or no condition at all) than a hardcoded condition added during `NodePrepareResources`.

nojnhuh:
FYI @guptaNswati I was hoping to get this PR merged before #129 to keep you from having to deal with those changes here. There are other changes in the works that are based on that PR though, so I may elect to merge that one first. If #129 merges first, I'm happy to help resolve conflicts with your PR here!

cmd/dra-example-kubeletplugin/state.go
guptaNswati: Make sense. i need to see whats more valuable to add here. My thought was to have this as an example.

guptaNswati:
> FYI @guptaNswati I was hoping to get this PR merged before #129 to keep you from having to deal with those changes here. There are other changes in the works that are based on that PR though, so I may elect to merge that one first. If #129 merges first, I'm happy to help resolve conflicts with your PR here!

Oh i see. thank you for the headsup. i will see if i can update this tomorrow or else you can go ahead and merge #129. Main blocker is the e2e, right? we can't merge this without the tests?

cmd/dra-example-kubeletplugin/state.go
nojnhuh: Actually I suppose attributes here could be useful in case the device is removed from a ResourceSlice because it became unhealthy or otherwise. Let's keep them here, but can we include a comment explaining when recording the attributes in the status is useful?

cmd/dra-example-kubeletplugin/state.go
guptaNswati: done.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: done.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: fixed.

nojnhuh:
@guptaNswati Have you pushed your recent changes? The most recent commit I see here is from when the PR was first opened.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: Okay since we don't have device health checking rn, i can just change it to just `allocated` for example purposes with a comment or remove it. 
```
// TODO: This condition currently reflects only that the device was allocated, not that
// it is healthy or ready for use. True readiness should be computed asynchronously by
// the health-monitoring pipeline (e.g., NVML Xid/error tracking). In the future, this
// should be replaced—or augmented—with a condition whose value is driven by actual
// device health state rather than allocation alone.
cond := metav1apply.Condition().
    WithType("Allocated").
    WithStatus(metav1.ConditionTrue).
    WithReason("GPUDeviceAllocated").
    WithMessage("GPU device allocated for this request").
    WithLastTransitionTime(metav1.Now())
```

what do you think?

nojnhuh: I would still prefer to either omit this condition entirely or create all of the scaffolding around driving the condition asynchronously. I don't think we should be implementing patterns that we don't recommend, even if we document the right way.

If we decide to omit the condition here, let's open an issue to track adding async conditions.

guptaNswati: okay sounds good. than i will remove it for now as it requires more work and i will open an issue and work on it.

test/e2e/e2e.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+!/usr/bin/env bash
```
nojnhuh: We should change this back.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: > which GPU is being used by a particular pod. right now this info is not readily available.

Is this accurate? Don't other parts of the ResourceClaim status already record this? This is the status I see without these changes:

```yaml
status:
  allocation:
    devices:
      results:
      - device: gpu-0
        driver: gpu.example.com
        pool: dra-example-driver-cluster-worker
        request: gpu
    nodeSelector:
      nodeSelectorTerms:
      - matchFields:
        - key: metadata.name
          operator: In
          values:
          - dra-example-driver-cluster-worker
  reservedFor:
  - name: pod0
    resource: pods
    uid: 39644f38-618f-49da-8948-2f793869ee69
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -380,3 +399,52 @@ func GetOpaqueDeviceConfigs(
...
+		WithData(data)
+}
+
+func (s *DeviceState) applyDeviceStatus(ctx context.Context, ns, name string, devices ...*resourceapply.AllocatedDeviceStatusApplyConfiguration) error {
+	claim := resourceapply.ResourceClaim(name, ns).
+		WithStatus(resourceapply.ResourceClaimStatus().WithDevices(devices...))
+
+	opts := metav1.ApplyOptions{FieldManager: consts.DriverName, Force: true}
+
+	_, err := s.config.coreclient.ResourceV1().ResourceClaims(ns).ApplyStatus(ctx, claim, opts)
```
nojnhuh: This is the only place AFAIK that the driver has a hard dependency on the v1 DRA APIs. The normal way to get this working with v1beta1 and v1beta2 is to use the converting wrapper: https://pkg.go.dev/k8s.io/dynamic-resource-allocation@v0.34.2/client#Client

The `ApplyStatus` (and the rest of the patch methods) aren't implemented by that though because that would require converting patch data between API versions, which isn't feasible AFAIK. `UpdateStatus` is implemented though. Could we try using that? I know in the issue I mentioned server-side apply, but I think handling the other API versions right now is a better tradeoff.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -211,6 +225,11 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
 		}
 	}
 
+	klog.Infof("Adding device attribute to claim %s/%s", claim.Namespace, claim.Name)
+	if err := s.applyDeviceStatus(ctx, claim.Namespace, claim.Name, devicesStatus...); err != nil {
+		klog.Warningf("Failed to update device attributes for claim %s/%s: %v", claim.Namespace, claim.Name, err)
+	}
```
nojnhuh: Could we wrap this in a toggle that can be set by users? The SIG-Scalability tests are one place where we may want to limit any extra API calls.

FYI @alaypatel07

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -211,6 +225,11 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
 		}
 	}
 
+	klog.Infof("Adding device attribute to claim %s/%s", claim.Namespace, claim.Name)
```
nojnhuh:
```suggestion
	klog.Infof("Adding device attributes to claim %s/%s", claim.Namespace, claim.Name)
```

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -211,6 +225,11 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (Prepared
...
+	if err := s.applyDeviceStatus(ctx, claim.Namespace, claim.Name, devicesStatus...); err != nil {
+		klog.Warningf("Failed to update device attributes for claim %s/%s: %v", claim.Namespace, claim.Name, err)
```
nojnhuh: I don't think we need to address this now, but we should think about how to handle errors here besides only logging them. Retries will help in some cases (HTTP 409), but we should prevent blocking subsequent `NodePrepareResources` calls on retries that may never succeed and I'm not sure what an appropriate amount of time would be to spend on that. I wonder if a workqueue running async would be the most appropriate way to handle status updates?

@guptaNswati Could you please open an issue to track improvements to error handling here?

deployments/helm/dra-example-driver/templates/clusterrole.yaml
```diff
@@ -8,6 +8,9 @@ rules:
 - apiGroups: ["resource.k8s.io"]
   resources: ["resourceclaims"]
   verbs: ["get"]
+- apiGroups: ["resource.k8s.io"]
+  resources: ["resourceclaims/status"]
+  verbs: ["patch", "update"]
```
nojnhuh: Can we get away with only one or the other, or do we need both if we only use `ApplyStatus` or `UpdateStatus`?

test/e2e/e2e.sh
guptaNswati: sorry typo.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: this is correct, but what is gpu-0? without looking at resource slice, we dont know the uuid or any attributes of gpu-0.

deployments/helm/dra-example-driver/templates/clusterrole.yaml
guptaNswati: Initially i tested with just `Patch` permission and got permission error. update is needed to actually apply the patch

cmd/dra-example-kubeletplugin/state.go
guptaNswati: a feature gate or an env?

cmd/dra-example-kubeletplugin/state.go
guptaNswati: yes let me look into it. this also ties back to your RBAC permissions comment. rn, for `ApplyStatus` we need both `patch and update`

cmd/dra-example-kubeletplugin/state.go
nojnhuh: Ah, so not just the mapping, but also the metadata. Makes sense.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: I think of feature gates as things that will eventually become default behavior and can't be disabled, but this is something I could see us wanting to control indefinitely, so I think the established CLI flag + env var combo would be better.

deployments/helm/dra-example-driver/templates/clusterrole.yaml
nojnhuh: Thanks for checking!

cmd/dra-example-kubeletplugin/state.go
guptaNswati: yes.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: done https://github.com/kubernetes-sigs/dra-example-driver/issues/133

alaypatel07: You also need this to de-couple the allocation times from apiserver update latency. May be as a hacky first implementation you can do this in a goroutine?

I guess the main question to ask is, is this update required for the driver to declare allocation as complete or is the information in this update auxiliary to allocation?

cmd/dra-example-kubeletplugin/state.go
alaypatel07: Do we need to emit metrics here? I.e. how long the update took, how many times the update has been retried, et.el? If we expect this to be the guide for writing production drivers these metrics could be helpful

cmd/dra-example-kubeletplugin/state.go
alaypatel07: +1 to cli flag and env var. Do we intend to make this switch default to on or off? My preference will be off so we don't have to change anyone who's consuming this driver in their tests (for example the scalability tests in one such consumer I can imagine there being others)

nojnhuh: Default off is fine by me.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: I agree metrics here would be valuable. Let's not block this PR on them though. I opened #134 to track that.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: done.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: To-do.

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -147,6 +148,12 @@ func newApp() *cli.App {
 			Destination: &flags.driverName,
 			EnvVars:     []string{"DRIVER_NAME"},
 		},
+		&cli.BoolFlag{
+			Name:        "device-attribute",
```
nojnhuh: Could we call this something like `gpu-device-status` instead to avoid confusing with #125?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -334,3 +359,61 @@ func GetOpaqueDeviceConfigs(
 
 	return resultConfigs, nil
 }
+
+func (s *DeviceState) buildDeviceStatus(res resourceapi.DeviceRequestAllocationResult) resourceapi.AllocatedDeviceStatus {
```
nojnhuh: Let's move this logic into the `gpu` profile so we can implement this differently for different kinds of devices. See #129 for more about the "profile" concept.

Ideally we shouldn't have to implement this for every kind of device.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -103,6 +111,7 @@ func NewDeviceState(config *Config) (*DeviceState, error) {
 		checkpointManager: checkpointManager,
 		configDecoder:     decoder,
 		configHandler:     configHandler,
+		config:            config,
```
nojnhuh: Let's keep `NewDeviceState` where we pull everything out of the `Config` that we need vs. storing the whole `Config` on the `DeviceState`. `Config` contains user-facing config like command line flags, so let's not plumb that too far down.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: done. resolving it for now. it helps in focussing the remaining items.

cmd/dra-example-kubeletplugin/state.go
guptaNswati: okay. looking to it. need sometime to understand the refractor. will be back at it.

nojnhuh:
We'll also need to handle this to support Kubernetes 1.36: https://github.com/kubernetes/kubernetes/issues/138149

guptaNswati:
@nojnhuh sorry took a long time to address this https://github.com/kubernetes-sigs/dra-example-driver/pull/128#discussion_r2765929565

Had to deal with rebase and refractor. New changes are tested and mainly in `gpu.go`. PTAL. I dont wana loose it for months again. 

```
go test -tags=e2e -v ./test/e2e/... -ginkgo.focus="should allocate 1 distinct GPU per pod"
=== RUN   TestE2e
Running Suite: E2E Suite - /home/nvidia/dra-example-driver/test/e2e
===================================================================
Random Seed: 1779489074

Will run 1 of 14 specs
•SSSSSSSSSSSSS

Ran 1 of 14 Specs in 20.226 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 13 Skipped
--- PASS: TestE2e (20.23s)
PASS
ok  	sigs.k8s.io/dra-example-driver/test/e2e	20.269s
```

test/e2e/e2e_setup_test.go
nojnhuh, line 1: @willie-yao Could you please take a look at the e2e changes here?

test/e2e/e2e_setup_test.go
```diff
@@ -535,3 +537,52 @@ func verifySharedGPUGroup(ctx context.Context, namespace string, group sharingGr
...
+
+				g.Expect(data).To(HaveKey("uuid"),
+					"ResourceClaim %s/%s device %s status.data is missing uuid: %v", namespace, *rcs.ResourceClaimName, d.Device, data)
+				g.Expect(data["uuid"].StringValue).NotTo(BeNil())
+				g.Expect(*data["uuid"].StringValue).NotTo(BeEmpty())
+
+				g.Expect(data).To(HaveKey("driverVersion"),
+					"ResourceClaim %s/%s device %s status.data is missing driverVersion: %v", namespace, *rcs.ResourceClaimName, d.Device, data)
+				g.Expect(data["driverVersion"].VersionValue).NotTo(BeNil())
+				g.Expect(*data["driverVersion"].VersionValue).NotTo(BeEmpty())
```
willie-yao, line 582: Earlier there is a comment saying GPUDeviceStatus includes uuid, model, and driverVersion, but it seems like model isn't being asserted here.

guptaNswati: good catch. fixing it

guptaNswati: @willie-yao done. PTAL

test/e2e/e2e_setup_test.go
willie-yao: looks good! comment below was addressed

nojnhuh (approved):
/lgtm
/approve

PR #129, opened 2025-11-21
Introduce device profiles

### LINKED ISSUE #92: demonstrating advanced features
We want to demonstrate advanced features like partitionable devices with the example driver. This begs the question: should we use some artificial, fake devices or mirror the behavior of some real vendor driver when it advertises real hardware?

On the one hand, we prefer to stay vendor-neutral. On the other hand, a real example is easier to understand and would make the example driver more useful for realistic benchmarking.

After some discussions at KubeCon and in https://kubernetes.slack.com/archives/C0409NGC1TK/p1743667181489199, here's a proposal. In the top-level README.md, we add a new section:

### Configuration

Vendors are encouraged to work with the Kubernetes maintainers to enhance DRA for their use cases. When this leads to new features, extending the example driver such that it demonstrates those features by emulating a vendor driver for certain hardware is desirable. Later, adding novel usages of existing features may also be worth extending the example driver.

At the moment, the driver supports the following profiles:
- **gpu** (default): 8 generic GPUs per node. Works on Kubernetes >= 1.32.
- **nvidia-mig**: two [NVIDIA A100 GPUs](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4815-dra-partitionable-devices#putting-it-all-together-for-the-mig-use-case) per node, with attributes that are the same as for real hardware. Works on Kubernetes >= 1.33.
- **google-tpu**:  models [multi-host devices](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4815-dra-partitionable-devices#defining-multi-host-devices). Works on Kubernetes >= 1.33.

Each deployment of the example driver uses exactly one profile and `<profile name>.dra.example.com` as driver name. To configure the profile, ... [TBD]. These profiles do not actually emulate any real hardware. Instead, they merely inject environment variables which mirror the devices that were allocated.

/assign @bg-chun

nojnhuh:
This PR lays the groundwork for the example driver to be able to demonstrate several different kinds of devices, each implemented as a separate "profile." The changes mostly involve extracting the GPU-specific details and exposing handles to plug in different logic.

To see how this works with multiple profiles, I have a POC based on these changes implementing a new `gpupart` profile which demonstrates partitionable devices: https://github.com/nojnhuh/dra-example-driver/compare/profiles...gpupart

Fixes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,20 @@ Finally, you can run the following to cleanup your environment and delete the
 ./demo/delete-cluster.sh
 ```
 
+## Device Profiles
+
+The example driver can manage several different kinds of devices to demonstrate
+a variety of DRA features. The functionality for each kind of device is
+organized into a "profile." Only one profile is active at a time for a given
+instance of the example driver, though the example driver may be installed
+multiple times in the same cluster with different active profiles. See the Helm
+chart's `deviceProfile` value in values.yaml for available profiles.
+
+For driver developers, this pattern is specific to the example driver and not
+intended to be a recommendation for all DRA drivers. Other drivers will likely
+be simpler by implementing their logic more directly than through an
+abstraction like the example driver's profiles.
+
 ## Anatomy of a DRA resource driver
 
 TBD


nojnhuh:
I wanted to open this PR before I'm off on vacation next week, but I intend to take a look at #128 first and incorporate those changes here before merging this. FYI @guptaNswati

/hold

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -25,26 +34,9 @@ serviceAccount:
...
-    operator: Exists
-    effect: NoSchedule
-  affinity: {}
-  containers:
-    controller:
-      securityContext: {}
-      resources: {}
-
 kubeletPlugin:
+  # numDevices describes how many GPUs to advertise on each node when the "gpu"
```
sunya-ch, line 38: I think we may want to pass this numDevices also to other profiles. Is there any reason to limit to gpu profile?

cmd/dra-example-webhook/main_test.go
```diff
@@ -168,7 +170,10 @@ func TestResourceClaimValidatingWebhook(t *testing.T) {
 		},
 	}
 
-	s := httptest.NewServer(newMux())
+	sb := gpu.ConfigSchemeBuilder
+	assert.NoError(t, sb.AddToScheme(configScheme))
+
+	s := httptest.NewServer(newMux(newConfigDecoder(), gpu.ValidateConfig, driverName))
```
sunya-ch: Are we expecting each profile to have its own ValidateConfig function? If so, we could define it as an abstract method on the profile struct.

deployments/helm/dra-example-driver/values.yaml
nojnhuh: I think eventually we should namespace these options for each profile, so this might eventually turn into something like `--gpu-num-devices`. Then another profile would have a separate `--myprofile-num-devices`. I imagine some parameters might be ambiguous and make it hard to keep track of how which options apply to which profiles unless each profile has its own set of options.

cmd/dra-example-webhook/main_test.go
nojnhuh: > @nojnhuh Could we define a `DeviceProfile` interface/struct with abstract methods like GetDevices, ApplyConfig, and ValidateConfig, and then keep a mapping from profile name to its implementation? If we take that approach, I could implement ConsumableNetDevProfile or ConsumableGPUProfile independently. What do you think?

I opted not to group everything into a single `Profile` struct or interface since the parts aren't ever really used together. e.g. The webhook only needs the `SchemeBuilder` and a validate function; the kubelet plugin doesn't need a validate function but does need a way to enumerate devices which is configurable (how many devices). Keeping those implementation extension points separate is simpler IMO for now.

Another profile can still be implemented this way like I've done here: https://github.com/nojnhuh/dra-example-driver/compare/profiles...gpupart. There are probably opportunities to make instantiating profiles easier, but I don't think we need to solve that right now if those changes wouldn't be user-facing.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -23,53 +23,42 @@ import (
...
+	"sigs.k8s.io/dra-example-driver/internal/profiles"
 )
 
 type AllocatableDevices map[string]resourceapi.Device
-type PreparedDevices []*PreparedDevice
-type PreparedClaims map[string]PreparedDevices
-type PerDeviceCDIContainerEdits map[string]*cdiapi.ContainerEdits
+type PreparedClaims map[string]profiles.PreparedDevices
+
+type ApplyConfigFunc func(cconfig runtime.Object, results []*resourceapi.DeviceRequestAllocationResult) (profiles.PerDeviceCDIContainerEdits, error)
```
sunya-ch:
```suggestion
type ApplyConfigFunc func(config runtime.Object, results []*resourceapi.DeviceRequestAllocationResult) (profiles.PerDeviceCDIContainerEdits, error)
```

cmd/dra-example-webhook/main_test.go
sunya-ch: @nojnhuh Sure. I can also follow the existing switch–case and function-variable pattern, especially since we may only add one or two profile handlers in the near term and I now understand which functions need to be defined for each profile. However, even though this is not driver-user facing, I still believe that using an interface provides a clearer structure and helps reduce contributor and reviewer effort, and the potential for human error in the long term.

Anyway, I don't want to block this PR. Please feel free to proceed with the current implementation once reviewer approve it (I don't think I have an authority to do so). I can raise an extended proposal to hear more opinions about that later.

---

To give a concrete idea of my proposal about interfacing it, I think that the current implementation already defines two functions that represent general behavior across profiles: ApplyConfigFunc and enumerateDevicesFunc. They are defined in different ways.  ApplyConfigFunc is defined as type as `type ApplyConfigFunc func(config runtime.Object, results []*resourceapi.DeviceRequestAllocationResult) (profiles.PerDeviceCDIContainerEdits, error)` while enumerateDevicesFunc is defined in more adhoc way here are there as `func() (resourceslice.DriverResources, error)`and the EnumerateDevices implementation is the function that takes some flags and return a function.

In my understanding, both functions are expected to be set inside the profile switch/case (and no default function set). By introducing an interface, we can ensure that both EnumerateDevices and ApplyConfig are consistently implemented and callable for any profile, rather than conditionally defined within each switch branch. This also removes the need for per-profile conditional setup inside the switch/case and gives us a consistent execution path for all profiles.

For example:

```go
type DeviceProfileHandler interface {
    ApplyConfig(
        config runtime.Object,
        results []*resourceapi.DeviceRequestAllocationResult,
    ) (profiles.PerDeviceCDIContainerEdits, error)

    EnumerateDevices(flags *Flags) (resourceslice.DriverResources, error)
}
...
var validProfiles = map[string]DeviceProfileHandler{
	gpu.ProfileName: GPUProfile,
}
...
	config := &Config{
		flags:               flags,
		coreclient:          clientSets.Core,
		configScheme:        configScheme,
		cdiClass:            cdiClass,
		deviceProfileHander: DeviceProfileHandler,
	}
	...
	if handler, found := validProfiles[profileName]; found {
	        config.deviceProfileHander = handler
	}
```

For the Validate function, if it is optional. we can simply skip it.

deployments/helm/dra-example-driver/values.yaml
sunya-ch: I see your point. I initially thought the `num-devices` is a common parameter that should be set for all profiles but that might not be true. Some profile may come with specific amount of devices in enumeration. So, we are expecting to replace `numDevices` with something like

```yaml
kubeletPlugin:
   profileConfigs:
     gpu:
       numDevices:
```

and parsed this to `gpu-num-devices` in the future?

nojnhuh: Yes, that's one option. I'd rather address that separately from this PR though.

cmd/dra-example-webhook/main_test.go
nojnhuh: I've given refactoring this a try. Let me know if that's closer to what you're describing here.

internal/profiles/profiles.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// SchemeBuilder implements [ConfigHandler].
+func (n NoopConfigHandler) SchemeBuilder() runtime.SchemeBuilder {
+	return runtime.NewSchemeBuilder()
+}
+
+// Validate implements [ConfigHandler].
+func (n NoopConfigHandler) Validate(config runtime.Object) error {
+	return errors.New("no configuration allowed")
```
sunya-ch: Is it expected behavior to return an error? The error message confuse me a bit whether it is allowed or error.

internal/profiles/profiles.go
```diff
@@ -0,0 +1,79 @@
...
+	// ApplyConfig applies a configuration to a set of device allocation results.
+	ApplyConfig(config runtime.Object, results []*resourceapi.DeviceRequestAllocationResult) (PerDeviceCDIContainerEdits, error)
+}
+
+// NoopConfigHandler implements a [ConfigHandler] that allows no configuration.
+type NoopConfigHandler struct{}
+
+// ApplyConfig implements [ConfigHandler].
+func (n NoopConfigHandler) ApplyConfig(config runtime.Object, results []*resourceapi.DeviceRequestAllocationResult) (PerDeviceCDIContainerEdits, error) {
+	return nil, errors.New("no configuration allowed")
```
sunya-ch: same as Validate. The wording confuse me a bit.

internal/profiles/gpu/gpu.go
```diff
@@ -0,0 +1,194 @@
...
+func hash(s string) int64 {
+	h := int64(0)
+	for _, c := range s {
+		h = 31*h + int64(c)
+	}
+	return h
+}
+
+// SchemeBuilder implements [profiles.ConfigHandler].
+func (p Profile) SchemeBuilder() runtime.SchemeBuilder {
```
sunya-ch, line 119: It might be more straightforward to have ConfigHandler implemented separately from Profile.

cmd/dra-example-webhook/main_test.go
sunya-ch: @nojnhuh Yes, it is even better than what I proposed above. Thank you!

internal/profiles/gpu/gpu.go
nojnhuh: I think these are logically coupled enough that it makes sense to keep them together. Tying the profile and config handling together in the profile I think makes more sense then somewhere higher-level like main.go. Implementing them together makes initialization a bit simpler: "`--device-profile` => `Profile` (also implementing `ConfigHandler`)" vs. "`--device-profile` => separate `Profile` and `ConfigHandler`". I don't think we'd ever want a user to specify the profile and config handler independently.

internal/profiles/profiles.go
nojnhuh: Overall, the goal of the `NoopConfigHandler` is to disallow any configuration to prevent users from supplying configuration that the driver doesn't know how to handle. I can see how the message can be unclear though whether it's read as "no (configuration allowed)" or "(no configuration) allowed" so I've reworded this to "configuration not allowed". Returning an error here is intended to help enforce that. Does that help clarify?

pohly (commented):
Can you update the top-level README.md to explain "profiles" for a) users of the example driver and b) for developers who want to fork the example driver to create their own DRA driver?

That second group needs to be told that "profiles" are probably something that they don't need to keep.

nojnhuh:
> Can you update the top-level README.md to explain "profiles" for a) users of the example driver and b) for developers who want to fork the example driver to create their own DRA driver?
> 
> That second group needs to be told that "profiles" are probably something that they don't need to keep.

Done.

pohly:
/hold

Until after a  patch release has been cut from the main branch.

nojnhuh:
Patch release is cut and I don't think we need to wait on anything else.

/hold cancel

PR #130, opened 2025-12-02
Update k8s.io dependencies to 1.35

nojnhuh:
This PR updates the example driver's dependencies to correspond with the Kubernetes 1.35 release.

PR #138, opened 2026-01-19
Update kubelet/pkg/apis/dra to v1

sunya-ch:
Update `k8s.io/kubelet/pkg/apis/dra` from `v1beta1` to `v1 `and support `ShareId` field.

Related to
- https://github.com/kubernetes-sigs/dra-example-driver/pull/130

### DIFF
--- a/cmd/dra-example-kubeletplugin/state.go
+++ b/cmd/dra-example-kubeletplugin/state.go
@@ -25,7 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/dynamic-resource-allocation/resourceslice"
-	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1beta1"
+	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1"
 	"k8s.io/kubernetes/pkg/kubelet/checkpointmanager"
 
 	"sigs.k8s.io/dra-example-driver/internal/profiles"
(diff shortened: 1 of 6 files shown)


pohly (commented):
/assign @nojnhuh 

Looks okay to me, but I'll defer to @nojnhuh for details.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -25,7 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/dynamic-resource-allocation/resourceslice"
-	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1beta1"
+	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1"
```
nojnhuh, line 39: I remember choosing not to do this sooner since it changes the format of checkpoints on disk: https://github.com/kubernetes-sigs/dra-example-driver/pull/123#discussion_r2417046248. I'm not really sure what the implications of that are, especially w.r.t. driver upgrades.

@bart0sh @klueska @pohly How should drivers generally be utilizing checkpoints? Is this the right way to integrate the `share_id` of a device into the checkpoint?

pohly: They should use a format that is entirely under their own control, ideally using versioning to enable conversion if the driver needs to change the format.

The kubelet checkpointing code is seriously flawed because it checksums the in-memory representation. See discussion in https://github.com/kubernetes-sigs/dra-example-driver/pull/90#discussion_r2053497278.

nojnhuh:
I'd still like to get #148 first. Then this won't affect anything that the driver writes to disk.

/remove-lifecycle stale

nojnhuh:
@sunya-ch This should be unblocked now that #148 merged. Could you please rebase?

sunya-ch:
@nojnhuh I'm sorry that I just noticed this. I have rebased the PR and as I was prevented from running the test with contextual logging check so I add a second comment to fix that.

nojnhuh (approved):
Thanks!

/lgtm
/approve

PR #148, opened 2026-02-05
Align checkpoint format with Kubernetes API conventions

nojnhuh:
This PR builds off of #90:

> This PR encapsulates the internal state of Checkpoint and prevents direct access to CheckpointV1 from DeviceState.

The second commit drops our dependency on k/k by refactoring away from the kubelet's [`CheckpointManager`](https://pkg.go.dev/k8s.io/kubernetes@v1.35.0/pkg/kubelet/checkpointmanager#CheckpointManager) to a pair of functions to read and write checkpoints from the local filesystem directly. Checksums have been dropped from the checkpoint format in favor of safely updating the file by writing a temporary copy and performing an atomic rename.

In addition, the checkpoint format has been changed to include an `apiVersion` and `kind` like other Kubernetes APIs. The new format also includes less data, relying on most container config for a ResourceClaim to be recomputed rather than restored directly from the checkpoint.


cmd/dra-example-kubeletplugin/checkpoint.go
```diff
@@ -1,14 +1,34 @@
...
 	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
 
-	"k8s.io/kubernetes/pkg/kubelet/checkpointmanager/checksum"
+	"sigs.k8s.io/dra-example-driver/internal/profiles"
 )
 
+type PreparedClaims map[string]profiles.PreparedDevices
```
pohly:
```Go
type PreparedDevice struct {
	drapbv1.Device
	ContainerEdits *cdiapi.ContainerEdits
	AdminAccess    bool
}
```

This does not look like a struct that is intended to be used in an on-disk file format. All it takes is for someone to rename one of these fields and encoding/decoding will break. The current unit test can catch that only for fields which are covered by a test case.

Should we use this opportunity (i.e. this PR) to introduced proper serialization of the internal state with a versioned on-disk format?

See https://github.com/kubernetes/kubernetes/pull/137086 for the machinery, used there for the on-disk metadata information file format. The problem is the same: one process writing a file, another reading it. That's an API, in this case between one version of the example driver and another.

nojnhuh: We might as well do that here. 

I'm trying to make the internal version contain a literal `k8s.io/kubelet/pkg/apis/dra/v1beta1.Device` and `tags.cncf.io/container-device-interface/pkg/cdi.ContainerEdits` since that's what the driver ultimately uses, but codegen is not liking those at all. The protobuf type has unexported fields which controller-gen tries to access in generated `DeepCopy` methods. The CDI types are structured kind of [weirdly](https://pkg.go.dev/tags.cncf.io/container-device-interface@v0.8.0/pkg/cdi#ContainerEdits) that's either weird to mirror in our API or needs a bunch of manual conversion to handle. To generate code that compiles it seems like we'll need to entirely duplicate all the types for every version and then manually convert from the internal API version to the types that the CDI and kubelet Go interfaces use. At that point I'd almost rather generate no code at all. I lost steam following this all the way through but pushed what I have so far in the newest commit here in case that's informative.

Am I way off base or do we need to accept that this just isn't going to be smooth?

pohly: Does all of that information really need to be persisted?

nojnhuh: I pared down the checkpoints all the way down to just the claim UID and rely on recomputing everything else from the ResourceClaim. I added some comments and refactored a bit to describe how real drivers can handle more complex scenarios.

nojnhuh:
I plan to squash before merging.

/hold

cmd/dra-example-kubeletplugin/checkpoint.go
```diff
@@ -1,53 +1,64 @@
...
-	cp.Checksum = checksum.New(out)
-	return json.Marshal(*cp)
-}
-
-func (cp *Checkpoint) UnmarshalCheckpoint(data []byte) error {
-	return json.Unmarshal(data, cp)
+	checkpoint := new(checkpointapi.Checkpoint)
+	_, _, err = decoder.Decode(data, ptr.To(checkpointapi.SchemeGroupVersion.WithKind("Checkpoint")), checkpoint)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal json from %s: %w", path, err)
```
pohly: Nit:

```suggestion
		return nil, fmt.Errorf("unmarshal JSON from %s: %w", path, err)
```

cmd/dra-example-kubeletplugin/checkpoint.go
```diff
@@ -1,53 +1,64 @@
...
-	out, err := json.Marshal(*cp)
-	if err != nil {
-		return err
+	if err := encoder.Encode(checkpoint, tmp); err != nil {
+		return fmt.Errorf("encode to temp file %s: %w", tmp.Name(), err)
+	}
+	if err := tmp.Sync(); err != nil {
+		return fmt.Errorf("sync temp file: %w", err)
+	}
+	if err := os.Rename(tmp.Name(), path); err != nil {
```
pohly, line 70: Won't `os.Rename` fail when the file already exists under the `path` name?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -95,29 +114,32 @@ func NewDeviceState(config *Config) (*DeviceState, error) {
...
 		configHandler:     configHandler,
+		checkpointPath:    filepath.Join(config.DriverPluginPath(), DriverPluginCheckpointFile),
+		checkpointDecoder: checkpointDecoder,
+		checkpointEncoder: checkpointEncoder,
 	}
 
-	checkpoints, err := state.checkpointManager.ListCheckpoints()
-	if err != nil {
-		return nil, fmt.Errorf("unable to list checkpoints: %v", err)
+	_, err = readCheckpoint(state.checkpointPath, state.checkpointDecoder)
```
pohly: Why read before writing and why is the result ignored?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -270,8 +305,30 @@ func (s *DeviceState) prepareDevices(claim *resourceapi.ResourceClaim) (profiles
...
+
+// removeClaimFromCheckpoint updates the checkpoint to remove all data
+// associated with the claim.
+func (*DeviceState) removeClaimFromCheckpoint(checkpoint *checkpointapi.Checkpoint, claimUID types.UID) {
+	checkpoint.PreparedClaims = slices.DeleteFunc(checkpoint.PreparedClaims, func(c checkpointapi.PreparedClaim) bool { return c.UID == claimUID })
+}
+
+// restoreCheckpoint returns the device definitions for devices already prepared
+// for the given claim. If the claim has not yet been prepared, it returns nil.
+func (s *DeviceState) restoreCheckpoint(checkpoint *checkpointapi.Checkpoint, claim *resourceapi.ResourceClaim) (PreparedDevices, error) {
```
pohly: Is `restoreClaimFromCheckpoint` perhaps a better name?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -286,6 +343,31 @@ func (s *DeviceState) checkAdminAccess(claim *resourceapi.ResourceClaim) bool {
...
+	if err := sb.AddToScheme(checkpointScheme); err != nil {
+		return nil, nil, fmt.Errorf("create checkpoint scheme: %w", err)
+	}
+	checkpointJSON := json.NewSerializerWithOptions(
+		json.DefaultMetaFactory,
+		checkpointScheme,
+		checkpointScheme,
+		json.SerializerOptions{
+			Pretty: true,
+			Strict: true,
```
pohly: "Strict" can bite us. This causes unknown fields during decoding to result in an error, right?

Suppose we need to extend the on-disk format with a field that older releases of the DRA driver don't know about. If we treat API versions like in the rest of Kubernetes, we'd add that field to all API versions and serialize it. Older clients then need to ignore unknown fields, which we can rely upon when adding fields that are okay to be ignored.

The alternative is to add new fields only in a new API version and then only serialize using that version. This immediately prevents downgrading, because older clients won't be able to read the file.

Which brings me to my next point: how do new API versions get rolled out? We need to document it. Probably something like this:
- In driver version x.y, support v1alpha1.
- In driver version x.y+1, support v1beta1, but still serialize as v1alpha1 (thus supporting downgrades).
- In driver version x.y+2, write as v1beta1 and then only support downgrades back to x.y+1.

internal/api/checkpoint/v1/register.go
```diff
@@ -0,0 +1,46 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name for checkpoints.
+const GroupName = "checkpoint.internal.example.com"
```
pohly, line 26: Shouldn't this be tied to the driver name instead of `example.com`?

Makefile
```diff
@@ -102,15 +102,24 @@ coverage: test
...
-			paths=$(CURDIR)/api/$(VENDOR)/resource/$${api}/ \
-			output:object:dir=$(CURDIR)/api/$(VENDOR)/resource/$${api}; \
+			paths=$${api}/ \
+			output:object:dir=$${api}; \
+	done
+
+generate-conversion:
+	for api in $(APIS); do \
+		rm -f $${api}/zz_generated.conversion.go; \
+		conversion-gen \
```
pohly, line 119: Where does this binary come from? Can we compile it from source if needed?

Makefile
```diff
@@ -128,7 +137,6 @@ teardown-e2e:
 .build-image: docker/Dockerfile.devel
 	if [ x"$(SKIP_IMAGE_BUILD)" = x"" ]; then \
 		$(CONTAINER_TOOL) build \
-			--progress=plain \
```
pohly, line 131: Seems unrelated?

pohly (commented):
I think this goes in the right direction, but needs documentation of the concept.

cmd/dra-example-kubeletplugin/checkpoint.go
nojnhuh: Updated.

cmd/dra-example-kubeletplugin/checkpoint.go
nojnhuh: No, per the docs: https://pkg.go.dev/os#Rename
> If newpath already exists and is not a directory, Rename replaces it.

I haven't noticed any crash-looping either in my testing and I can still see the checkpoint change on disk.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: This is emulating the previous behavior where an empty checkpoint is written to disk unless one already exists. I assume that was to make sure that the first `NodePrepareResources` can assume that a checkpoint already exists. I don't think that assumption really buys us anything though. I updated this to only write a checkpoint on the first `NodePrepareResources`.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: I think so, updated.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: The docs suggest it means something different: https://pkg.go.dev/k8s.io/apimachinery@v0.35.4/pkg/runtime/serializer/json#SerializerOptions.Strict

> Strict: configures the Serializer to return strictDecodingError's when duplicate fields are present decoding JSON or YAML. Note that enabling this option is not as performant as the non-strict variant, and should not be used in fast paths.

I tried manually editing the checkpoint on disk to add an unknown field and the next `NodePrepareResources succeeds`, removing the unknown field when it serializes the checkpoint again. I think that's the expected behavior?

> * In driver version x.y+1, support v1beta1, but still serialize as v1alpha1 (thus supporting downgrades).

Does it make sense to even define a v1beta1 if the driver only writes v1alpha1? If v1beta1 is added to support some feature, then the driver should start writing v1beta1 right away. Maybe the driver can be more intelligent and continue writing v1alpha1 if it can tell it doesn't need any of the features in v1beta1. I think we can worry about that though once we need a new API version. Added some go docs to describe that process.

internal/api/checkpoint/v1/register.go
nojnhuh: The driver is named `<profile>.example.com`. Should it have a k8s.io name instead?

Makefile
nojnhuh: The scripts run this in a docker container based on an image where this is already installed via `go install`. This could be improved but `controller-gen` is already set up that same way, so I'd rather refactor that in a separate PR.

Makefile
nojnhuh: It is. I'd like to clean a lot of this up at some point but we can do that later.

cmd/dra-example-kubeletplugin/state.go
pohly: > I don't think that assumption really buys us anything though. I updated this to only write a checkpoint on the first NodePrepareResources.

I agree. "No checkpoint file" should be treated the same as "empty checkpoint file", so there's no point in explicitly writing an empty one.

cmd/dra-example-kubeletplugin/state.go
pohly: Interesting. In https://pkg.go.dev/sigs.k8s.io/yaml#UnmarshalStrict, "strict" includes "presence of fields in the serialized data unknown to the struct will yield an error".

> I tried manually editing the checkpoint on disk to add an unknown field and the next NodePrepareResources succeeds, removing the unknown field when it serializes the checkpoint again. I think that's the expected behavior?

Yes, thanks for checking.

internal/api/checkpoint/v1/register.go
pohly: Yes, I think `dra-example-driver.sigs.k8s.io` would be a better suffix. But that's then a separate PR, using `example.com` here for consistency with the current code is fine.

@klueska, @johnbelamaric: thoughts?

cmd/dra-example-kubeletplugin/state.go
pohly: > Does it make sense to even define a v1beta1 if the driver only writes v1alpha1?

I think we can do the opposite and go straight to v1. Basically the driver then commits to only use that version and maybe extend it with new fields, with the caveat that older releases will ignore and drop those. Versioning as "v1" is just future-proofing the storage format because if structural changes are ever going to be needed, they can be done.

klueska: Note this code was pulled from a (very) old version of the DRA driver for GPUs, and we have since relaxed the use of `strict` for (most) code paths. The example driver should probably do the same: https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/commit/fb4822f7614feb37222e46aeea3a944e22a5c7ae

nojnhuh: Thanks, updated to match. Strict for config, non-strict for checkpoints.

internal/api/checkpoint/v1/register.go
nojnhuh: There might be a bit more risk that new drivers will misuse the k8s.io domain, but a comment should be enough to steer them in the right direction.

cmd/dra-example-kubeletplugin/state.go
nojnhuh: > I think we can do the opposite and go straight to v1. Basically the driver then commits to only use that version and maybe extend it with new fields, with the caveat that older releases will ignore and drop those. Versioning as "v1" is just future-proofing the storage format because if structural changes are ever going to be needed, they can be done.

Makes sense, updated to v1.

internal/api/checkpoint/v1/register.go
nojnhuh: Opened #174 to continue this discussion.

pohly:
@nojnhuh: this needs a rebase.

cmd/dra-example-kubeletplugin/state.go
pohly: > we have since relaxed the use of strict for (most) code paths

Why? As pointed out earlier, in this context unknown fields would be okay, strict only checks for duplicate fields (https://pkg.go.dev/k8s.io/apimachinery@v0.35.4/pkg/runtime/serializer/json#SerializerOptions.Strict)?

I don't mind relaxing that check because it doesn't really buy us much, I was just wondering.

nojnhuh:
> @nojnhuh: this needs a rebase.

I pushed a merge in case that's easier to review. I need to squash all the commits here down to one anyway.

/hold for squash

cmd/dra-example-kubeletplugin/checkpoint.go
```diff
@@ -1,53 +1,74 @@
...
+)
 
-func newCheckpoint() *Checkpoint {
-	pc := &Checkpoint{
-		Checksum: 0,
-		V1: &CheckpointV1{
-			PreparedClaims: make(PreparedClaims),
-		},
+// readCheckpoint returns the Checkpoint at the given path in the format
+// expected by the given decoder. If the path doesn't exist, returns an empty
```
pohly, line 33:
```suggestion
// expected by the given decoder. If the path doesn't exist, returns a nil
```

cmd/dra-example-kubeletplugin/checkpoint_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+
+	checkpointapi "sigs.k8s.io/dra-example-driver/internal/api/checkpoint"
+)
+
+func TestReadWriteCheckpointRoundtrip(t *testing.T) {
```
pohly, line 30: This doesn't cover overwriting an existing checkpoint file, does it?

I think this test works better when walking through the normal steps in a single test:
- create initial, empty checkpoint
- read to verify
- overwrite with some content
- read to verify again

cmd/dra-example-kubeletplugin/checkpoint.go
nojnhuh: Returning a non-nil `*Checkpoint` here is intentional since that eliminates the need for nil checks elsewhere. Should I clarify that in the comment?

internal/api/checkpoint/v1/doc.go
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// +k8s:deepcopy-gen=package
+// +k8s:conversion-gen=sigs.k8s.io/dra-example-driver/internal/api/checkpoint
+
+// Package v1 contains the v1 serialization format for checkpoints.
+// These types include JSON tags and are used for reading/writing files on disk.
```
pohly, line 21:
```suggestion
// These types include JSON tags and are used for reading/writing files on disk.
//
// In practice, this is likely to be the only supported version because introducing
// a new version would be difficult. See the checkpoint package documentation
// for more information on versioning.
```

pohly: I was expecting information on versioning here and only found the other docs later.

internal/api/checkpoint/doc.go
```diff
@@ -0,0 +1,45 @@
...
+// [Checkpoint] type must also be updated such that it can be converted to and
+// from the new API version.
+//
+// The driver should be able to read API versions written by older versions of
+// the driver to facilitate upgrades of the driver. The driver writes an API
+// version which fulfills the needs of its allocated devices. If the set of
+// allocated devices requires the latest checkpoint API version, then that
+// version must be written and the driver cannot be downgraded until after the
+// devices have been unprepared and the driver can write the checkpoint in an
+// older API version.
```
pohly: According to this documentation, a new version is introduced to store information that cannot be represented in the older versions. That's not how the Kubernetes APIs work: they always allow storing all information in all versions, but maybe using a different structure. So a newer driver could always write using the oldest version to support downgrades.

I think we should use the approach used by Kubernetes. The alternative is to introduce a new version each time a new field gets added, which breaks downgrades each time whether that's necessary or not; it's not necessary if the new field can be safely ignored by an older release.

cmd/dra-example-kubeletplugin/checkpoint_test.go
nojnhuh: Updated with a single sequence of steps.

internal/api/checkpoint/v1/doc.go
nojnhuh: Added your suggestion. Were you looking for more here?

internal/api/checkpoint/doc.go
nojnhuh: I updated this section to try to clarify how this should work.

internal/api/checkpoint/v1/doc.go
pohly: No, that's all.

pohly (approved):
/lgtm
/approve
/hold

For cleanup of the commits.

nojnhuh:
Squashed

/hold cancel

nojnhuh:
/retitle Align checkpoint format with Kubernetes API conventions

PR #150, opened 2026-02-05
Add DRAPartitionableDevices support

harche:
## Summary

Adds support for the [DRAPartitionableDevices](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4815-dra-partitionable-devices) feature (Beta in Kubernetes 1.36). Each physical GPU is represented as a `CounterSet` with shared memory and compute counters. Multiple partition devices consume from these counters, enabling flexible GPU partitioning. A full-GPU device is also available that consumes all counters.

## How to test

```bash
# Build the driver image
./demo/build-driver.sh

# Create a kind cluster (uses k8s v1.36.0)
./demo/clusters/kind/create-cluster.sh

# Install with partitionable devices enabled
helm upgrade -i \
  --create-namespace \
  --set kubeletPlugin.partitionableDevices=true \
  --set kubeletPlugin.partitionsPerGPU=4 \
  --set kubeletPlugin.numDevices=2 \
  dra-example-driver \
  deployments/helm/dra-example-driver

# Verify ResourceSlices (should see two — one for counters, one for devices)
kubectl get resourceslices -o wide

# Run the demo
kubectl apply -f demo/partitionable-devices.yaml

# Check allocated partitions
kubectl logs -n partitionable-devices pod0 -c ctr0 | grep GPU_DEVICE
```

## Test Report

<details>
<summary>Click to expand test report (kind cluster, k8s v1.36.0)</summary>

**Date:** 2026-05-05
**Kubernetes Version:** v1.36.0
**Test Environment:** kind cluster (`dra-example-driver-cluster`)

### Test Environment Setup

| Component | Details |
|---|---|
| Kind node image | `kindest/node:v1.36.0` (built from source) |
| Cluster nodes | 1 control-plane + 1 worker |
| Driver image | `registry.k8s.io/dra-example-driver/dra-example-driver:v0.2.1` (built from PR branch) |
| Helm values | `partitionableDevices=true`, `partitionsPerGPU=4`, `numDevices=2` |
| Feature gate | `DRAPartitionableDevices` (Beta, default-enabled in v1.36.0) |

### 1. API Type Validation Against k8s v1.36.0 Source

**Result: PASS**

Cross-referenced every API type used in the PR against the Kubernetes v1.36.0 source (`staging/src/k8s.io/api/resource/v1/types.go`):

| Type/Field | Expected | Actual in v1.36.0 | Match |
|---|---|---|---|
| `CounterSet.Name` | `string` | `string` | Yes |
| `CounterSet.Counters` | `map[string]Counter` | `map[string]Counter` | Yes |
| `Counter.Value` | `resource.Quantity` | `resource.Quantity` | Yes |
| `DeviceCounterConsumption.CounterSet` | `string` | `string` | Yes |
| `DeviceCounterConsumption.Counters` | `map[string]Counter` | `map[string]Counter` | Yes |
| `Device.ConsumesCounters` | `[]DeviceCounterConsumption` | `[]DeviceCounterConsumption` | Yes |
| `resourceslice.Slice.SharedCounters` | `[]resourceapi.CounterSet` | `[]resourceapi.CounterSet` | Yes |

### 2. Unit Tests

**Result: PASS**

```
ok   sigs.k8s.io/dra-example-driver/api/example.com/resource/gpu/v1alpha1
ok   sigs.k8s.io/dra-example-driver/cmd/dra-example-kubeletplugin
ok   sigs.k8s.io/dra-example-driver/cmd/dra-example-webhook
ok   sigs.k8s.io/dra-example-driver/internal/profiles/gpu
```

### 3. ResourceSlice Creation

**Result: PASS**

Two separate ResourceSlices created successfully (one for counters, one for devices):

```
$ kubectl get resourceslices -o wide
NAME                                                            DRIVER            POOL
00000-gpu.example.com-dra-example-driver-cluster-worker-l4vwd   gpu.example.com   dra-example-driver-cluster-worker
00001-gpu.example.com-dra-example-driver-cluster-worker-hv7m9   gpu.example.com   dra-example-driver-cluster-worker
```

### 4. SharedCounters Slice Validation

**Result: PASS**

The counter slice correctly contains:

- `gpu-0-counters`: memory=80Gi, compute=100
- `gpu-1-counters`: memory=80Gi, compute=100
- No `devices` field (as required by the API)

### 5. Devices Slice Validation

**Result: PASS**

The device slice correctly contains 10 devices:

| Device | Type | ConsumesCounters | Memory | Compute |
|---|---|---|---|---|
| `gpu-0-partition-0` | partition | `gpu-0-counters` | 20Gi | 25 |
| `gpu-0-partition-1` | partition | `gpu-0-counters` | 20Gi | 25 |
| `gpu-0-partition-2` | partition | `gpu-0-counters` | 20Gi | 25 |
| `gpu-0-partition-3` | partition | `gpu-0-counters` | 20Gi | 25 |
| `gpu-0-full` | full GPU | `gpu-0-counters` | 80Gi | 100 |
| `gpu-1-partition-0` | partition | `gpu-1-counters` | 20Gi | 25 |
| `gpu-1-partition-1` | partition | `gpu-1-counters` | 20Gi | 25 |
| `gpu-1-partition-2` | partition | `gpu-1-counters` | 20Gi | 25 |
| `gpu-1-partition-3` | partition | `gpu-1-counters` | 20Gi | 25 |
| `gpu-1-full` | full GPU | `gpu-1-counters` | 80Gi | 100 |

### 6. Scheduler Allocation with Shared Counters

**Result: PASS**

Created a `ResourceClaim` requesting 2 devices. The scheduler successfully allocated two partition devices from the same physical GPU:

```yaml
status:
  allocation:
    devices:
      results:
      - device: gpu-0-partition-0
        pool: dra-example-driver-cluster-worker
      - device: gpu-0-partition-1
        pool: dra-example-driver-cluster-worker
```

### 7. Container Environment Variable Injection

**Result: PASS**

The test pod received the expected GPU environment variables:

```
GPU_DEVICE_0-partition-0=gpu-0-partition-0
GPU_DEVICE_0-partition-1=gpu-0-partition-1
GPU_DEVICE_0-partition-0_SHARING_STRATEGY=TimeSlicing
GPU_DEVICE_0-partition-1_SHARING_STRATEGY=TimeSlicing
GPU_DEVICE_0-partition-0_TIMESLICE_INTERVAL=Default
GPU_DEVICE_0-partition-1_TIMESLICE_INTERVAL=Default
```

</details>

### DIFF
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -53,6 +53,7 @@ type Flags struct {
 	profile                       string
 	driverName                    string
 	podUID                        string
+	gpuPartitions                 int
 }
 
 type Config struct {
@@ -65,7 +66,7 @@ type Config struct {
 
 var validProfiles = map[string]func(flags Flags) profiles.Profile{
 	gpu.ProfileName: func(flags Flags) profiles.Profile {
-		return gpu.NewProfile(flags.nodeName, flags.numDevices)
+		return gpu.NewProfile(flags.nodeName, flags.numDevices, flags.gpuPartitions)
 	},
 }
 
@@ -153,6 +154,13 @@ func newApp() *cli.App {
 			Usage:       "UID of the pod (used for seamless upgrades to create unique socket names).",
 			Destination: &flags.podUID,
 			EnvVars:     []string{"POD_UID"},
+		},
+		&cli.IntFlag{
+			Name:        "gpu-partitions",
+			Usage:       "Number of partitions per GPU. When set to a value greater than 0, GPUs are exposed with shared counters allowing flexible partitioning (DRAPartitionableDevices feature).",
+			Value:       0,
+			Destination: &flags.gpuPartitions,
+			EnvVars:     []string{"GPU_PARTITIONS"},
 		},
 	}
 	cliFlags = append(cliFlags, flags.kubeClientConfig.Flags()...)
--- a/demo/partitionable-devices.yaml
+++ b/demo/partitionable-devices.yaml
@@ -1,0 +1,71 @@
+# Example: DRA Partitionable Devices
+#
+# This demo shows the DRAPartitionableDevices feature, which allows GPUs to be
+# exposed with shared counters enabling flexible partitioning. Each physical GPU
+# has a counter set (memory and compute), and multiple partition devices consume
+# from those counters. A full-GPU device is also available that consumes all
+# counters.
+#
+# Helm install (required before running this demo):
+#   helm upgrade -i \
+#     --create-namespace \
+#     --namespace dra-example-driver \
+#     --set kubeletPlugin.gpuPartitions=4 \
+#     --set kubeletPlugin.numDevices=2 \
+#     dra-example-driver \
+#     deployments/helm/dra-example-driver
+#
+# Verify ResourceSlices are created (should see two slices per node — one for
+# shared counters and one for partition devices):
+#   kubectl get resourceslices -o wide
+#
+# Expected: pod0 gets two GPU partitions. The scheduler uses shared counters
+# to track that partitions share GPU resources. Check with:
+#   kubectl logs -n partitionable-devices pod0 -c ctr0 | grep GPU_DEVICE
+#
+# Driver requirements:
+#   Profile: gpu
+#   Helm values: kubeletPlugin.gpuPartitions=4
+#
+# Cluster requirements:
+#   Kubernetes 1.35+ with feature gate DRAPartitionableDevices enabled
+#   (Beta and enabled by default in 1.36+)
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: partitionable-devices
+---
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaimTemplate
+metadata:
+  namespace: partitionable-devices
+  name: gpu-partitions
+spec:
+  spec:
+    devices:
+      requests:
+      - name: gpu-partition
+        exactly:
+          deviceClassName: gpu.example.com
+          count: 2
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: partitionable-devices
+  name: pod0
+spec:
+  restartPolicy: Never
+  containers:
+  - name: ctr0
+    image: ubuntu:22.04
+    command: ["bash", "-c"]
+    args: ["export; trap 'exit 0' TERM; sleep 9999 & wait"]
+    resources:
+      claims:
+      - name: gpu-partitions
+  resourceClaims:
+  - name: gpu-partitions
+    resourceClaimTemplateName: gpu-partitions
--- a/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
+++ b/deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
@@ -84,6 +84,8 @@ spec:
         - name: HEALTHCHECK_PORT
           value: {{ .Values.kubeletPlugin.containers.plugin.healthcheckPort | quote }}
         {{- end }}
+        - name: GPU_PARTITIONS
+          value: {{ .Values.kubeletPlugin.gpuPartitions | quote }}
         - name: POD_UID
           valueFrom:
             fieldRef:
--- a/internal/profiles/gpu/gpu.go
+++ b/internal/profiles/gpu/gpu.go
@@ -18,7 +18,9 @@ package gpu
 
 import (
 	"fmt"
+	"maps"
 	"math/rand"
+	"strings"
 
 	"github.com/google/uuid"
 	resourceapi "k8s.io/api/resource/v1"
@@ -36,14 +38,16 @@ import (
 const ProfileName = "gpu"
 
 type Profile struct {
-	nodeName string
-	numGPUs  int
+	nodeName         string
+	numGPUs          int
+	partitionsPerGPU int
 }
 
-func NewProfile(nodeName string, numGPUs int) Profile {
+func NewProfile(nodeName string, numGPUs int, partitionsPerGPU int) Profile {
 	return Profile{
-		nodeName: nodeName,
-		numGPUs:  numGPUs,
+		nodeName:         nodeName,
+		numGPUs:          numGPUs,
+		partitionsPerGPU: partitionsPerGPU,
 	}
 }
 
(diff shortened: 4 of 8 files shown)


harche:
> @harche could you update this PR? [kubernetes/kubernetes#137350](https://github.com/kubernetes/kubernetes/pull/137350) has already merged.

Done, thanks.

bitoku:
@nojnhuh should it be implemented as a different profile according to the PoC in https://github.com/kubernetes-sigs/dra-example-driver/pull/129 ?

nojnhuh:
> @nojnhuh should it be implemented as a different profile according to the PoC in #129 ?

Not necessarily. Using this to configure the existing GPU profile seems reasonable.

nojnhuh:
@harche Could you please add an e2e test that runs the new example?

harche:
> @harche Could you please add an e2e test that runs the new example?

Thanks, added.

harche:
> > @harche Could you please add an e2e test that runs the new example?
> 
> Thanks, added.

/hold something isn't right with the tests. Debugging.

harche:
With partitionable devices enabled, device names like `gpu-0-partition-0` produce env vars with hyphens (`GPU_DEVICE_0-partition-0`), which are silently dropped by the container runtime. Fixed by converting device IDs to uppercase with underscores (`GPU_DEVICE_0_PARTITION_0`) in `applyGpuConfig`, and updated e2e test regexes to match on the value pattern (`gpu-...`) instead of the key pattern. Also simplified the admin-access manifest to request 1 GPU instead of `allocationMode: All`, which exceeds the [32-device per-claim limit](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/resource/v1/types.go#L1760) when partitions multiply the device count.

harche:
Local e2e test results with partitionable devices enabled (`numDevices=8, partitionsPerGPU=4`) on KinD v0.31.0 / Kubernetes v1.36.0:

```
Ran 13 of 13 Specs in 12.398 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 0 Skipped
```

nojnhuh:
> Also simplified the admin-access manifest to request 1 GPU instead of `allocationMode: All`, which exceeds the [32-device per-claim limit](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/resource/v1/types.go#L1760) when partitions multiply the device count.

Use cases for admin access usually require `allocationMode: All`, so I'd like to keep that in that example. Can we try to decouple the tests?

@willie-yao Have you already started the work to install a separate driver instance for each test?

harche:
> Can we try to decouple the tests?

ok, let me think through on this one.

harche:
Reverted `admin-access.yaml` to keep `allocationMode: All`. Decoupled the partitionable-devices test by marking it `Serial` and having it `helm upgrade` the driver with `partitionableDevices=true` before running, then `DeferCleanup` restores the original config. Existing tests run unmodified with standard devices.

harche:
> Reverted `admin-access.yaml` to keep `allocationMode: All`. Decoupled the partitionable-devices test by marking it `Serial` and having it `helm upgrade` the driver with `partitionableDevices=true` before running, then `DeferCleanup` restores the original config. Existing tests run unmodified with standard devices.

@nojnhuh WDYT?

willie-yao:
> Have you already started the work to install a separate driver instance for each test?

@nojnhuh I have not started yet on installing a separate driver instance. Should I get started on that or would it be a part of this PR?

nojnhuh:
@willie-yao That should be a separate PR.

nojnhuh:
We can stick with doing one `helm upgrade` for now to unblock the rest of this feature.

willie-yao:
@nojnhuh Sounds good, I'll work on it

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -154,6 +159,20 @@ func newApp() *cli.App {
...
 		},
+		&cli.BoolFlag{
+			Name:        "partitionable-devices",
+			Usage:       "Enable partitionable devices support (DRAPartitionableDevices feature). When enabled, GPUs are exposed with shared counters allowing flexible partitioning.",
+			Value:       false,
+			Destination: &flags.partitionableDevices,
+			EnvVars:     []string{"PARTITIONABLE_DEVICES"},
+		},
+		&cli.IntFlag{
+			Name:        "partitions-per-gpu",
```
nojnhuh: As we add flags like this that are relevant to one profile, I'd like to kind of namespace those to the profile, i.e. `--profile-*`. Could we call this `--gpu-partitions`?

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -154,6 +159,20 @@ func newApp() *cli.App {
 			Destination: &flags.podUID,
 			EnvVars:     []string{"POD_UID"},
 		},
+		&cli.BoolFlag{
+			Name:        "partitionable-devices",
```
nojnhuh: Do we need a separate flag for this, or can we say that if a user asks for at least one partition, then partitionable devices is enabled? If the default number of partitions is 0 it seems like that should work.

demo/partitionable-devices.yaml
```diff
@@ -0,0 +1,73 @@
...
+# The scheduler uses shared counters to track that partitions share GPU
+# resources. Check with:
+#   kubectl logs -n partitionable-devices pod0 -c ctr0 | grep GPU_DEVICE
+#
+# Driver requirements:
+#   Profile: gpu
+#   Helm values: kubeletPlugin.partitionableDevices=true
+#
+# Cluster requirements:
+#   Kubernetes 1.36+
```
nojnhuh: Does this work with older versions when the feature gate is explicitly enabled? Or does this example depend on significant changes to the feature in 1.36?

demo/partitionable-devices.yaml
```diff
@@ -0,0 +1,73 @@
...
+#     --set kubeletPlugin.partitionsPerGPU=4 \
+#     --set kubeletPlugin.numDevices=2 \
+#     dra-example-driver \
+#     deployments/helm/dra-example-driver
+#
+# Verify ResourceSlices are created (should see two slices per node — one for
+# shared counters and one for partition devices):
+#   kubectl get resourceslices -o wide
+#
+# Expected: pod0 gets two GPU partitions from the same physical GPU.
```
nojnhuh: Nothing in this example is preventing two partitions being allocated from separate GPUs. Instead of updating the example to make that true, let's keep the example simple and reword this statement.

internal/profiles/gpu/gpu.go
```diff
@@ -146,6 +316,10 @@ func (p Profile) ApplyConfig(config runtime.Object, results []*resourceapi.Devic
 // define a set of environment variables to be injected into the containers
 // that include a given device. A real driver would likely need to do some sort
 // of hardware configuration as well, based on the config passed in.
```
nojnhuh: Let's make sure this comment stays right before `applyGpuConfig`

cmd/dra-example-kubeletplugin/main.go
harche: Good call. Removed the separate `--partitionable-devices` bool flag and consolidated into a single `--gpu-partitions` int flag (default 0). When `> 0`, partitionable devices are enabled with that many partitions per GPU.

cmd/dra-example-kubeletplugin/main.go
harche: Done. Renamed to `--gpu-partitions` (env var `GPU_PARTITIONS`), and updated Helm values to `kubeletPlugin.gpuPartitions`.

demo/partitionable-devices.yaml
harche: Agreed — `count: 2` with `exactly` doesn't constrain allocation to the same counter set. Reworded to just "Expected: pod0 gets two GPU partitions."

demo/partitionable-devices.yaml
harche: The `DRAPartitionableDevices` feature gate was introduced as Alpha in 1.33 ([#130764](https://github.com/kubernetes/kubernetes/pull/130764)), with breaking changes to the `CounterSet`/`ConsumesCounters` types in 1.35 ([#134189](https://github.com/kubernetes/kubernetes/pull/134189)), and promoted to Beta in 1.36 ([#137350](https://github.com/kubernetes/kubernetes/pull/137350)). This example should work on 1.35+ with the feature gate explicitly enabled. Updated the comment to reflect that.

internal/profiles/gpu/gpu.go
harche: Fixed — moved `envVarSafeID` above the comment block so the comment stays directly before `applyGpuConfig`.

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -65,6 +66,9 @@ type Config struct {
 
 var validProfiles = map[string]func(flags Flags) profiles.Profile{
 	gpu.ProfileName: func(flags Flags) profiles.Profile {
+		if flags.gpuPartitions > 0 {
+			return gpu.NewPartitionableProfile(flags.nodeName, flags.numDevices, flags.gpuPartitions)
+		}
 		return gpu.NewProfile(flags.nodeName, flags.numDevices)
```
nojnhuh: Can we add the partitions parameter to the existing `gpu.NewProfile` function?

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
```diff
@@ -84,6 +84,10 @@ spec:
         - name: HEALTHCHECK_PORT
           value: {{ .Values.kubeletPlugin.containers.plugin.healthcheckPort | quote }}
         {{- end }}
+        {{- if gt (int .Values.kubeletPlugin.gpuPartitions) 0 }}
+        - name: GPU_PARTITIONS
+          value: {{ .Values.kubeletPlugin.gpuPartitions | quote }}
+        {{- end }}
```
nojnhuh: This condition doesn't seem strictly necessary:

```suggestion
        - name: GPU_PARTITIONS
          value: {{ .Values.kubeletPlugin.gpuPartitions | quote }}
```

internal/profiles/gpu/gpu.go
```diff
@@ -93,6 +112,151 @@ func (p Profile) EnumerateDevices() (resourceslice.DriverResources, error) {
 	return resources, nil
 }
 
+// enumeratePartitionableDevices creates devices with shared counters for partitionable devices.
+// Each physical GPU is represented as a CounterSet with memory and compute counters.
+// Multiple partition devices are created that consume from these counters.
+func (p Profile) enumeratePartitionableDevices(uuids []string) (resourceslice.DriverResources, error) {
```
nojnhuh: Can we integrate this with the existing `EnumerateDevices` method instead of duplicating the logic for the full devices in both places?

internal/profiles/gpu/gpu.go
```diff
@@ -160,12 +328,13 @@ func applyGpuConfig(config *configapi.GpuConfig, results []*resourceapi.DeviceRe
 	}
 
 	for _, result := range results {
+		envID := envVarSafeID(result.Device[4:])
```
nojnhuh, line 260: Can you add a comment explaining the assumptions about the device names here?

cmd/dra-example-kubeletplugin/main.go
harche: Done. Merged `NewPartitionableProfile` into `NewProfile`, which now takes `partitionsPerGPU` as a third parameter. `0` means standard mode.

deployments/helm/dra-example-driver/templates/kubeletplugin.yaml
harche: Agreed. Removed the conditional, `GPU_PARTITIONS` is now always passed. The driver handles `0` as no partitioning.

internal/profiles/gpu/gpu.go
harche: Done. Merged `enumeratePartitionableDevices` into `EnumerateDevices`. Base attributes are built once per GPU, then reused via `maps.Clone` for partition and full devices. The standard path uses them directly.

internal/profiles/gpu/gpu.go
harche: Added a comment: `// Device names are prefixed with "gpu-" (e.g. "gpu-0", "gpu-0-partition-1", "gpu-0-full").`

nojnhuh (approved):
/lgtm
/approve

Thanks!

harche:
> /lgtm /approve
> 
> Thanks!

Thanks @nojnhuh

PR #164, opened 2026-04-20
Enable DRA device metadata (KEP-5304)

carlory:
## Summary

Add support for [KEP-5304](https://github.com/kubernetes/enhancements/issues/5304) in-container device metadata by passing device attributes from the allocatable device pool through to prepared devices.

- New `--enable-device-metadata` CLI flag and `ENABLE_DEVICE_METADATA` env var
- New `kubeletPlugin.enableDeviceMetadata` Helm chart value
- Device attributes from the driver's allocatable pool are propagated as metadata on prepared devices
- Bumps Go to 1.26.0 and all k8s.io dependencies to v0.36.0-rc.1

/cc @alaypatel07 
I add you as co-author. If you want to submit a PR, feel free to close this PR.


go.mod
```diff
@@ -1,22 +1,22 @@
...
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.11.1
 	github.com/urfave/cli/v2 v2.27.7
 	google.golang.org/grpc v1.80.0
-	k8s.io/api v0.35.4
-	k8s.io/apimachinery v0.35.4
-	k8s.io/client-go v0.35.4
-	k8s.io/component-base v0.35.4
-	k8s.io/dynamic-resource-allocation v0.35.4
+	k8s.io/api v0.36.0-rc.1
```
nojnhuh: I have a draft of the changes to bump to 1.36 here but haven't opened a PR yet: https://github.com/kubernetes-sigs/dra-example-driver/compare/main...nojnhuh:dra-example-driver:k8s-1.36

I'm ok if we want to do that in this PR, but can we at least make those changes a separate commit here? There are also some other places that need to be updated here that you can check against my branch.

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -48,6 +48,7 @@ kubeletPlugin:
   affinity: {}
   kubeletRegistrarDirectoryPath: /var/lib/kubelet/plugins_registry
   kubeletPluginsDirectoryPath: /var/lib/kubelet/plugins
+  enableDeviceMetadata: false
```
nojnhuh, line 74: Can we update the e2e tests to enable and exercise this flag?

#153 refactors all of the e2e logic but I think that will merge soon, so you may want to hold off on that until that merges.

go.mod
carlory: v1.36 will be released in two days, once it is released, I will update this PR.

demo/scripts/common.sh
```diff
@@ -36,7 +36,7 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 
 # The kubernetes tag to build the kind cluster from
 # From ${KIND_K8S_REPO}/tags
-: ${KIND_K8S_TAG:="v1.35.0"}
+: ${KIND_K8S_TAG:="v1.36.0"}
```
carlory, line 39: v1.36.0 is unavailable.
```
~ » skopeo list-tags docker://docker.io/kindest/node | jq '.Tags[-1]'                                          kiki@zhanxis-MacBook-Pro
"v1.35.1"
```

nojnhuh: We could set `BUILD_KIND_IMAGE` based on whether `$KIND_IMAGE` exists instead of a static true/false. https://github.com/kubernetes-sigs/dra-example-driver/blob/a538bc889a19650b20b8e84d91a5b53dd9f0e443/demo/scripts/common.sh#L41-L44

I've noticed resource constraint issues with building the kind image from source in CI. Then we could refactor `demo/scripts/build-kind-image.sh` to allow building the image from release artifacts. We would need a way to pass `--type release <version>` here instead of a file path to the cloned repo: https://github.com/kubernetes-sigs/dra-example-driver/blob/a538bc889a19650b20b8e84d91a5b53dd9f0e443/demo/scripts/build-kind-image.sh#L55

A generic variable like `KIND_BUILD_NODE_IMAGE_ARGS` is probably good enough to handle a variety of use cases.

nojnhuh: I'm trying this out in #188.

PR #167, opened 2026-04-22
feat(build): support multi-arch image push (amd64/arm64)

### LINKED ISSUE #161: ARM64 Image
Requesting that in next release, can we please make sure we publish a multi-arch image.

Will really appreciate that.

vishalanarase:
### Summary
- Add Docker Buildx-based multi-arch image publishing for the container image build flow.
- Keep local developer workflow simple by adding a `build` target that loads a single-arch image into Docker.
- Document container image make recipes in `README.md` for build, local load, and multi-arch push use cases.

Fixes #161 

### What changed
- `deployments/container/Makefile`
  - Add `PLATFORMS` (default `linux/amd64,linux/arm64`)
  - Add `BUILDX_BUILDER` and `ensure-buildx-builder` helper target
  - Add `build` target (`buildx --load`) for local image testing
  - Add `push` target (`buildx --push`) for multi-arch publishing
- `README.md`
  - Add a new "Container image make recipes" section with usage examples for:
    - `build`
    - `push`

### Why
Issue request asks for ARM64 image availability. Multi-arch publish via Buildx ensures released tags include both `linux/amd64` and `linux/arm64`, while still preserving straightforward local build/test flows.

### Validation
- Ran `build` and verified command wiring and flags.
- Ran `push` and verified Buildx builder/platform settings.
- Verified multi-arch manifest output after push includes:
  - `linux/amd64`
  - `linux/arm64`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,5 +182,4 @@ push-release-artifacts:
 		HELM=$(HELM) \
 		demo/scripts/push-driver-chart.sh
 	export DRIVER_IMAGE_TAG="${IMAGE_GIT_TAG}"; \
-	demo/scripts/build-driver-image.sh && \
 	demo/scripts/push-driver-image.sh
--- a/README.md
+++ b/README.md
@@ -50,6 +50,84 @@ From here we will build the image for the example resource driver:
 ./demo/build-driver.sh
 ```
 
+### Container image make recipes
+
+The image build logic lives in `deployments/container/Makefile`.
+If variables are not provided, defaults are:
+
+- `IMAGE_NAME=registry.example.com/dra-example-driver`
+- `VERSION=latest`
+- `PLATFORMS=<current host platform>` (for example `linux/amd64`, `linux/arm64`, or `linux/ppc64le`) when `PLATFORMS` is **unset**
+- `CONTAINER_TOOL=docker`
+
+For demo scripts, `PLATFORMS` is the canonical variable and `DRIVER_IMAGE_PLATFORMS`
+is only a backward compatible fallback. Setting both is treated as an error.
+
+`cloudbuild.yaml` and `demo/scripts/push-driver-image.sh` may provide different
+`PLATFORMS` defaults depending on the workflow:
+
+1. **If `PLATFORMS` is set (e.g. by `cloudbuild.yaml`)**, the Makefile uses it as-is.
+2. **If `PLATFORMS` is unset**, `demo/scripts/push-driver-image.sh` fills a fallback
+   (currently `linux/amd64,linux/arm64,linux/ppc64le`) and `deployments/container/Makefile`
+   falls back to the host platform.
+3. **If `PLATFORMS` is set to an empty string**, `deployments/container/Makefile` fails
+   with a clear error (to avoid confusing silent buildx behavior).
+
+- Build a single-arch image with the standard Docker/Podman build flow:
+  ```bash
+  make -f deployments/container/Makefile build VERSION=<tag> IMAGE_NAME=<name|registry/name> CONTAINER_TOOL=<docker|podman>
+  ```
+- Build for specific platform(s):
+  ```bash
+  make -f deployments/container/Makefile build VERSION=<tag> IMAGE_NAME=<name|registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64'
+  ```
+- Push for current platform:
+  ```bash
+  make -f deployments/container/Makefile push VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=<docker|podman>
+  ```
+- Push for specific platform(s):
+  ```bash
+  make -f deployments/container/Makefile push VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64'
+  ```
+
+For Docker, `build` with multiple platforms performs a Buildx build without loading an image into the local Docker daemon; use `push` to publish multi-arch images.
+
+#### Multi-platform builds on Linux (amd64)
+
+Building for a platform other than your host CPU (for example `linux/arm64` on an
+`x86_64` machine) requires Docker Buildx to **run** container steps for that
+architecture during the image build. That needs either native hardware or
+userspace emulation via [QEMU and `binfmt_misc`](https://docs.docker.com/build/building/multi-platform/#qemu).
+
+- **Docker Desktop** (macOS/Windows) and many CI images ship with this enabled.
+- **Linux on amd64** often does not. The same applies to an explicit single
+  platform that does not match the host (for example `PLATFORMS=linux/arm64` on
+  x86_64): the Makefile uses buildx for that case. If a build fails with
+  `exec format error` on an `linux/arm64` step, install emulation support:
+
+  ```bash
+  docker run --privileged --rm tonistiigi/binfmt --install all
+  ```
+
+  Then create or bootstrap a buildx builder (`build` and `push` do this
+  automatically via `ensure-buildx-builder` for multi-platform or cross-platform
+  single-platform builds):
+
+  ```bash
+  make -f deployments/container/Makefile ensure-buildx-builder
+  ```
+
+  Verify with:
+
+  ```bash
+  docker run --rm --platform linux/arm64 alpine uname -m
+  ```
+
+  The output should be `aarch64`.
+
+On Apple Silicon, single-arch `linux/arm64` builds work natively; building
+`linux/amd64` uses emulation the same way.
+
 And create a `kind` cluster to run it in:
 ```bash
 ./demo/clusters/kind/create-cluster.sh
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,6 +7,7 @@ steps:
   entrypoint: make
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
+  - PLATFORMS=linux/amd64,linux/arm64,linux/ppc64le
   - DRIVER_CHART_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/charts
   args:
     - push-release-artifacts
--- a/demo/scripts/build-driver-image.sh
+++ b/demo/scripts/build-driver-image.sh
@@ -26,6 +26,7 @@ set -ex
 set -o pipefail
 
 source "${CURRENT_DIR}/common.sh"
+check_demo_config || exit 1
 
 # Create a temorary directory to hold all the artifacts we need for building the image
 TMP_DIR="$(mktemp -d)"
@@ -43,6 +44,7 @@ export IMAGE="${DRIVER_IMAGE_NAME}"
 export VERSION="${DRIVER_IMAGE_TAG}"
 export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
-# Regenerate the CRDs and build the container image
+# Regenerate the CRDs and build the container image locally (demo / kind workflow).
 make docker-generate
-make -f deployments/container/Makefile "${DRIVER_IMAGE_PLATFORM}"
+
+make -f deployments/container/Makefile "${DRIVER_IMAGE_OS}"
--- a/demo/scripts/common.sh
+++ b/demo/scripts/common.sh
@@ -29,7 +29,18 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 : ${DRIVER_IMAGE_REGISTRY:="registry.k8s.io/dra-example-driver"}
 : ${DRIVER_IMAGE_NAME:="${DRIVER_NAME}"}
 : ${DRIVER_IMAGE_TAG:="$(cat $(git rev-parse --show-toplevel)/deployments/helm/${DRIVER_NAME}/Chart.yaml | grep appVersion | sed 's/"//g' | sed -n 's/^appVersion: //p')"}
-: ${DRIVER_IMAGE_PLATFORM:="ubuntu22.04"}
+# Use DRIVER_IMAGE_OS as the canonical variable name.
+# DRIVER_IMAGE_PLATFORM is a deprecated compatibility fallback.
+if [[ -n "${DRIVER_IMAGE_PLATFORM:-}" && -z "${DRIVER_IMAGE_OS:-}" ]]; then
+    DRIVER_IMAGE_OS="${DRIVER_IMAGE_PLATFORM}"
+fi
+: ${DRIVER_IMAGE_OS:="ubuntu22.04"}
+
+# Use PLATFORMS as the canonical variable name.
+# DRIVER_IMAGE_PLATFORMS is a deprecated compatibility fallback.
+if [[ -z "${PLATFORMS:-}" && -n "${DRIVER_IMAGE_PLATFORMS:-}" ]]; then
+    PLATFORMS="${DRIVER_IMAGE_PLATFORMS}"
+fi
 
 # The kubernetes repo to build the kind cluster from
 : ${KIND_K8S_REPO:="https://github.com/kubernetes/kubernetes.git"}
@@ -65,3 +76,23 @@ if [[ -z "${CONTAINER_TOOL}" ]]; then
 fi
 
 : ${KIND:="env KIND_EXPERIMENTAL_PROVIDER=${CONTAINER_TOOL} kind"}
+
+# check_demo_config validates image-build env (DRIVER_IMAGE_OS/PLATFORM and
+# explicit PLATFORMS settings). Call from image build/push scripts only, after
+# sourcing this file.
+check_demo_config() {
+    if [[ -n "${DRIVER_IMAGE_PLATFORM:-}" && -n "${DRIVER_IMAGE_OS:-}" && "${DRIVER_IMAGE_PLATFORM}" != "${DRIVER_IMAGE_OS}" ]]; then
+        echo "Both DRIVER_IMAGE_PLATFORM and DRIVER_IMAGE_OS are set with different values."
+        echo "Use DRIVER_IMAGE_OS only, or set both to the same value."
+        return 1
+    fi
+    if [[ -n "${PLATFORMS:-}" && -n "${DRIVER_IMAGE_PLATFORMS:-}" ]]; then
+        echo "Both PLATFORMS and DRIVER_IMAGE_PLATFORMS are set."
+        echo "Use PLATFORMS only. DRIVER_IMAGE_PLATFORMS is deprecated."
+        return 1
+    fi
+    if [[ -z "${CONTAINER_TOOL}" ]]; then
+        echo "No container tool detected. Please install Docker or Podman."
+        return 1
+    fi
+}
--- a/demo/scripts/push-driver-image.sh
+++ b/demo/scripts/push-driver-image.sh
@@ -20,12 +20,21 @@ CURRENT_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 set -ex
 set -o pipefail
 
+
 source "${CURRENT_DIR}/common.sh"
+check_demo_config || exit 1
+
+cd "${CURRENT_DIR}/../.."
 
 # Set build variables
+: "${PLATFORMS:=linux/amd64,linux/arm64,linux/ppc64le}"
 export REGISTRY="${DRIVER_IMAGE_REGISTRY}"
 export IMAGE="${DRIVER_IMAGE_NAME}"
 export VERSION="${DRIVER_IMAGE_TAG}"
+export PLATFORMS
 export CONTAINER_TOOL="${CONTAINER_TOOL}"
+
+# Regenerate CRDs/deepcopy in the repo's devel container (root Makefile docker-% target).
+make docker-generate
 
 make -f deployments/container/Makefile push
--- a/deployments/container/Makefile
+++ b/deployments/container/Makefile
@@ -17,27 +17,134 @@ CONTAINER_TOOL ?= docker
 MKDIR ?= mkdir
 DISTRIBUTIONS := ubuntu22.04
 DOCKERFILE := $(CURDIR)/deployments/container/Dockerfile
+SELF_MAKEFILE := $(lastword $(MAKEFILE_LIST))
+COMMA := ,
 
 include $(CURDIR)/common.mk
 
+# Provide a usable default tag for local container builds.
+ifeq ($(strip $(VERSION)),)
+VERSION := latest
+endif
+
 IMAGE_TAG := $(VERSION)
 IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
+BASE_IMAGE ?= docker.io/ubuntu:22.04
+
+# Map host CPU to a Docker platform.
+UNAME_M ?= $(shell uname -m)
+ifeq ($(UNAME_M),x86_64)
+	HOST_PLATFORM := linux/amd64
+else ifeq ($(UNAME_M),arm64)
+	HOST_PLATFORM := linux/arm64
+else ifeq ($(UNAME_M),aarch64)
+	HOST_PLATFORM := linux/arm64
+else ifeq ($(UNAME_M),ppc64le)
+	HOST_PLATFORM := linux/ppc64le
+else
+	$(error Unsupported local platform: $(UNAME_M))
+endif
+
+# Default PLATFORMS to the host when unset. An explicit non-empty value overrides
+# detection; an explicit empty value is treated as an error to avoid confusing
+# Docker/buildx failures.
+ifeq ($(origin PLATFORMS), undefined)
+  PLATFORMS := $(HOST_PLATFORM)
+else ifeq ($(strip $(PLATFORMS)),)
+  $(error PLATFORMS is set but empty. Unset it to use the host platform, or provide one or more values)
+endif
+
+BUILDX_BUILDER ?= dra-example-driver-builder
+
+# Default to disabled so release builds are stable across older buildx setups.
+# Can be overridden by CI/release jobs that require attestations.
+BUILDX_PROVENANCE ?= false
+# Default to disabled to avoid extra metadata generation unless explicitly needed.
+BUILDX_SBOM ?= false
+BUILDX_METADATA_FLAGS = \
+	--provenance=$(BUILDX_PROVENANCE) \
+	--sbom=$(BUILDX_SBOM)
+
+# Shared docker build / buildx arguments (expanded at use time so target-specific
+# overrides such as BASE_IMAGE on distribution targets are respected).
+IMAGE_BUILD_ARGS = \
+	--build-arg GO_VERSION="$(GO_VERSION)" \
+	--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
+	--build-arg VERSION="$(VERSION)"
+
+IMAGE_BUILD_FLAGS = \
+	--pull \
+	--tag $(IMAGE) \
+	$(IMAGE_BUILD_ARGS) \
+	-f $(DOCKERFILE) \
+	$(CURDIR)
+
+BUILDX_BUILD_FLAGS = \
+	$(IMAGE_BUILD_FLAGS) \
+	--builder $(BUILDX_BUILDER) \
+	$(BUILDX_METADATA_FLAGS) \
+	--platform $(PLATFORMS)
 
 ##### Public rules #####
-.PHONY: $(DISTRIBUTIONS)
+.PHONY: $(DISTRIBUTIONS) build push ensure-buildx-builder
 
 ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
 
 # Use a generic build target to build the relevant images
 $(DISTRIBUTIONS):
-	DOCKER_BUILDKIT=1 \
-		$(CONTAINER_TOOL) build --pull \
-		--tag $(IMAGE) \
-		--build-arg GO_VERSION="$(GO_VERSION)" \
-		--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
-		--build-arg VERSION="$(VERSION)" \
-		-f $(DOCKERFILE) \
-		$(CURDIR)
+	$(MAKE) -f $(SELF_MAKEFILE) build BASE_IMAGE="$(BASE_IMAGE)"
+
+build:
+# Multi-arch builds must use buildx (docker only).
+ifneq ($(findstring $(COMMA),$(PLATFORMS)),)
+	@echo "Building image for multiple platforms with buildx: $(PLATFORMS)"
+	@if [ "$(CONTAINER_TOOL)" != "docker" ]; then \
+		echo "Multi-platform build requires CONTAINER_TOOL=docker, got $(CONTAINER_TOOL)"; \
+		exit 1; \
+	fi
+	@echo "Note: multi-platform build does not load image into local Docker daemon."
+	$(MAKE) -f $(SELF_MAKEFILE) ensure-buildx-builder
+	$(CONTAINER_TOOL) buildx build $(BUILDX_BUILD_FLAGS)
+else ifeq ($(PLATFORMS),$(HOST_PLATFORM))
+	# Native single-arch build for the host platform.
+	@echo "Building image for single platform: $(PLATFORMS)"
+	$(if $(filter docker,$(CONTAINER_TOOL)),DOCKER_BUILDKIT=1 ,)$(CONTAINER_TOOL) build --platform $(PLATFORMS) $(IMAGE_BUILD_FLAGS)
+else
+	# Cross-arch single-platform build (e.g. linux/arm64 on x86_64): use buildx --load
+	# so --platform is honoured; requires QEMU/binfmt on Linux amd64 (see README).
+	@echo "Building image for single platform: $(PLATFORMS) (host is $(HOST_PLATFORM))"
+	@if [ "$(CONTAINER_TOOL)" != "docker" ]; then \
+		echo "Cross-platform single-arch build requires CONTAINER_TOOL=docker, got $(CONTAINER_TOOL)"; \
+		exit 1; \
+	fi
+	@echo "Note: cross-platform build uses buildx and loads the image into the local Docker daemon."
+	$(MAKE) -f $(SELF_MAKEFILE) ensure-buildx-builder
+	$(CONTAINER_TOOL) buildx build $(BUILDX_BUILD_FLAGS) --load
+endif
 
 push:
+# Multi-arch pushes publish a multi-platform manifest via buildx.
+ifneq ($(findstring $(COMMA),$(PLATFORMS)),)
+	@if [ "$(CONTAINER_TOOL)" != "docker" ]; then \
+		echo "Multi-platform push requires CONTAINER_TOOL=docker, got $(CONTAINER_TOOL)"; \
+		exit 1; \
+	fi
+	@echo "Pushing image for multiple platforms with buildx: $(PLATFORMS)"
+	$(MAKE) -f $(SELF_MAKEFILE) ensure-buildx-builder
+	$(CONTAINER_TOOL) buildx build $(BUILDX_BUILD_FLAGS) --push
+else
+	# Single-arch pushes build locally first, then push the image tag.
+	$(MAKE) -f $(SELF_MAKEFILE) build
 	$(CONTAINER_TOOL) push $(IMAGE)
+endif
+
+ensure-buildx-builder:
+ifeq ($(CONTAINER_TOOL),docker)
+	@docker buildx inspect $(BUILDX_BUILDER) >/dev/null 2>&1 || \
+		docker buildx create --name $(BUILDX_BUILDER) --driver docker-container --use
+	@docker buildx use $(BUILDX_BUILDER)
+	@docker buildx inspect --bootstrap $(BUILDX_BUILDER) >/dev/null
+else
+	@echo "ensure-buildx-builder requires CONTAINER_TOOL=docker"
+	@exit 1
+endif
--- a/demo/build-driver.sh
+++ b/demo/build-driver.sh
@@ -26,6 +26,7 @@ set -ex
 set -o pipefail
 
 source "${CURRENT_DIR}/scripts/common.sh"
+check_demo_config || exit 1
 
 # Build the example driver image
 ${SCRIPTS_DIR}/build-driver-image.sh


demo/scripts/common.sh
```diff
@@ -30,6 +30,7 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 : ${DRIVER_IMAGE_NAME:="${DRIVER_NAME}"}
 : ${DRIVER_IMAGE_TAG:="$(cat $(git rev-parse --show-toplevel)/deployments/helm/${DRIVER_NAME}/Chart.yaml | grep appVersion | sed 's/"//g' | sed -n 's/^appVersion: //p')"}
 : ${DRIVER_IMAGE_PLATFORM:="ubuntu22.04"}
+: ${DRIVER_IMAGE_PLATFORMS:="linux/amd64,linux/arm64"}
```
pohly:
```suggestion
: ${DRIVER_IMAGE_ARCHITECTURES:="linux/amd64,linux/arm64"}
```

Calling both the OS and the CPU "platform" is confusing.

I also wouldn't mind renaming "DRIVER_IMAGE_PLATFORM" to "DRIVER_IMAGE_OS" and avoiding the term "platform" altogether (also below in `PLATFORMS=...`).

nojnhuh: I'd prefer to keep "platform" given that's the name of the `docker build` flag. Maybe naming the variable something like `DOCKER_BUILD_PLATFORMS` would help clarify?

deployments/container/Makefile
```diff
@@ -41,3 +43,49 @@ $(DISTRIBUTIONS):
...
+
+build-local:
+ifeq ($(CONTAINER_TOOL),docker)
+	DOCKER_BUILDKIT=1 \
+		$(CONTAINER_TOOL) buildx build --pull \
+		--platform linux/amd64 \
+		--tag $(IMAGE) \
+		--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
+		--build-arg BASE_IMAGE="docker.io/ubuntu:22.04" \
+		--build-arg VERSION="$(VERSION)" \
```
nojnhuh: Is there a way we can avoid duplicating these common arguments for each Makefile target?

demo/scripts/push-driver-image.sh
```diff
@@ -26,6 +26,11 @@ source "${CURRENT_DIR}/common.sh"
 export REGISTRY="${DRIVER_IMAGE_REGISTRY}"
 export IMAGE="${DRIVER_IMAGE_NAME}"
 export VERSION="${DRIVER_IMAGE_TAG}"
+export PLATFORMS="${DRIVER_IMAGE_PLATFORMS}"
 export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
-make -f deployments/container/Makefile push
+if [[ "${CONTAINER_TOOL}" == "docker" ]]; then
+    make -f deployments/container/Makefile push-multiarch
```
nojnhuh: It looks like this is invoking a second `docker build` after the one in `build-driver-image.sh`. Is it possible to avoid that?

cloudbuild.yaml
```diff
@@ -7,6 +7,7 @@ steps:
   entrypoint: make
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
+  - DRIVER_IMAGE_PLATFORMS=linux/amd64,linux/arm64
```
nojnhuh: I tried emulating these steps locally and ended up with a manifest linking to 4 other manifests instead of one for each of these platforms like I'd expect. Do we know what these extra `unknown/unknown` ones are?
```json
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 868,
         "digest": "sha256:f327ddbbaa1aa031b4bf42e18ec1bceb74c5d5a3d5b51c5f245d76b50d354f2f",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 868,
         "digest": "sha256:87e149e507b80ea24a2478acca86ef48734a56794d70a0652af2ee70c6649313",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 564,
         "digest": "sha256:5cd21256d66a0ab646f9b756c311b1dffb758fda143353dab049cca954330032",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 564,
         "digest": "sha256:22cbe17321a9eb991d8c30148fbdf50f280781dc2f584a7cfdf0b04887a4ac43",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      }
   ]
}
```

deployments/container/Makefile
vishalanarase: Done

demo/scripts/common.sh
vishalanarase: Done

demo/scripts/push-driver-image.sh
vishalanarase: Added check in Makefile's - push-release-artifacts target

cloudbuild.yaml
vishalanarase: Fixed by adding args to buildx `--provenance=$(BUILDX_PROVENANCE) --sbom=$(BUILDX_SBOM)`

```bash
❯ docker buildx imagetools inspect ttl.sh/dra-example-driver-1776928448:1h
Name:      ttl.sh/dra-example-driver-1776928448:1h
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:2eb40d6b0629ef0472add84d0d37cc4dc13a10dc21a6dc1d921c4809a2906db4

Manifests:
  Name:      ttl.sh/dra-example-driver-1776928448:1h@sha256:2b84ccf9d64e0aa0d27862355ecb13d031ca2c69a2a13457d4f8cb3defa68fc4
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64

  Name:      ttl.sh/dra-example-driver-1776928448:1h@sha256:b1e0da775c7385843b727ae97534fd24bfea8fe617cf524b3aab272e88285892
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64
```

demo/scripts/common.sh
```diff
@@ -29,7 +29,8 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 : ${DRIVER_IMAGE_REGISTRY:="registry.k8s.io/dra-example-driver"}
 : ${DRIVER_IMAGE_NAME:="${DRIVER_NAME}"}
 : ${DRIVER_IMAGE_TAG:="$(cat $(git rev-parse --show-toplevel)/deployments/helm/${DRIVER_NAME}/Chart.yaml | grep appVersion | sed 's/"//g' | sed -n 's/^appVersion: //p')"}
-: ${DRIVER_IMAGE_PLATFORM:="ubuntu22.04"}
+: ${DRIVER_IMAGE_OS:="${DRIVER_IMAGE_PLATFORM:-ubuntu22.04}"}
```
pohly: A comment about these env variables somewhere would be good (Along the lines of "Use DRIVER_IMAGE_OS, DRIVER_IMAGE_PLATFORM is a deprecated fallback").

demo/scripts/common.sh
```diff
@@ -29,7 +29,8 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
+: ${DOCKER_BUILD_PLATFORMS:="${DRIVER_IMAGE_PLATFORMS:-linux/amd64,linux/arm64}"}
```
pohly: Do we need two different names for the same thing here and should it be "PLATFORMS" or "ARCHITECTURES"?

demo/scripts/common.sh
vishalanarase: Comment added

demo/scripts/common.sh
vishalanarase: I kept `DOCKER_BUILD_PLATFORMS` as the canonical variable to align with Docker Buildx (--platform) terminology and retained `DRIVER_IMAGE_PLATFORMS` only as a deprecated compatibility fallback. I also added comments in `demo/scripts/common.sh` to make this explicit

demo/scripts/common.sh
```diff
@@ -29,7 +29,20 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 : ${DRIVER_IMAGE_REGISTRY:="registry.k8s.io/dra-example-driver"}
 : ${DRIVER_IMAGE_NAME:="${DRIVER_NAME}"}
 : ${DRIVER_IMAGE_TAG:="$(cat $(git rev-parse --show-toplevel)/deployments/helm/${DRIVER_NAME}/Chart.yaml | grep appVersion | sed 's/"//g' | sed -n 's/^appVersion: //p')"}
-: ${DRIVER_IMAGE_PLATFORM:="ubuntu22.04"}
+# Use DRIVER_IMAGE_OS as the canonical variable name.
+# DRIVER_IMAGE_PLATFORM is a deprecated compatibility fallback.
+if [[ -n "${DRIVER_IMAGE_PLATFORM:-}" && -z "${DRIVER_IMAGE_OS:-}" ]]; then
+    DRIVER_IMAGE_OS="${DRIVER_IMAGE_PLATFORM}"
+fi
```
bart0sh, line 36: If both variables are set this code silently ignores DRIVER_IMAGE_PLATFORM. Should it be considered as an error to avoid user confusion?

Makefile
```diff
@@ -172,5 +172,9 @@ push-release-artifacts:
...
 		demo/scripts/push-driver-chart.sh
 	export DRIVER_IMAGE_TAG="${IMAGE_GIT_TAG}"; \
-	demo/scripts/build-driver-image.sh && \
-	demo/scripts/push-driver-image.sh
+	if [ "$(CONTAINER_TOOL)" = "docker" ]; then \
+		demo/scripts/push-driver-image.sh; \
+	else \
+		demo/scripts/build-driver-image.sh && \
+		demo/scripts/push-driver-image.sh; \
+	fi
```
bart0sh: This looks confusing to me. I read it as `if CONTAINER_TOOL is "docker" we don't build the image`.

deployments/container/Makefile
```diff
@@ -41,3 +52,41 @@ $(DISTRIBUTIONS):
...
+	@echo "ensure-buildx-builder requires CONTAINER_TOOL=docker"
+	@exit 1
+endif
+
+build-local:
+ifeq ($(CONTAINER_TOOL),docker)
+	DOCKER_BUILDKIT=1 \
+		$(CONTAINER_TOOL) buildx build \
+		$(BUILDX_COMMON_ARGS) \
+		--platform linux/amd64 \
```
bart0sh: How this would work on linux/arm64 or Mac/Apple Silicon?

deployments/container/Makefile
```diff
@@ -41,3 +52,41 @@ $(DISTRIBUTIONS):
...
+		--load
+else
+	@echo "build-local requires CONTAINER_TOOL=docker (uses buildx --load)"
+	@exit 1
+endif
+
+push-multiarch: ensure-buildx-builder
+ifeq ($(CONTAINER_TOOL),docker)
+	DOCKER_BUILDKIT=1 \
```
bart0sh: Do we need to set DOCKER_BUILDKIT=1 if we use buildx anyway? I thought it's a legacy var that enables buildkit on old docker versions.

README.md
```diff
@@ -48,6 +48,23 @@ From here we will build the image for the example resource driver:
...
+  make -f deployments/container/Makefile ubuntu22.04 VERSION=<tag> IMAGE_NAME=<name> CONTAINER_TOOL=<docker|podman>
+  ```
+- Build and load a local Docker image (visible via `docker images`):
+  ```bash
+  make -f deployments/container/Makefile build-local VERSION=<tag> IMAGE_NAME=<name> CONTAINER_TOOL=docker
+  ```
+- Build and push a multi-arch image (`linux/amd64,linux/arm64`) with Docker Buildx:
+  ```bash
+  make -f deployments/container/Makefile push-multiarch VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=docker
+  ```
```
bart0sh, line 91: This makes the user to use different commands instead of simply specifying the set of platforms user wants to build (or push). Would it be better to provide just 2 commands: build and push plus a set of platforms (PLATFORMS), e.g.

- build for current platform: make -f deployments/container/Makefile build
- build for specific platform(s): make -f deployments/container/Makefile build PLATFORMS='linux/amd64,linux/arm64'
- push for current platform (depends on buid): make -f deployments/container/Makefile push
- push for specific platform(s) (depends on build): make -f deployments/container/Makefile push PLATFORMS='linux/amd64,linux/arm64'

And let the build machinery to figure out the actual commands to run?

Makefile
vishalanarase: It's not like that, If that var is set then it's calling `demo/scripts/push-driver-image.sh` and it's calling `make -f deployments/container/Makefile push` it's building and pushing images using buildx (it's doing build and push together)

demo/scripts/common.sh
vishalanarase: Will add check and error out

deployments/container/Makefile
vishalanarase: Will remove it

deployments/container/Makefile
vishalanarase: Identify platform using `uname -m` and use to build image

README.md
vishalanarase: Done

deployments/container/Makefile
```diff
@@ -38,6 +90,36 @@ $(DISTRIBUTIONS):
...
+else
+	@echo "ensure-buildx-builder requires CONTAINER_TOOL=docker"
+	@exit 1
+endif
+
+build-local:
+	$(MAKE) -f $(SELF_MAKEFILE) build PLATFORMS="$(LOCAL_PLATFORM)"
+
+push-multiarch: ensure-buildx-builder
+	$(MAKE) -f $(SELF_MAKEFILE) push PLATFORMS="$(MULTIARCH_PLATFORMS)"
```
bart0sh:
```suggestion
build:
	$(MAKE) -f $(SELF_MAKEFILE) build PLATFORMS="$(LOCAL_PLATFORM)"

push:
	$(MAKE) -f $(SELF_MAKEFILE) push PLATFORMS="$(MULTIARCH_PLATFORMS)"
```

deployments/container/Makefile
```diff
@@ -17,19 +17,71 @@ CONTAINER_TOOL ?= docker
...
+UNAME_M ?= $(shell uname -m)
+ifeq ($(UNAME_M),x86_64)
+LOCAL_PLATFORM ?= linux/amd64
+else ifeq ($(UNAME_M),arm64)
+LOCAL_PLATFORM ?= linux/arm64
+else ifeq ($(UNAME_M),aarch64)
+LOCAL_PLATFORM ?= linux/arm64
+else
+LOCAL_PLATFORM ?= linux/amd64
+endif
```
bart0sh, line 55:
```suggestion
UNAME_M ?= $(shell uname -m)
ifeq ($(UNAME_M),x86_64)
	LOCAL_PLATFORM ?= linux/amd64
else ifeq ($(UNAME_M),arm64)
	LOCAL_PLATFORM ?= linux/arm64
else ifeq ($(UNAME_M),aarch64)
	LOCAL_PLATFORM ?= linux/arm64
else
	LOCAL_PLATFORM ?= linux/amd64
endif
```

deployments/container/Makefile
vishalanarase: We already have build & push targets in place, this were added to just test on local machine

vishalanarase: We can remove if you think it's not needed for local test

bart0sh: Isn't `build-local` is the same as `build`? If not, it should be, I believe.
`push-multiarch` should be the same as `make push PLATFORMS='linux/amd64,linux/arm64'`, unless I missed something.

deployments/container/Makefile
```diff
@@ -17,19 +17,71 @@ CONTAINER_TOOL ?= docker
...
+# used by build-local (`linux/amd64` on x86_64, `linux/arm64` on arm64/aarch64).
+UNAME_M ?= $(shell uname -m)
+ifeq ($(UNAME_M),x86_64)
+	LOCAL_PLATFORM ?= linux/amd64
+else ifeq ($(UNAME_M),arm64)
+	LOCAL_PLATFORM ?= linux/arm64
+else ifeq ($(UNAME_M),aarch64)
+	LOCAL_PLATFORM ?= linux/arm64
+else
+	LOCAL_PLATFORM ?= linux/amd64
```
bart0sh: Please avoid silent assumptions, they can cause hard to debug issues.
```suggestion
else:
	@echo "unsupported local platform: @LOCAL_PLATFORM"
	@exit 1
```

deployments/container/Makefile
```diff
@@ -17,19 +17,71 @@ CONTAINER_TOOL ?= docker
...
+endif
+
+BUILDX_BUILDER ?= dra-example-driver-builder
+BUILDX_PROVENANCE ?= false
+BUILDX_SBOM ?= false
+BUILDX_COMMON_ARGS = --pull \
+	--tag $(IMAGE) \
+	--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
+	--build-arg BASE_IMAGE="docker.io/ubuntu:22.04" \
```
bart0sh:
```suggestion
	--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
```

deployments/container/Makefile
```diff
@@ -17,19 +17,71 @@ CONTAINER_TOOL ?= docker
...
+# Provide a usable default tag for local container builds.
+ifeq ($(strip $(VERSION)),)
+VERSION := latest
+endif
+
 IMAGE_TAG := $(VERSION)
 IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
+BASE_IMAGE ?= docker.io/ubuntu:22.04
+PLATFORMS ?= $(LOCAL_PLATFORM)
+MULTIARCH_PLATFORMS ?= linux/amd64,linux/arm64
```
bart0sh: This complicates UI. Why isn't PLATFORMS not enough?

deployments/container/Makefile
bart0sh: Should we just use PLATFORM as is and default it to local platform if it's not specified?

deployments/container/Makefile
```diff
@@ -17,19 +17,71 @@ CONTAINER_TOOL ?= docker
...
+	LOCAL_PLATFORM ?= linux/arm64
+else ifeq ($(UNAME_M),aarch64)
+	LOCAL_PLATFORM ?= linux/arm64
+else
+	LOCAL_PLATFORM ?= linux/amd64
+endif
+
+BUILDX_BUILDER ?= dra-example-driver-builder
+BUILDX_PROVENANCE ?= false
+BUILDX_SBOM ?= false
```
bart0sh, line 63: These 2 values require explanations.

demo/scripts/common.sh
```diff
@@ -29,7 +29,25 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
 : ${DRIVER_IMAGE_REGISTRY:="registry.k8s.io/dra-example-driver"}
 : ${DRIVER_IMAGE_NAME:="${DRIVER_NAME}"}
 : ${DRIVER_IMAGE_TAG:="$(cat $(git rev-parse --show-toplevel)/deployments/helm/${DRIVER_NAME}/Chart.yaml | grep appVersion | sed 's/"//g' | sed -n 's/^appVersion: //p')"}
-: ${DRIVER_IMAGE_PLATFORM:="ubuntu22.04"}
+# Use DRIVER_IMAGE_OS as the canonical variable name.
+# DRIVER_IMAGE_PLATFORM is a deprecated compatibility fallback.
+if [[ -n "${DRIVER_IMAGE_PLATFORM:-}" && -n "${DRIVER_IMAGE_OS:-}" && "${DRIVER_IMAGE_PLATFORM}" != "${DRIVER_IMAGE_OS}" ]]; then
+    echo "Both DRIVER_IMAGE_PLATFORM and DRIVER_IMAGE_OS are set with different values."
+    echo "Use DRIVER_IMAGE_OS only, or set both to the same value."
+    return 1
```
bart0sh: This requires checking return value when sourcing this file.

Makefile
bart0sh: I understand that it's not like that, but it looks like that.

cloudbuild.yaml
```diff
@@ -7,6 +7,7 @@ steps:
   entrypoint: make
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
+  - DOCKER_BUILD_PLATFORMS=linux/amd64,linux/arm64
```
bart0sh:
```suggestion
  - PLATFORMS=linux/amd64,linux/arm64
```

demo/scripts/build-driver-image.sh
```diff
@@ -45,4 +45,4 @@ export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
 # Regenerate the CRDs and build the container image
 make docker-generate
```
bart0sh, line 48: `make docker-generate` will be skipped for `push-release-artifacts` target for docker builder as we don't call `build-driver-image.sh` there. Even if it's ok, it looks confusing.

Makefile
bart0sh: I'd suggest not to change this Makefile, but do the branching in the build-driver-image.sh instead. That would be much more readable, I hope.

demo/scripts/build-driver-image.sh
```diff
@@ -43,6 +43,11 @@ export IMAGE="${DRIVER_IMAGE_NAME}"
 export VERSION="${DRIVER_IMAGE_TAG}"
 export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
-# Regenerate the CRDs and build the container image
-make docker-generate
```
bart0sh, line 47: so, this is not needed at all? Can you explain why?

demo/scripts/build-driver-image.sh
```diff
@@ -43,6 +43,11 @@ export IMAGE="${DRIVER_IMAGE_NAME}"
...
-make -f deployments/container/Makefile "${DRIVER_IMAGE_PLATFORM}"
+# Docker release path builds and pushes in one step via buildx in
+# demo/scripts/push-driver-image.sh, so skip the standalone build here.
```
bart0sh: As far as I understand we still can build with docker for one arch, like we did before. Would it make sense to do it here?

deployments/container/Makefile
```diff
@@ -17,19 +17,76 @@ CONTAINER_TOOL ?= docker
...
+	LOCAL_PLATFORM ?= linux/amd64
+else ifeq ($(UNAME_M),arm64)
+	LOCAL_PLATFORM ?= linux/arm64
+else ifeq ($(UNAME_M),aarch64)
+	LOCAL_PLATFORM ?= linux/arm64
+else
+	@echo "Unsupported local platform: $(UNAME_M)"
+	@exit 1
+endif
+
```
bart0sh: This only makes sense to do if PLATFORMS is not set, right?

deployments/container/Makefile
```diff
@@ -17,19 +17,76 @@ CONTAINER_TOOL ?= docker
...
+# Default to disabled to avoid extra metadata generation unless explicitly needed.
+BUILDX_SBOM ?= false
+
+BUILDX_COMMON_ARGS = --pull \
+	--tag $(IMAGE) \
+	--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
+	--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
+	--build-arg VERSION="$(VERSION)" \
+	-f $(DOCKERFILE) \
+	$(CURDIR)
```
bart0sh: This makes sense to set up only for docker platform and most probably only for multi-arch build.

deployments/container/Makefile
```diff
@@ -38,6 +95,30 @@ $(DISTRIBUTIONS):
 		--build-arg VERSION="$(VERSION)" \
 		-f $(DOCKERFILE) \
 		$(CURDIR)
+endif
 
 push:
+ifeq ($(CONTAINER_TOOL),docker)
```
bart0sh: instead of checking container tool, I'd suggest to check one or multi-arch platform and only use buildx for multi-arch. That would make the change much more clear in my opinion.

deployments/container/Makefile
```diff
@@ -17,19 +17,75 @@ CONTAINER_TOOL ?= docker
...
+		--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
+		--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
+		--build-arg VERSION="$(VERSION)" \
+		-f $(DOCKERFILE) \
+		$(CURDIR) \
+		--platform $(PLATFORMS)
+else
+	@echo "Multi-platform build requires CONTAINER_TOOL=docker, got $(CONTAINER_TOOL)"
+	@exit 1
+endif
```
bart0sh:
```suggestion
	ifneq ($(CONTAINER_TOOL),docker)
		@echo "Multi-platform build requires CONTAINER_TOOL=docker, got $(CONTAINER_TOOL)"
		@exit 1
	endif
	@echo "Note: multi-platform build does not load image into local Docker daemon."
	$(CONTAINER_TOOL) buildx build \
		--pull \
		--tag $(IMAGE) \
		--build-arg GOLANG_VERSION="$(GOLANG_VERSION)" \
		--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
		--build-arg VERSION="$(VERSION)" \
		-f $(DOCKERFILE) \
		$(CURDIR) \
		--platform $(PLATFORMS)
endif
```

deployments/container/Makefile
```diff
@@ -38,6 +94,43 @@ $(DISTRIBUTIONS):
 		--build-arg VERSION="$(VERSION)" \
 		-f $(DOCKERFILE) \
 		$(CURDIR)
+endif
 
 push:
+# Multi-arch pushes publish a multi-platform manifest via buildx.
+ifneq ($(findstring $(COMMA),$(PLATFORMS)),)
```
bart0sh, line 127: the same here - please use indentation and move the error check to the top, thanks.

demo/scripts/common.sh
```diff
@@ -29,7 +29,26 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
+fi
+if [[ -n "${DRIVER_IMAGE_PLATFORM:-}" && -z "${DRIVER_IMAGE_OS:-}" ]]; then
+    DRIVER_IMAGE_OS="${DRIVER_IMAGE_PLATFORM}"
+fi
+: ${DRIVER_IMAGE_OS:="ubuntu22.04"}
+
+# Use PLATFORMS as the canonical variable name.
+# DRIVER_IMAGE_PLATFORMS is a deprecated compatibility fallback.
```
bart0sh, line 40: What will happen if both PLATFORMS and DRIVER_IMAGE_PLATFORMS are set?

deployments/container/Makefile
```diff
@@ -17,19 +17,74 @@ CONTAINER_TOOL ?= docker
...
+
+# Default to disabled so release builds are stable across older buildx setups.
+# Can be overridden by CI/release jobs that require attestations.
+BUILDX_PROVENANCE ?= false
+# Default to disabled to avoid extra metadata generation unless explicitly needed.
+BUILDX_SBOM ?= false
 
 ##### Public rules #####
-.PHONY: $(DISTRIBUTIONS)
+.PHONY: $(DISTRIBUTIONS) build push build-local push-multiarch ensure-buildx-builder
```
bart0sh:
```suggestion
.PHONY: $(DISTRIBUTIONS) build push ensure-buildx-builder
```

deployments/container/Makefile
```diff
@@ -17,19 +17,74 @@ CONTAINER_TOOL ?= docker
...
+endif
+
 IMAGE_TAG := $(VERSION)
 IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
+BASE_IMAGE ?= docker.io/ubuntu:22.04
+
+# Only detect host architecture when PLATFORMS is not provided explicitly.
+ifeq ($(strip $(PLATFORMS)),)
+	# Detect host CPU architecture via `uname -m` and map it to a Docker platform
+	# used by build-local (`linux/amd64` on x86_64, `linux/arm64` on arm64/aarch64).
```
bart0sh:
```suggestion
	# used by build (`linux/amd64` on x86_64, `linux/arm64` on arm64/aarch64).
```

demo/scripts/common.sh
vishalanarase: `PLATFORMS` is the canonical variable and takes precedence.
`DRIVER_IMAGE_PLATFORMS` is only used as a backward compatibility fallback when `PLATFORMS` is unset/empty.

bart0sh: Should we consider this a error if both are set?

demo/scripts/build-driver-image.sh
```diff
@@ -45,4 +45,12 @@ export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
 # Regenerate the CRDs and build the container image
 make docker-generate
-make -f deployments/container/Makefile "${DRIVER_IMAGE_PLATFORM}"
+
+# In push-release-artifacts, Docker multi-arch builds and pushes in one step via
+# buildx in demo/scripts/push-driver-image.sh, so skip standalone build there.
+if [[ "${SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH:-}" == "1" && "${CONTAINER_TOOL}" == "docker" && "${PLATFORMS}" == *,* ]]; then
```
bart0sh: What happens if user sets SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH=0 ?

demo/scripts/common.sh
bart0sh: Should we move these checks out of common.sh so it can be sourced without checking errors? We only need this check in a couple of places, but because we added it to the common.sh we now have to check its return code everywhere.

test/e2e/e2e_setup_test.go
```diff
@@ -69,6 +69,8 @@ func init() {
 const (
 	checkPodLogsTimeout  = "30s"
 	checkPodLogsInterval = "1s"
+	webhookReadyTimeout  = "120s"
+	webhookReadyInterval = "1s"
```
bart0sh: How is this change related to this PR?

test/e2e/setup-e2e.sh
```diff
@@ -33,6 +33,8 @@ helm upgrade -i \
 helm upgrade -i \
   --create-namespace \
   --namespace dra-example-driver \
+  --wait \
+  --timeout 5m \
```
bart0sh: The same here. Let's concentrate on building and pushing for multi-arch. Isn't it a purpose of this PR?

deployments/container/Makefile
```diff
@@ -17,19 +17,74 @@ CONTAINER_TOOL ?= docker
...
 
 ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
 
 # Use a generic build target to build the relevant images
 $(DISTRIBUTIONS):
+	$(MAKE) -f $(SELF_MAKEFILE) build
+
+build:
+# Multi-arch builds must use buildx (docker only).
+ifneq ($(findstring $(COMMA),$(PLATFORMS)),)
```
bart0sh, line 99: I'm not sure we need a variable for comma, let's use it directly as it's more readable.

Makefile
```diff
@@ -172,5 +172,5 @@ push-release-artifacts:
 		HELM=$(HELM) \
 		demo/scripts/push-driver-chart.sh
 	export DRIVER_IMAGE_TAG="${IMAGE_GIT_TAG}"; \
-	demo/scripts/build-driver-image.sh && \
+	SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH=1 demo/scripts/build-driver-image.sh && \
```
bart0sh: I'm still not sure I understand the purpose of SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH variable to be honest. If user sets it to 0, what's expected to happen?

vishalanarase: SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH is only meant for the `push-release-artifacts` path. There we already run make docker-generate in `build-driver-image.sh` and the actual image is built and published in one step by `push-driver-image.sh` -> `make -f deployments/container/Makefile push` with `PLATFORMS=linux/amd64,linux/arm64` (buildx --push). So we set the variable to 1 only to skip the extra local `make -f deployments/container/Makefile <os>` image build in `build-driver-image.sh`, which would otherwise produce a single-arch image that we do not use for that release flow.

The skip runs only when the value is exactly 1 and CONTAINER_TOOL=docker and PLATFORMS contains a comma (multi-arch). If a user sets it to 0 or leaves it unset, that condition is false and behavior is unchanged: `build-driver-image.sh` runs `docker-generate` and then the normal local container build before any later push step.

deployments/container/Makefile
vishalanarase: Done

nojnhuh:
Sorry, I haven't been keeping close tabs on these changes lately.

Overall this seems like a lot of change for something I'd expect to be a bit more straightforward. Has the existing structure of the scripts backed us into a corner that's making this more difficult? I'm not married to the current layout if reorganizing things makes changes like this simpler to do in the future.

vishalanarase:
> Sorry, I haven't been keeping close tabs on these changes lately.
> 
> Overall this seems like a lot of change for something I'd expect to be a bit more straightforward. Has the existing structure of the scripts backed us into a corner that's making this more difficult? I'm not married to the current layout if reorganizing things makes changes like this simpler to do in the future.

From my side, the initial change was simple, but I’ve been iterating based on multiple rounds of review comments, so the scope has gradually expanded.

Could you take a look at the current state and share your feedback on the approach?

vishalanarase:
@bart0sh @nojnhuh Can you please take look?

bart0sh:
> @bart0sh @nojnhuh Can you please take look?

Can you address [this comment](https://github.com/kubernetes-sigs/dra-example-driver/pull/167/changes#r3161153271), please?

deployments/container/Makefile
bart0sh: > Done

$ git grep 'build-arg BASE_IMAGE' deployments/container/Makefile
deployments/container/Makefile:         --build-arg BASE_IMAGE="$(BASE_IMAGE)" \
deployments/container/Makefile:         --build-arg BASE_IMAGE="$(BASE_IMAGE)" \
deployments/container/Makefile:         --build-arg BASE_IMAGE="$(BASE_IMAGE)" \

bart0sh:
@vishalanarase multi-arch build breaks this way in my environment:

```
$ make -f deployments/container/Makefile build PLATFORMS='linux/arm64,linux/amd64'
Building image for multiple platforms with buildx: linux/arm64,linux/amd64
Note: multi-platform build does not load image into local Docker daemon.
docker buildx build \
	--pull \
	--tag registry.example.com/dra-example-driver:latest \
	--build-arg GOLANG_VERSION="1.26.2" \
	--build-arg BASE_IMAGE="docker.io/ubuntu:22.04" \
	--build-arg VERSION="latest" \
	-f /home/ed/git/dra-example-driver/deployments/container/Dockerfile \
	/home/ed/git/dra-example-driver \
	--platform linux/arm64,linux/amd64
[+] Building 22.7s (20/23)                                                                                                                                                                         docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                         0.0s
 => => transferring dockerfile: 1.36kB                                                                                                                                                                       0.0s
 => [linux/arm64 internal] load metadata for docker.io/library/ubuntu:22.04                                                                                                                                  0.8s
 => [linux/arm64 internal] load metadata for docker.io/library/golang:1.26.2                                                                                                                                 0.8s
 => [linux/amd64 internal] load metadata for docker.io/library/ubuntu:22.04                                                                                                                                  0.8s
 => [linux/amd64 internal] load metadata for docker.io/library/golang:1.26.2                                                                                                                                 0.8s
 => [internal] load .dockerignore                                                                                                                                                                            0.0s
 => => transferring context: 2B                                                                                                                                                                              0.0s
 => [linux/amd64 stage-1 1/3] FROM docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                                    0.1s
 => => resolve docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                                                        0.1s
 => [internal] load build context                                                                                                                                                                           18.3s
 => => transferring context: 1.02GB                                                                                                                                                                         18.1s
 => [linux/amd64 build 1/5] FROM docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                     0.1s
 => => resolve docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                                       0.1s
 => CACHED [linux/arm64 stage-1 1/3] FROM docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                             0.1s
 => => resolve docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                                                        0.1s
 => [linux/arm64 build 1/5] FROM docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                     0.1s
 => => resolve docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                                       0.1s
 => CACHED [linux/arm64 build 2/5] WORKDIR /build                                                                                                                                                            0.0s
 => CACHED [linux/arm64 build 3/5] COPY . .                                                                                                                                                                  0.0s
 => ERROR [linux/arm64 build 4/5] RUN mkdir /artifacts                                                                                                                                                       3.1s
 => CACHED [linux/amd64 build 2/5] WORKDIR /build                                                                                                                                                            0.0s
 => CACHED [linux/amd64 build 3/5] COPY . .                                                                                                                                                                  0.0s
 => CACHED [linux/amd64 build 4/5] RUN mkdir /artifacts                                                                                                                                                      0.0s
 => CACHED [linux/amd64 build 5/5] RUN make PREFIX=/artifacts cmds                                                                                                                                           0.0s
 => CACHED [linux/amd64 stage-1 2/3] COPY --from=build /artifacts/dra-example-kubeletplugin /usr/bin/dra-example-kubeletplugin                                                                               0.0s
 => CACHED [linux/amd64 stage-1 3/3] COPY --from=build /artifacts/dra-example-webhook       /usr/bin/dra-example-webhook                                                                                     0.0s
------                                                                                                                                                                                                            
 > [linux/arm64 build 4/5] RUN mkdir /artifacts:
0.326 exec /bin/sh: exec format error
------
Dockerfile:22
--------------------
  20 |     COPY . .
  21 |     
  22 | >>> RUN mkdir /artifacts
  23 |     RUN make PREFIX=/artifacts cmds
  24 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c mkdir /artifacts" did not complete successfully: exit code: 255
make: *** [deployments/container/Makefile:75: build] Error 1
```

Am I missing something or it requries something (qemu?) for arm build? Should it be explained in the doc?

Makefile
bart0sh: Can you explain what's `docker-generate`? I only see a caller side:
```
$ git grep docker-generate
demo/scripts/build-driver-image.sh:make docker-generate
```

Anyway, SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH still looks like a workaround to me.

bart0sh: found docker-generate target:
$(DOCKER_TARGETS): docker-%: .build-image

bart0sh: Regarding SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH, would calling 'make docker-generate' from `push-driver-image.sh` help to drop it?

vishalanarase:
> @vishalanarase multi-arch build breaks this way in my environment:
> 
> ```
> $ make -f deployments/container/Makefile build PLATFORMS='linux/arm64,linux/amd64'
> Building image for multiple platforms with buildx: linux/arm64,linux/amd64
> Note: multi-platform build does not load image into local Docker daemon.
> docker buildx build \
> 	--pull \
> 	--tag registry.example.com/dra-example-driver:latest \
> 	--build-arg GOLANG_VERSION="1.26.2" \
> 	--build-arg BASE_IMAGE="docker.io/ubuntu:22.04" \
> 	--build-arg VERSION="latest" \
> 	-f /home/ed/git/dra-example-driver/deployments/container/Dockerfile \
> 	/home/ed/git/dra-example-driver \
> 	--platform linux/arm64,linux/amd64
> [+] Building 22.7s (20/23)                                                                                                                                                                         docker:default
>  => [internal] load build definition from Dockerfile                                                                                                                                                         0.0s
>  => => transferring dockerfile: 1.36kB                                                                                                                                                                       0.0s
>  => [linux/arm64 internal] load metadata for docker.io/library/ubuntu:22.04                                                                                                                                  0.8s
>  => [linux/arm64 internal] load metadata for docker.io/library/golang:1.26.2                                                                                                                                 0.8s
>  => [linux/amd64 internal] load metadata for docker.io/library/ubuntu:22.04                                                                                                                                  0.8s
>  => [linux/amd64 internal] load metadata for docker.io/library/golang:1.26.2                                                                                                                                 0.8s
>  => [internal] load .dockerignore                                                                                                                                                                            0.0s
>  => => transferring context: 2B                                                                                                                                                                              0.0s
>  => [linux/amd64 stage-1 1/3] FROM docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                                    0.1s
>  => => resolve docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                                                        0.1s
>  => [internal] load build context                                                                                                                                                                           18.3s
>  => => transferring context: 1.02GB                                                                                                                                                                         18.1s
>  => [linux/amd64 build 1/5] FROM docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                     0.1s
>  => => resolve docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                                       0.1s
>  => CACHED [linux/arm64 stage-1 1/3] FROM docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                             0.1s
>  => => resolve docker.io/library/ubuntu:22.04@sha256:962f6cadeae0ea6284001009daa4cc9a8c37e75d1f5191cf0eb83fe565b63dd7                                                                                        0.1s
>  => [linux/arm64 build 1/5] FROM docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                     0.1s
>  => => resolve docker.io/library/golang:1.26.2@sha256:b54cbf583d390341599d7bcbc062425c081105cc5ef6d170ced98ef9d047c716                                                                                       0.1s
>  => CACHED [linux/arm64 build 2/5] WORKDIR /build                                                                                                                                                            0.0s
>  => CACHED [linux/arm64 build 3/5] COPY . .                                                                                                                                                                  0.0s
>  => ERROR [linux/arm64 build 4/5] RUN mkdir /artifacts                                                                                                                                                       3.1s
>  => CACHED [linux/amd64 build 2/5] WORKDIR /build                                                                                                                                                            0.0s
>  => CACHED [linux/amd64 build 3/5] COPY . .                                                                                                                                                                  0.0s
>  => CACHED [linux/amd64 build 4/5] RUN mkdir /artifacts                                                                                                                                                      0.0s
>  => CACHED [linux/amd64 build 5/5] RUN make PREFIX=/artifacts cmds                                                                                                                                           0.0s
>  => CACHED [linux/amd64 stage-1 2/3] COPY --from=build /artifacts/dra-example-kubeletplugin /usr/bin/dra-example-kubeletplugin                                                                               0.0s
>  => CACHED [linux/amd64 stage-1 3/3] COPY --from=build /artifacts/dra-example-webhook       /usr/bin/dra-example-webhook                                                                                     0.0s
> ------                                                                                                                                                                                                            
>  > [linux/arm64 build 4/5] RUN mkdir /artifacts:
> 0.326 exec /bin/sh: exec format error
> ------
> Dockerfile:22
> --------------------
>   20 |     COPY . .
>   21 |     
>   22 | >>> RUN mkdir /artifacts
>   23 |     RUN make PREFIX=/artifacts cmds
>   24 |     
> --------------------
> ERROR: failed to solve: process "/bin/sh -c mkdir /artifacts" did not complete successfully: exit code: 255
> make: *** [deployments/container/Makefile:75: build] Error 1
> ```
> 
> Am I missing something or it requries something (qemu?) for arm build? Should it be explained in the doc?

>it requries something (qemu?)
Yes, you can refer -  https://docs.docker.com/build/building/multi-platform/#qemu

What is you host platform? will add doc to build an image other than your host platform

Makefile
vishalanarase: docker-generate is the root Makefile docker-% pattern, it runs make generate in the devel container (CRD/deepcopy).

Removed SKIP_LOCAL_BUILD_FOR_DOCKER_MULTIARCH and push-driver-image.sh now runs docker-generate then make push. build-driver-image.sh stays for local/demo (generate + single-arch build). push-release-artifacts only calls push-driver-image.sh for the image.

deployments/container/Makefile
vishalanarase: Done

bart0sh:
@vishalanarase 
> What is you host platform? will add doc to build an image other than your host platform

$ uname -a
Linux dev 6.17.0-19-generic #19~24.04.2-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar  6 23:08:46 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

demo/clusters/kind/create-cluster.sh
```diff
@@ -23,6 +23,7 @@ set -ex
 set -o pipefail
 
 source "${CURRENT_DIR}/../../scripts/common.sh"
+check_demo_config || exit 1
```
bart0sh: I don't think we need to check everywhere. The check only makes sense for the build-related targets that use DRIVER_IMAGE_PLATFORM, PLATFORM or CONTAINER_TOOL.

vishalanarase: Agreed, updating PR

deployments/container/Makefile
```diff
@@ -17,27 +17,116 @@ CONTAINER_TOOL ?= docker
...
+ifeq ($(strip $(VERSION)),)
+VERSION := latest
+endif
+
 IMAGE_TAG := $(VERSION)
 IMAGE = $(IMAGE_NAME):$(IMAGE_TAG)
+BASE_IMAGE ?= docker.io/ubuntu:22.04
+
+# Only detect host architecture when PLATFORMS is not provided explicitly.
+ifeq ($(strip $(PLATFORMS)),)
```
bart0sh: If user wants to build image for one platform (e.g. PLATFORMS='linux/arm64' on the x86_64 machine), this will silently skip platform detection, I guess.

bart0sh:
@vishalanarase Please address remaining comments and squash commits into one, thanks!

deployments/container/Makefile
vishalanarase: Correct, When PLATFORMS is set (even to a single value), we skip host detection, explicit PLATFORMS overrides the default. Detection only runs when PLATFORMS is unset (local "build for this machine" case). Added a comment in the Makefile to spell that out. Single-arch docker build now passes `--platform $(PLATFORMS)` so e.g. `PLATFORMS=linux/arm64` on x86_64 is honoured, cross-arch still needs QEMU as documented in the README.

vishalanarase:
> @vishalanarase Please address remaining comments and squash commits into one, thanks!

@bart0sh done, thank you!

deployments/container/Makefile
bart0sh: > explicit PLATFORMS overrides the default.

Does it? In my environment it ends up calling `docker build` (not buildx), which builds for the current platform. This made me think that PLATFORMS value is ignored. Here is an example:

```
$ make -f deployments/container/Makefile build PLATFORMS='linux/arm64'
# Single-arch builds use the standard container build flow.
Building image for single platform: linux/arm64
docker build --pull --tag registry.example.com/dra-example-driver:latest --build-arg GO_VERSION="1.26.2" --build-arg BASE_IMAGE="docker.io/ubuntu:22.04" --build-arg VERSION="latest" -f /home/ed/git/dra-example-driver/deployments/container/Dockerfile /home/ed/git/dra-example-driver
...
```
It says that it builds for linux/arm64, but in fact it doesn't. It builds for the current platform, which is linux/amd64 in my case.

demo/scripts/common.sh
```diff
@@ -29,7 +29,19 @@ SCRIPTS_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
+if [[ -z "${PLATFORMS:-}" && -n "${DRIVER_IMAGE_PLATFORMS:-}" ]]; then
+    PLATFORMS="${DRIVER_IMAGE_PLATFORMS}"
+fi
+: ${PLATFORMS:="linux/amd64,linux/arm64"}
```
prb112: Hey @Shilpa-Gokul heads up.... it'll break existing builds. Right? the implicit arch is ignored. Thanks, Paul

deployments/container/Makefile
vishalanarase: You’re right, Without `--platform`, single-arch docker build used the host arch despite the log.

Fixed: native builds pass `--platform $(PLATFORMS)` cross-arch single-platform (e.g. linux/arm64 on x86_64) uses buildx build `--platform … --load`. Please pull latest and re-test. You should see buildx with `--platform linux/arm64`. QEMU/binfmt still needed on Linux amd64 (README).

deployments/container/Makefile
```diff
@@ -17,27 +17,134 @@ CONTAINER_TOOL ?= docker
...
+	$(error Unsupported local platform: $(UNAME_M))
+endif
+
+# Default PLATFORMS to the host when unset. An explicit value overrides detection.
+ifeq ($(strip $(PLATFORMS)),)
+	PLATFORMS := $(HOST_PLATFORM)
+endif
+
+# Split PLATFORMS on commas; more than one entry implies a multi-platform (buildx) build.
+PLATFORM_LIST := $(subst $(shell printf ','), ,$(PLATFORMS))
```
bart0sh: nit: Build would incorrectly perform a multi-arch build and fail if PLATFORMS env is empty:

```
$ make -f deployments/container/Makefile build PLATFORMS=''
Building image for multiple platforms with buildx: 
Note: multi-platform build does not load image into local Docker daemon.
make -f deployments/container/Makefile ensure-buildx-builder
make[1]: Entering directory '/home/ed/git/dra-example-driver'
dra-example-driver-builder
[+] Building 43.1s (1/1) FINISHED                                                                                                                                       
 => [internal] booting buildkit                                                                                                                                   43.1s
 => => pulling image moby/buildkit:buildx-stable-1                                                                                                                41.8s
 => => creating container buildx_buildkit_dra-example-driver-builder0                                                                                              1.3s
make[1]: Leaving directory '/home/ed/git/dra-example-driver'
docker buildx build \
        --pull --tag registry.example.com/dra-example-driver:latest --build-arg GO_VERSION="1.26.2" --build-arg BASE_IMAGE="docker.io/ubuntu:22.04" --build-arg VERSION="latest" -f /home/ed/git/dra-example-driver/deployments/container/Dockerfile /home/ed/git/dra-example-driver \
        --builder dra-example-driver-builder \
        --platform 
flag needs an argument: --platform
See 'docker buildx build --help'.
make: *** [deployments/container/Makefile:91: build] Error 125
```

I'd suggest to check this edge case and report an error before running docker.

demo/scripts/push-driver-image.sh
```diff
@@ -20,12 +20,20 @@ CURRENT_DIR="$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)"
...
 source "${CURRENT_DIR}/common.sh"
+check_demo_config || exit 1
+
+cd "${CURRENT_DIR}/../.."
 
 # Set build variables
 export REGISTRY="${DRIVER_IMAGE_REGISTRY}"
 export IMAGE="${DRIVER_IMAGE_NAME}"
 export VERSION="${DRIVER_IMAGE_TAG}"
+export PLATFORMS="${PLATFORMS}"
```
bart0sh: nit: redundand self-assessment:
```suggestion
export PLATFORMS
```

deployments/container/Makefile
```diff
@@ -17,27 +17,134 @@ CONTAINER_TOOL ?= docker
...
-		--build-arg GO_VERSION="$(GO_VERSION)" \
-		--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
-		--build-arg VERSION="$(VERSION)" \
-		-f $(DOCKERFILE) \
-		$(CURDIR)
+	$(MAKE) -f $(SELF_MAKEFILE) build
+
+build:
+# Multi-arch builds must use buildx (docker only).
+ifneq ($(words $(PLATFORM_LIST)),1)
```
bart0sh: nit:
```suggestion
COMMA := ,
ifneq ($(findstring $(COMMA),$(PLATFORMS)),)
```

PS: I suggested to change this code previously, but the result (using `words`  and extra variable PLATFORM_LIST turned out to be much less readable. Sorry for that.

bart0sh (changes requested):

deployments/container/Makefile
```diff
@@ -17,27 +17,134 @@ CONTAINER_TOOL ?= docker
...
 
 ##### Public rules #####
-.PHONY: $(DISTRIBUTIONS)
+.PHONY: $(DISTRIBUTIONS) build push ensure-buildx-builder
 
 ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
 
 # Use a generic build target to build the relevant images
 $(DISTRIBUTIONS):
-	DOCKER_BUILDKIT=1 \
```
bart0sh, line 33: I only suggested to remove DOCKER_BUILDKIT=1 from buildx builds. It makes sense to keep it for docker builds as it still makes sense for old docker versions.

demo/scripts/common.sh
prb112:
```suggestion
: ${PLATFORMS:="linux/amd64,linux/arm64,linux/ppc64le"}
```

deployments/container/Makefile
```diff
@@ -17,27 +17,134 @@ CONTAINER_TOOL ?= docker
...
+BASE_IMAGE ?= docker.io/ubuntu:22.04
+
+# Map host CPU to a Docker platform (`linux/amd64` on x86_64, `linux/arm64` on arm64).
+UNAME_M ?= $(shell uname -m)
+ifeq ($(UNAME_M),x86_64)
+	HOST_PLATFORM := linux/amd64
+else ifeq ($(UNAME_M),arm64)
+	HOST_PLATFORM := linux/arm64
+else ifeq ($(UNAME_M),aarch64)
+	HOST_PLATFORM := linux/arm64
```
prb112, line 41:
```suggestion
else ifeq ($(UNAME_M),aarch64)
	HOST_PLATFORM := linux/arm64
else ifeq ($(UNAME_M),ppc64le)
	HOST_PLATFORM := linux/ppc64le
```

README.md
```diff
@@ -48,6 +48,74 @@ From here we will build the image for the example resource driver:
...
+  ```bash
+  make -f deployments/container/Makefile build VERSION=<tag> IMAGE_NAME=<name|registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64'
+  ```
+- Push for current platform:
+  ```bash
+  make -f deployments/container/Makefile push VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=<docker|podman>
+  ```
+- Push for specific platform(s):
+  ```bash
+  make -f deployments/container/Makefile push VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64'
```
prb112, line 90:
```suggestion
  make -f deployments/container/Makefile push VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64,linux/ppc64le'
```

prb112:
As a user of this repo, and I support ppc64le, I suggested small changes to add ppc64le to your build (rather than exclude).

deployments/container/Makefile
vishalanarase: Done

```bash
❯ make -f deployments/container/Makefile build PLATFORMS=''
deployments/container/Makefile:54: *** PLATFORMS is set but empty. Unset it to use the host platform, or provide one or more values.  Stop.
```

demo/scripts/push-driver-image.sh
vishalanarase: Done

deployments/container/Makefile
vishalanarase: Done

deployments/container/Makefile
vishalanarase: Okay

README.md
vishalanarase: Done

vishalanarase:
> As a user of this repo, and I support ppc64le, I suggested small changes to add ppc64le to your build (rather than exclude).

@prb112 Suggested changes are included

deployments/container/Makefile
```diff
@@ -17,27 +17,142 @@ CONTAINER_TOOL ?= docker
...
+	$(MAKE) -f $(SELF_MAKEFILE) ensure-buildx-builder
+	$(CONTAINER_TOOL) buildx build \
+		$(IMAGE_BUILD_FLAGS) \
+		--builder $(BUILDX_BUILDER) \
+		$(BUILDX_METADATA_FLAGS) \
+		--platform $(PLATFORMS) \
+		--load
+endif
 
 push:
```
bart0sh, line 125: Would it make sense to make `push` target depend on `build` target? From the brief look it would allow to avoid some code duplication.

deployments/container/Makefile
```diff
@@ -17,27 +17,142 @@ CONTAINER_TOOL ?= docker
...
+		echo "Multi-platform push requires CONTAINER_TOOL=docker, got $(CONTAINER_TOOL)"; \
+		exit 1; \
+	fi
+	@echo "Pushing image for multiple platforms with buildx: $(PLATFORMS)"
+	$(MAKE) -f $(SELF_MAKEFILE) ensure-buildx-builder
+	$(CONTAINER_TOOL) buildx build \
+		$(IMAGE_BUILD_FLAGS) \
+		--builder $(BUILDX_BUILDER) \
+		--platform $(PLATFORMS) \
+		$(BUILDX_METADATA_FLAGS) \
```
bart0sh: This code is duplicated in the `build` target above.

deployments/container/Makefile
vishalanarase: Good point. I don't think multi-arch `push` should depend on `build`, because that would either duplicate the build or change build semantics (`--push` vs local/cache-only output). But the duplication itself is real. I’ll factor the common buildx flags/command shape into a shared variable/helper and keep `push/build` as separate entry points.

bart0sh: > But the duplication itself is real.

It is, especially for multi-platform pushes. I hope that an overhead is not significant as docker caches successfully built layers. Can you check this? If this is the case we can make `push` depend on `build`.

vishalanarase: I checked this locally and cache reuse does not make `push` after `build` cheap enough here. Re-running the same multi-platform buildx build still re-executed the expensive `COPY` . . / `RUN make PREFIX=/artifacts cmds` steps, so `push: build` would add noticeable extra work. Since multi-platform `build` and `push` also have different outputs (`cache-only` vs manifest export), I’d prefer to keep them separate and deduplicate the shared `buildx` flags/boilerplate instead. which is already addressed.

bart0sh: OK, thanks for checking this out!

deployments/container/Makefile
bart0sh: Okay means you're going to return it back or what?

demo/scripts/common.sh
```diff
@@ -58,10 +69,27 @@ if [[ -z "${CONTAINER_TOOL}" ]]; then
     elif [[ -n "$(which podman)" ]]; then
         echo "Podman found in PATH."
         CONTAINER_TOOL=podman
-    else
-        echo "No container tool detected. Please install Docker or Podman."
-        return 1
```
bart0sh, line 63: This check should be returned back, since we're not always checking return code when sourcing common.sh

cloudbuild.yaml
```diff
@@ -7,6 +7,7 @@ steps:
   entrypoint: make
   env:
   - DRIVER_IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver
+  - PLATFORMS=linux/amd64,linux/arm64
```
bart0sh: Should we unify PLATFORMS somehow?
```
$ git grep PLATFORMS |grep linux
README.md:- `PLATFORMS=<current host platform>` (for example `linux/amd64`, `linux/arm64`, or `linux/ppc64le`)
README.md:  make -f deployments/container/Makefile build VERSION=<tag> IMAGE_NAME=<name|registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64'
README.md:  make -f deployments/container/Makefile push VERSION=<tag> IMAGE_NAME=<registry/name> CONTAINER_TOOL=docker PLATFORMS='linux/amd64,linux/arm64'
README.md:  platform that does not match the host (for example `PLATFORMS=linux/arm64` on
cloudbuild.yaml:  - PLATFORMS=linux/amd64,linux/arm64
demo/scripts/push-driver-image.sh:: "${PLATFORMS:=linux/amd64,linux/arm64,linux/ppc64le}"
```

deployments/container/Makefile
```diff
@@ -17,27 +17,134 @@ CONTAINER_TOOL ?= docker
...
 $(DISTRIBUTIONS):
-	DOCKER_BUILDKIT=1 \
-		$(CONTAINER_TOOL) build --pull \
-		--tag $(IMAGE) \
-		--build-arg GO_VERSION="$(GO_VERSION)" \
-		--build-arg BASE_IMAGE="$(BASE_IMAGE)" \
-		--build-arg VERSION="$(VERSION)" \
-		-f $(DOCKERFILE) \
-		$(CURDIR)
+	$(MAKE) -f $(SELF_MAKEFILE) build
```
bart0sh: This change causes some variables (BASE_IMAGE and may be others) to be lost for sub-make, e.g. with the new distro target debian12:
```
##### Global variables #####
 CONTAINER_TOOL ?= docker
 MKDIR ?= mkdir
-DISTRIBUTIONS := ubuntu22.04
+DISTRIBUTIONS := ubuntu22.04 debian12
 DOCKERFILE := $(CURDIR)/deployments/container/Dockerfile
 SELF_MAKEFILE := $(lastword $(MAKEFILE_LIST))
 COMMA := ,
@@ -89,6 +89,7 @@ BUILDX_BUILD_FLAGS = \
 .PHONY: $(DISTRIBUTIONS) build push ensure-buildx-builder
 
 ubuntu22.04: BASE_IMAGE = docker.io/ubuntu:22.04
+debian12: BASE_IMAGE = docker.io/debian:12
 
 # Use a generic build target to build the relevant images
 $(DISTRIBUTIONS):
```

it will still build for ubuntu:22.04:

```
make -f deployments/container/Makefile debian12
make -f deployments/container/Makefile build
make[1]: Entering directory '/home/ed/git/dra-example-driver'
# Native single-arch build for the host platform.
Building image for single platform: linux/amd64
docker build --platform linux/amd64 --pull --tag registry.example.com/dra-example-driver:latest --build-arg GO_VERSION="1.26.2" --build-arg BASE_IMAGE="docker.io/ubuntu:22.04" --build-arg VERSION="latest" -f /home/ed/git/dra-example-driver/deployments/container/Dockerfile /home/ed/git/dra-example-driver
[+] Building 15.5s (9/13)                                                                                                                              docker:default
 => [internal] load build definition from Dockerfile                                                                                                             0.0s
 => => transferring dockerfile: 1.35kB                                                                                                                           0.0s
 => [internal] load metadata for docker.io/library/ubuntu:22.04               
...
```

deployments/container/Makefile
vishalanarase: I kept `DOCKER_BUILDKIT=1` only for the plain `docker build` (native single-arch) path, and it’s still not set for any `buildx build`/multi-platform invocations.

bart0sh:
lgtm, but considering amount of iteractions and changes, it would be good if other reviewers look at it.

vishalanarase:
> lgtm, but considering amount of iteractions and changes, it would be good if other reviewers look at it.

@pohly @nojnhuh Need lgtm from you as well, can you take a look please?

bart0sh (approved):

bart0sh:
@nojnhuh @pohly it's your last chance to look at this PR. I'll merge it tomorrow if there are no objections.

nojnhuh:
> @nojnhuh @pohly it's your last chance to look at this PR. I'll merge it tomorrow if there are no objections.

I don't think I'll have a chance to dig into this soon, so I'm happy to get this in now and iterate later if we need to.

@vishalanarase Could you please rebase this branch to make sure CI passes with the latest changes to the main branch?

vishalanarase:
> @vishalanarase Could you please rebase this branch to make sure CI passes with the latest changes to the main branch?

@nojnhuh Done

nojnhuh (commented):
/lgtm
based on @bart0sh's review.

@vishalanarase Could you please keep an eye on the [`post-dra-example-driver-image`](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-dra-example-driver-image) to make sure it passes the first time after this merges and the manifest pushed to the staging repo contains the expected platforms?

vishalanarase:
> @vishalanarase Could you please keep an eye on the [`post-dra-example-driver-image`](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-dra-example-driver-image) to make sure it passes the first time after this merges and the manifest pushed to the staging repo contains the expected platforms?

@nojnhuh Verified the pushed image and its manifest. The manifest list contains the expected platforms:

* linux/amd64
* linux/arm64
* linux/ppc64le

The `post-dra-example-driver-image` job has also passed successfully.

```bash
❯ docker pull us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/dra-example-driver:v0.3.0-23-gf87ea1d
v0.3.0-23-gf87ea1d: Pulling from k8s-staging-images/dra-example-driver/dra-example-driver
99267111abcc: Already exists
5fe12aab1e84: Pull complete
ab1b27b2ffe0: Pull complete
Digest: sha256:012da40739abd5493bbd6f032f624c9739a383450d8e66729d243b4e84e8f444
Status: Downloaded newer image for us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/dra-example-driver:v0.3.0-23-gf87ea1d
us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/dra-example-driver:v0.3.0-23-gf87ea1d

What's next:
    View a summary of image vulnerabilities and recommendations → docker scout quickview us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/dra-example-driver:v0.3.0-23-gf87ea1d
```

```bash
❯ docker manifest inspect \
  us-central1-docker.pkg.dev/k8s-staging-images/dra-example-driver/dra-example-driver:v0.3.0-23-gf87ea1d
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 902,
         "digest": "sha256:fe27253e60db75681dbf0813cc05e49ca9303d596ebd66d26d53febf5dd3262d",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 902,
         "digest": "sha256:0cc8c2f01beba3c25c2f5db02cf866a447e42f50b7cc32d7a576ab1934ff4fe9",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 902,
         "digest": "sha256:f09c3ce722fd2aa9d2ffb792780ac028963affaa8595fc8555b60c1359ea0687",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      }
   ]
}
```

Issue #94, opened 2025-04-09
Restructure the README file to allow separate instructions for creating clusters and installing driver


mortent:
Currently the README file goes through a single long section that includes building the image, creating a kind cluster, and instructions for running and validating the examples. If we want to add instructions for running the example driver on other Kubernetes distributions, we should restructure it so that we can have separate sections for what differs, but keep a single section for running and validating the examples.

I propose a structure like this:
```
- Intro
- Demo
  - High-level description of the demo
    - Prerequisites shared by all ways of running the demo (i.e. Helm and kubectl)
  - Creating a cluster and building/installing the example driver
    - Kind
    - GKE
    - Others...
  - Run through the examples with validation
  - Cleanup
    - Kind
    - GKE
    - Others...
- Anatomy of a DRA resource driver
- Code organization
- Best practices
```

An alternative would be to split this out into multiple files, but I think we can do that later if we find that the size of the README file becomes unwieldy.

nojnhuh:
/reopen
/lifecycle frozen

I think this is something we still want to do.

vishalanarase:
**Update: Scope Expansion & Refactoring Plan**

The scope has been expanded to introduce a more structured layout for cluster-related demos and scripts.

### Proposed Changes

* Create a new `demo/clusters/` directory to act as a central registry for all supported/demoed platforms.
* Add GKE support under:

  * `demo/clusters/gke/create-cluster.sh`
* Refactor existing KIND scripts into:

  * `demo/clusters/kind/`

#### Notes on Implementation

* Please keep the refactoring (moving existing KIND scripts into the new structure) as a **separate change** from the GKE additions.

nojnhuh:
#177 doesn't restructure the README, but we may want to consider doing that once we have scripts for more than kind.

/reopen

vishalanarase:
> [#177](https://github.com/kubernetes-sigs/dra-example-driver/pull/177) doesn't restructure the README, but we may want to consider doing that once we have scripts for more than kind.
> 
> /reopen

Definitely we have another PR for that I will work on that tomorrow #168

nojnhuh:
I think we have a decent pattern established here now.

/close

PR #168, opened 2026-04-23
docs: Add instructions for running the example driver on GKE  and optional ResourceQuota support

vishalanarase:
## Summary
This PR revives and updates the GKE documentation work from #93.

- Add a dedicated README section for installing `dra-example-driver` on GKE
- Add Helm chart support for optional `ResourceQuota` creation:
  - `resourcequota.enabled`
  - `resourcequota.pods`
  - `resourcequota.scopeSelector.matchExpressions`
- Add `templates/resourcequota.yaml` to render the ResourceQuota when enabled
- Incorporate prior review feedback

Fixes part of #94 
Supersedes: #93

## Why
GKE users need slightly different setup steps than kind, especially around running pods with critical priority classes.  
This PR documents the GKE flow and provides a chart toggle to create the required quota policy cleanly.

## Changes
- `README.md`
  - Added `Installing the example driver on a GKE cluster`
  - Added note about avoiding chart/image skew by using matching source revisions
- `deployments/helm/dra-example-driver/values.yaml`
  - Added `resourcequota` configuration block
- `deployments/helm/dra-example-driver/templates/resourcequota.yaml`
  - New template, conditional on `.Values.resourcequota.enabled`

## Test Plan

Validated on GKE cluster (`v1.35.2-gke.1485000`).

- Lint and render chart in both modes:
  - `helm lint deployments/helm/dra-example-driver`
  - `helm lint deployments/helm/dra-example-driver --set resourcequota.enabled=true`
  - `helm template ... --namespace dra-example-driver` (with and without `resourcequota.enabled=true`)
- Install with ResourceQuota enabled:
  - `helm upgrade -i --create-namespace --namespace dra-example-driver --set resourcequota.enabled=true dra-example-driver ./deployments/helm/dra-example-driver`
- Verify ResourceQuota exists and is active:
  - `kubectl get resourcequota -n dra-example-driver`
  - observed: `dra-example-driver-resourcequota` with pod quota usage (`pods: 1/10`)
- Apply demo workloads:
  - `kubectl apply --filename=demo/gpu-test{1,2,3,4,5}.yaml`
- Verify workloads and driver are running:
  - `kubectl get pods -A`
- Verify allocation semantics in logs:
  - checked GPU env vars from demo pods/containers
  - observed expected `TimeSlicing` / `SpacePartitioning` behavior
- Toggle ResourceQuota off and verify removal:
  - `helm upgrade -i --namespace dra-example-driver --set resourcequota.enabled=false dra-example-driver ./deployments/helm/dra-example-driver`
  - `kubectl get resourcequota -n dra-example-driver`
  - observed: `No resources found in dra-example-driver namespace.`

## Notes
- Scope is intentionally limited to docs + Helm chart support needed by the GKE install flow.
- Broader README restructuring discussed in #93/#94 can be handled separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,9 @@ The procedure below has been tested and verified on both Linux and Mac.
 * [helm v3.7.0+](https://helm.sh/docs/intro/install/)
 * [kubectl v1.18+](https://kubernetes.io/docs/reference/kubectl/)
 
-### Demo
+### Creating a cluster and installing the example driver
+
+#### Kind
 We start by first cloning this repository and `cd`ing into it. All of the
 scripts and example Pod specs used in this demo are contained here, so take a
 moment to browse through the various files and see what's available:
@@ -240,6 +242,12 @@ metadata:
   resourceVersion: ""
 ```
 
+#### Other platforms
+This demo uses kind by default. Additional platform-specific setup and cleanup
+guides are documented in [`demo/clusters`](demo/clusters/README.md), including
+GKE instructions in [`demo/clusters/gke`](demo/clusters/gke/README.md).
+
+### Run example workloads (shared across kind and GKE)
 Next, deploy four example apps that demonstrate how `ResourceClaim`s,
 `ResourceClaimTemplate`s, and custom `GpuConfig` objects can be used to
 select and configure resources in various ways:
@@ -391,7 +399,7 @@ To run this demo:
 
 This demonstration shows the end-to-end flow of the DRA AdminAccess feature. In a production environment, drivers could use this admin access indication to provide additional privileged capabilities or information to authorized workloads.
 
-### Clean Up
+### Cleanup
 
 Once you have verified everything is running correctly, delete all of the
 example apps:
@@ -420,11 +428,15 @@ admin-access                           pod0   1/1     Terminating   0          3
 ...
 ```
 
-Finally, you can run the following to cleanup your environment and delete the
-`kind` cluster started previously:
+#### Kind
+Finally, you can run the following to clean up your environment and delete the
+kind cluster started previously:
 ```bash
 ./demo/clusters/kind/delete-cluster.sh
 ```
+
+#### Other platforms
+Use the cleanup steps documented in [`demo/clusters`](demo/clusters/README.md).
 
 ## Device Profiles
 
--- a/demo/clusters/gke/README.md
+++ b/demo/clusters/gke/README.md
@@ -10,8 +10,13 @@ install/uninstall the driver.
 - [kubectl](https://kubernetes.io/docs/reference/kubectl/)
 - Authenticated GCP account and a selected project
 
-This flow uses the pre-built image for the kubelet plugin, so no local image
-build is required.
+This flow installs published release artifacts from `registry.k8s.io` and pins
+both the Helm chart and driver image to the same version to avoid chart/image
+skew.
+
+The steps below were validated with:
+- GKE: `v1.35.2-gke.1485000`
+- Driver/chart release: `0.2.0`
 
 To keep the walkthrough simple and close to the kind demo, use a single-node
 GKE cluster.
@@ -39,8 +44,11 @@ Supported environment variables:
 
 - `DRIVER_RELEASE_NAME` (default: `dra-example-driver`)
 - `DRIVER_NAMESPACE` (default: `dra-example-driver`)
+- `DRIVER_VERSION` (default: `0.2.0`)
+- `DRIVER_CHART_REF` (default: `oci://registry.k8s.io/dra-example-driver/charts/dra-example-driver`)
 
-The install script enables `resourceQuota.enabled=true` for GKE.
+The install script enables `resourceQuota.enabled=true` for GKE and sets
+`image.tag` to `DRIVER_VERSION` so chart/image versions remain aligned.
 
 ## Uninstall driver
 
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -424,6 +424,63 @@ Finally, you can run the following to cleanup your environment and delete the
 ./demo/delete-cluster.sh
 ```
 
+## Installing the example driver on a GKE cluster
```
nojnhuh: The README is already describing too many things in too much detail IMO. If you're already working on #94, can we try to solve that first? We could even have one PR that addresses both #94 and adds instructions for GKE in that new structure in separate sets of commits.

vishalanarase: Thanks, agreed. I restructured this PR per #94: separate Kind/GKE setup+cleanup, shared run/validation flow. I kept it in one README for now to keep this PR focused. We can do a follow-up docs split if needed.

vishalanarase:
@nojnhuh Can you please take a look, I have updated the structure as per issue description.

README.md
```diff
@@ -240,6 +242,51 @@ metadata:
   resourceVersion: ""
 ```
 
+#### GKE
```
nojnhuh: I would like to see if we can move these steps out of the README entirely. I think we should have docs and scripts somewhere else dedicated to describing how to create or delete a cluster for each platform/cloud/tool. Reproducing the steps inline here for `kind` makes sense so users don't have to go digging too deep just to get up and running somewhere, but ideally those should be one-line commands/scripts.

vishalanarase: Added GKE specific setup and cleanup separately via dedicated docs/scripts

demo/gke/create-cluster.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Kubernetes Authors.
```
nojnhuh: We can skip the year:

```suggestion
# Copyright The Kubernetes Authors.
```

demo/gke/create-cluster.sh
nojnhuh, line 1: Could we put this under a new `clusters` subdirectory, making this `demo/clusters/gke/create-cluster.sh`? We could move the existing kind scripts to `demo/clusters/kind`. Then `demo/clusters` serves as a registry of documented platforms.

If you could keep the refactoring separate from the GKE additions that would be great to have a clean example that others can follow.

demo/clusters/gke/README.md
nojnhuh, line 1: Can we move this to the same directory as the scripts and call it README.md?

README.md
nojnhuh: If we have all of the documented platforms in one place like `demo/clusters`, can we say something like "this demo uses kind, steps for other platforms are documented in `demo/clusters`" instead of naming each one here?

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -85,3 +85,14 @@ webhook:
     # The name of the service account to use.
     # If not set and create is true, a name is generated using the fullname template
     name: ""
+
+resourcequota:
```
nojnhuh: nit:

```suggestion
resourceQuota:
```

docs/GKE.md
```diff
@@ -0,0 +1,67 @@
...
+- Authenticated GCP account and a selected project
+
+This flow uses the pre-built image for the kubelet plugin, so no local image
+build is required.
+
+To keep the walkthrough simple and close to the kind demo, use a single-node
+GKE cluster.
+
+CDI must be enabled in containerd for the DRA driver to work. In GKE, CDI is
+enabled by default in recent versions, and the default script configuration
```
nojnhuh: I see this was copied over from #93. Is this still true? How "recent"? Can we call out specific versions instead?

docs/GKE.md
```diff
@@ -0,0 +1,67 @@
...
+uses the rapid release channel.
+
+Because DRA is still a beta feature, the required resource APIs must be
+explicitly enabled when creating the cluster. The create script handles that by
+default.
```
nojnhuh: DRA is GA. I would imagine a default installation doesn't require tweaking any settings in order to run the demo. If that's true, let's remove this.

docs/GKE.md
```diff
@@ -0,0 +1,67 @@
...
+
+Supported environment variables:
+
+- `DRIVER_RELEASE_NAME` (default: `dra-example-driver`)
+- `DRIVER_NAMESPACE` (default: `dra-example-driver`)
+
+The install script enables `resourcequota.enabled=true` for GKE.
+
+To avoid chart/image skew, prefer matching source revisions for chart and
+image when changing default image tags.
```
nojnhuh: Can we update the scripts to either use published release artifacts for both the chart and images or build them both from source? Right now the script installs the chart from source with the prebuilt images. Even if that works now, I don't want to assume that will hold true later.

demo/gke/create-cluster.sh
vishalanarase: Created new PR for kind work https://github.com/kubernetes-sigs/dra-example-driver/pull/177

vishalanarase:
@nojnhuh you can review https://github.com/kubernetes-sigs/dra-example-driver/pull/177 and merge first

vishalanarase:
@nojnhuh Can you please take a look?

nojnhuh (approved):
/lgtm
/approve

Thanks!

PR #170, opened 2026-04-24
Rename demo examples and add per-example documentation

### LINKED ISSUE #144: Restructure examples and demo to support more examples
The current structure of the examples and demo won't scale as we continue to add examples:

The opaque names `gpu-test1.yaml`, `gpu-test2.yaml`, etc. are hard to reason about without the [diagram](https://github.com/kubernetes-sigs/dra-example-driver/blob/main/demo/demo-apps.png?raw=true). The examples should have reasonably descriptive names that don't require some sort of separate index to figure out what they are. Supporting commentary for each example should live closer to the rest of that example and separate from other examples instead of all together in the README.

As we add examples, more mock devices are required to fit all of the examples at once. With the direction we're going, the driver will eventually advertise a couple dozen "GPUs" on each node by default, which isn't realistic and makes reading the ResourceSlices burdensome. Adding more nodes to the default kind cluster config isn't much better since the driver should run on any cluster. IMO the default number of devices should at least fit the largest example, but should not need to satisfy running every example all at once. We should update the demo to show a limited number of representative examples that can be installed with the default instead of all of the examples.

willie-yao:
Fixes #144 

This PR renames demos and keeps documentation for each demo in the yaml file itself. The main README is modified to run through the first 3 examples and a separate README under `/demo` gives an overview of the rest of the demos. Also the default number of GPUs was changed to 2 to more accurately reflect real world scenarios, with the trade-off of having to clean up after running each demo.


demo/admin-access.yaml
```diff
@@ -1,22 +1,43 @@
...
-# One pod with one container requesting all GPUs with admin access
-# This demo shows the DRA admin access feature with DRA_ADMIN_ACCESS environment variable
+# Example: DRA Admin Access
+#
+# One namespace with admin access label.
+# One pod with one container requesting all GPUs with admin access.
+# This demo shows the DRA admin access feature with DRA_ADMIN_ACCESS
+# environment variable.
+#
+# Key requirements:
```
nojnhuh, line 8: I'm wondering if it makes sense to organize these examples by "profile"? i.e. `demo/admin-access.yaml` -> `demo/gpu/admin-access.yaml`

As we add more profiles, that would help users know how to install the driver to run this particular example. On the other hand, more generic features like admin access could work with any profile, so it might become harder to find than in a flat list like this.

Let's not change that right now, but something to consider later.

nojnhuh:
@alaypatel07 Do the DRA scalability tests use any of these demo manifests? Or just the driver itself and the tests define their own Pods, ResourceClaims, etc.?

alaypatel07:
The tests define their own pods.

demo/admin-access.yaml
```diff
@@ -1,22 +1,43 @@
...
+#   - The namespace must have the label:
+#       resource.kubernetes.io/admin-access: "true"
+#   - The request must set adminAccess: true
+#   - allocationMode: All is used here to access all available GPUs
+#
+# Expected: The container has DRA_ADMIN_ACCESS=true and GPU_DEVICE env vars
+# for all available GPUs. Check with:
+#   kubectl logs -n admin-access pod0 -c ctr0 | grep DRA_ADMIN_ACCESS
+#   kubectl logs -n admin-access pod0 -c ctr0 | grep GPU_DEVICE
```
nojnhuh, line 19: Definitely not something we need to do now, but it would be awesome if we could make comments like this drive the e2e tests as well as serve as human-readable documentation. Then we wouldn't need to update the test logic at all when adding an example.

demo/admin-access.yaml
```diff
@@ -1,22 +1,43 @@
...
+#
+# Resources created:
+#   - 1 Namespace (with admin-access label)
+#   - 1 ResourceClaimTemplate (multiple-gpus-admin)
+#   - 1 Pod (pod0) with 1 container
+#
+# GPUs required: all available (uses allocationMode: All)
```
nojnhuh: Could we organize this in a new section titled something like "driver requirements" and also include the profile required? Right now only the "gpu" profile is implemented but we'll be adding more.

nojnhuh: Could we add a new section for cluster requirements? For example, Kubernetes version and feature gates. I'll go through and comment on each example. This one needs Kubernetes 1.34+ and the DRAAdminAccess feature.

demo/cel-selector.yaml
```diff
@@ -1,18 +1,30 @@
-# One pod, one container
-# Uses CEL expression selectors instead of matching on a DeviceClass
-# Selects a GPU with a specific model and at least 4Gi of memory
+# Example: CEL Expression Selectors
```
nojnhuh, line 1: This one requires Kubernetes 1.34+, no additional features.

demo/admin-access.yaml
nojnhuh:
```suggestion
# GPUs required: all available on a Node (uses allocationMode: All)
```

demo/admin-access.yaml
```diff
@@ -1,22 +1,43 @@
...
+# One namespace with admin access label.
+# One pod with one container requesting all GPUs with admin access.
+# This demo shows the DRA admin access feature with DRA_ADMIN_ACCESS
+# environment variable.
+#
+# Key requirements:
+#   - The namespace must have the label:
+#       resource.kubernetes.io/admin-access: "true"
+#   - The request must set adminAccess: true
+#   - allocationMode: All is used here to access all available GPUs
```
nojnhuh:
```suggestion
#   - "allocationMode: All" is used here to access all available GPUs on a Node. Admins typically require access to all devices on a node to perform maintenance or monitoring.
```

demo/admin-access.yaml
```diff
@@ -1,22 +1,43 @@
...
+#
+# Expected: The container has DRA_ADMIN_ACCESS=true and GPU_DEVICE env vars
+# for all available GPUs. Check with:
+#   kubectl logs -n admin-access pod0 -c ctr0 | grep DRA_ADMIN_ACCESS
+#   kubectl logs -n admin-access pod0 -c ctr0 | grep GPU_DEVICE
+#
+# Resources created:
+#   - 1 Namespace (with admin-access label)
+#   - 1 ResourceClaimTemplate (multiple-gpus-admin)
+#   - 1 Pod (pod0) with 1 container
```
nojnhuh: None of these demos are large enough to justify a shortlist like this IMO. I think the high-level description towards the top is sufficient.

demo/basic-resourceclaim-opaque-config.yaml
```diff
@@ -1,17 +1,36 @@
-# One pod, 1 container
-# Run as deployment with 1 replica
+# Example: GPU Sharing Strategies (TimeSlicing + SpacePartitioning)
```
nojnhuh, line 1: This one requires Kubernetes 1.34+, no additional features.

demo/initcontainer-shared-gpu.yaml
```diff
@@ -1,17 +1,30 @@
-# One pod: one initContainer + one container
-# Each asking for shared access to a single GPU
+# Example: InitContainer Sharing a GPU with a Container
```
nojnhuh, line 1: This one requires Kubernetes 1.34+, no additional features.

demo/basic-multiple-requests.yaml
```diff
@@ -1,17 +1,29 @@
-# One pod, one container
-# Asking for 2 distinct GPUs
+# Example: One Pod, Two GPUs
```
nojnhuh, line 1: This one requires Kubernetes 1.34+, no additional features.

demo/shared-global-claim.yaml
```diff
@@ -1,17 +1,33 @@
...
+# One shared, global claim providing access to a GPU.
+# Two pods, each asking for access to the shared GPU.
+#
+# Uses a ResourceClaim (not a ResourceClaimTemplate), which is a standalone
+# object that multiple pods can reference by name.
+#
+# Expected: Both pods see the same GPU with TimeSlicing. Check with:
+#   kubectl logs -n shared-global-claim pod0 -c ctr0 | grep GPU_DEVICE
+#   kubectl logs -n shared-global-claim pod1 -c ctr0 | grep GPU_DEVICE
+# Both pods should show the same GPU ID with SHARING_STRATEGY=TimeSlicing.
```
nojnhuh: We can skip mentioning anything about sharing strategy or timeslicing in these docs when the examples don't include a `GpuConfig`.

demo/shared-global-claim.yaml
```diff
@@ -1,17 +1,33 @@
-# One shared, global claim providing access to a GPU
-# Two pods, each asking for access to the shared GPU
+# Example: Shared Global ResourceClaim
```
nojnhuh: This one requires Kubernetes 1.34+, no additional features.

nojnhuh: Let's avoid saying "global" since ResourceClaims are namespace-scoped and not cluster-scoped.
```suggestion
# Example: Shared ResourceClaim Across Pods
```

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -37,7 +37,7 @@ serviceAccount:
 kubeletPlugin:
   # numDevices describes how many GPUs to advertise on each node when the "gpu"
   # deviceProfile is used. Not relevant for other profiles.
-  numDevices: 9
+  numDevices: 2
```
nojnhuh: The original default here is 8, could go back to that? I think that's a good balance between being somewhat realistic and enough to run a few examples without having to be super careful about cleaning things up.

demo/README.md
```diff
@@ -0,0 +1,51 @@
+# Demo Examples
+
+This directory contains example workloads that demonstrate different ways to
+request and configure GPU devices using Dynamic Resource Allocation (DRA).
+
+## Quick Start
+
+The following three examples are featured in the [main README walkthrough](../README.md)
```
nojnhuh: I think we can skip calling out which examples are already part of the top-level README.

demo/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+| Example | Description | GPUs |
+|---|---|---|
+| [two-pods-one-gpu-each.yaml](two-pods-one-gpu-each.yaml) | Two pods each get their own exclusive GPU | 2 |
+| [shared-gpu-across-containers.yaml](shared-gpu-across-containers.yaml) | Two containers in one pod share a single GPU | 1 |
+| [gpu-sharing-strategies.yaml](gpu-sharing-strategies.yaml) | TimeSlicing and SpacePartitioning on two GPUs | 2 |
+
+## All Examples
+
+| Example | Description | GPUs | Key Concept |
```
nojnhuh: I would love not to have to update this table when adding an example. I think we can remove this entirely and rely on having reasonably descriptive names for each example. That will probably never be perfect, but it should save people from looking at every single example to figure out which one does admin access, for example.

I do like how the "gpu-testN" names sort such that the basic examples are first and the more advanced ones are later. This table helps preserve that orderign which is nice. We could add `01-`, `02-`, etc. prefixes to the file names, but then it becomes a bit of a pain to manage merge conflicts or inserting into the middle.

Maybe a happy medium is to include some common prefix like `basic-` for the examples referenced in the top-level README which should be the first ones users look at to learn about DRA? That would then make those sort all together, be reasonably close to the top of the list, and doesn't make it any harder to name more examples as we add them (other than preferring to sort after "basic").

demo/README.md
```diff
@@ -0,0 +1,51 @@
+# Demo Examples
+
+This directory contains example workloads that demonstrate different ways to
+request and configure GPU devices using Dynamic Resource Allocation (DRA).
```
nojnhuh: Soon the driver will be able to handle more than GPUs:

```suggestion
request and configure devices using Dynamic Resource Allocation (DRA).
```

demo/test-admin-access.sh
nojnhuh, line 1: Unrelated to this PR: I forgot about this script when reviewing #153 and now it's no longer called. Could you check to make sure these are all tested and port anything missing as part of #171?

demo/basic-shared-claim-across-containers.yaml
```diff
@@ -1,17 +1,30 @@
-# One pod, two containers
-# Each asking for shared access to a single GPU
+# Example: Shared GPU Across Containers
```
nojnhuh, line 1: This one requires Kubernetes 1.34+, no additional features.

demo/basic-resourceclaimtemplate.yaml
```diff
@@ -1,17 +1,30 @@
-# Two pods, one container each
-# Each container asking for 1 distinct GPU
+# Example: Two Pods, One GPU Each
```
nojnhuh, line 1: This one requires Kubernetes 1.34+, no additional features.

README.md
```diff
@@ -104,322 +104,108 @@ dra-example-driver-webhook-7d465fbd5b-n2wxt           1/1     Running   0
...
-...
-gpu-test1   pod0   0/1     Pending             0          2s
-gpu-test1   pod1   0/1     Pending             0          2s
-gpu-test2   pod0   0/2     Pending             0          2s
-gpu-test3   pod0   0/1     ContainerCreating   0          2s
-gpu-test3   pod1   0/1     ContainerCreating   0          2s
-gpu-test4   pod0   0/1     Pending             0          2s
-gpu-test5   pod0   0/4     Pending             0          2s
-...
+Clean up before the next example:
```
nojnhuh: If we bump the default to 8 devices, then we wouldn't need to clean up in between each example. I've found it valuable before to have multiple examples running side-by-side at the same time to understand what's different between them. ResourceClaim status in particular is one of the more interesting parts that only exists at runtime and not in any of the manifests.

demo/admin-access.yaml
nojnhuh: I see you [added](https://github.com/kubernetes-sigs/dra-example-driver/commit/1b800b1af6ca7711c046c32efd91aad8f5d6b7ba#diff-56bfed9cbeeea2c86be284c1d4b444f76ba4db8d10f388a5027de2340710d1c2) then [removed](https://github.com/kubernetes-sigs/dra-example-driver/compare/1b800b1af6ca7711c046c32efd91aad8f5d6b7ba..db6ebebeda30e631850f50d00adbeda619dc7e9d) the profile requirement. Was that intentional?

If we still want to include that, we can fold this "GPUs required" note under the "Driver requirements" section.

demo/README.md
```diff
@@ -0,0 +1,32 @@
+# Demo Examples
+
+This directory contains example workloads that demonstrate different ways to
+request and configure devices using Dynamic Resource Allocation (DRA).
+
+Examples prefixed with `basic-` are featured in the
+[main README walkthrough](../README.md) and are a good starting point for
+learning about DRA.
```
nojnhuh: nit: we can let this be a one-way dependency:

```suggestion
Examples prefixed with `basic-` are a good starting point for
learning about DRA.
```

README.md
```diff
@@ -104,322 +104,85 @@ dra-example-driver-webhook-7d465fbd5b-n2wxt           1/1     Running   0
...
-        uuid:
-          string: gpu-657bd2e7-f5c2-a7f2-fbaa-0d1cdc32f81b
-      capacity:
-        memory:
-          value: 80Gi
-      name: gpu-7
-kind: List
-metadata:
-  resourceVersion: ""
+```bash
```
nojnhuh: Can we try to keep the README changes scoped to only renaming the examples? I'm happy to incorporate drive-by improvements, but let's at least keep those in separate commits.

demo/basic-resourceclaim-opaque-config.yaml
nojnhuh, line 1: Can we name this something like `basic-resourceclaim-opaque-config` instead? This is really demonstrating how to set config in a ResourceClaim. "GPU sharing" just happens to be what is being configured.

demo/basic-shared-claim-across-containers.yaml
nojnhuh, line 1: Likewise, can we rename this `basic-shared-claim-across-containers`? This shows a general pattern that works for any kind of device, not just GPUs.

demo/basic-resourceclaimtemplate.yaml
nojnhuh, line 1: Can we call this `basic-resourceclaimtemplate`?

demo/basic-multiple-requests.yaml
nojnhuh, line 1: I think we can also call this one `basic-` even if it's not part of the demo. Maybe something like `basic-multiple-requests`?

demo/admin-access.yaml
willie-yao: Whoops I thought it was redundant and forgot that you mentioned it in the comment! I'll add it back

README.md
nojnhuh: I think it's time to remove this since it uses the old names. It would be nice if each example had a README with its own Mermaid diagram or something. Not worth blocking this PR though IMO. I opened #176 to track that separately.

README.md
```diff
@@ -244,25 +244,29 @@ Next, deploy four example apps that demonstrate how `ResourceClaim`s,
 `ResourceClaimTemplate`s, and custom `GpuConfig` objects can be used to
 select and configure resources in various ways:
 ```bash
-kubectl apply --filename=demo/gpu-test{1,2,3,4,5}.yaml
+kubectl apply --filename=demo/basic-resourceclaimtemplate.yaml \
+  --filename=demo/basic-multiple-requests.yaml \
+  --filename=demo/basic-shared-claim-across-containers.yaml \
+  --filename=demo/shared-claim-across-pods.yaml \
```
nojnhuh: I think it makes sense to make this a `basic-` one too.

demo/test-admin-access.sh
nojnhuh: Actually I don't think this was ever part of the e2e script and is only referenced in the README, so nothing to do here for #171.

nojnhuh (commented):
I think this is good to go aside from renaming one more example.

nojnhuh (approved):
/lgtm
/approve

Thanks!

Issue #171, opened 2026-04-24
Improve e2e test isolation, error reporting, and extensibility


willie-yao:
The e2e tests were recently converted from bash scripts to Go/Ginkgo. They work well but carry over some design choices from the bash era that make them harder to debug, extend, and maintain. This tracks the overall cleanup effort.

Key areas:
- **Error messages**: Several `Eventually` calls swallow errors, making failures hard to debug (e.g. pod-not-found shows up as a generic timeout after 120s)
- **Test isolation**: All 8 test manifests are deployed together in `BeforeSuite` with a global `observedGPUs` map tracking state across tests. Tests can't run independently and order matters (test7 is deliberately deployed early)
- **Boilerplate**: The sharing tests (3, 4, 5, 6) repeat nearly identical verification logic with minor variations
- **Per-test driver config**: The driver is installed once with fixed Helm values. No test can use different values, and this makes future work like upgrade testing harder

Planned Improvements:

- [ ] Make each test self-contained (deploy/cleanup per test context)
- [ ] Reduce boilerplate with table-driven tests and shared helpers
- [ ] Support per-test Helm values (foundation for upgrade tests)

nojnhuh:
Feel free to combine any/all of those things into a single PR in separate commits.

willie-yao:
@nojnhuh I think for supporting per-test helm values, it would be better to have that in a separate PR since it adds a new feature for the tests whereas the other items in the PR are refactoring/cleanup. wdyt?

nojnhuh:
Landing those things in the same PR won't mess anything up necessarily, so feel free to combine however you want. In general, whatever makes for easiest branch maintenance for you is probably what's easiest to review too.

PR #175, opened 2026-04-27
Improve E2E test isolation, error reporting, and extensibility

willie-yao:
Fixes part of #171 
Refactors the e2e tests to make each test self-contained, improve failure diagnostics, and reduce boilerplate. Split into two commits for easier review.

## Commit 1: Make tests self-contained

- Each test deploys its own manifest via `BeforeEach`/`AfterEach` instead of deploying all 8 manifests at once in `BeforeSuite`
- `observedGPUs` is now per-test (verifies distinct GPUs within a test, not across tests)
- `BeforeSuite` only sets up K8s clients and verifies webhook readiness
- `AfterSuite` removed as cleanup is per-test
- `checkPodsReadyAndRunning` uses `Gomega` callback style with pod phase/conditions in errors
- Tests can now run in any order

## Commit 2: Reduce boilerplate and add failure diagnostics

- Flattens `Context`/`It` nesting to single `It` blocks with `deployManifest(ctx, namespace, file)` 
- `deployManifest` handles deploy, cleanup (with timeout), and failure diagnostics via `DeferCleanup` (LIFO ordering ensures diagnostics run before cleanup)
- Adds `verifySharedGPUGroup` helper — sharing tests (3, 4, 5, 6) go from ~20 lines of loop+branch each to a single call
- `getGPUsFromPodLogs` now takes a `Gomega` parameter so transient log-fetch errors retry within `Eventually` instead of immediately failing
- On failure, `kubectl describe pods`, `kubectl get events`, and driver logs are automatically dumped to test output
- Drops redundant `expectedPodCount` parameter from `checkPodsReadyAndRunning` (was always `len(pods)`)

### DIFF
--- a/test/e2e/e2e_setup_test.go
+++ b/test/e2e/e2e_setup_test.go
@@ -27,13 +27,14 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strings"
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
 	resourceapi "k8s.io/api/resource/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -50,6 +51,9 @@ var clientset *kubernetes.Clientset
 var dynamicClient dynamic.Interface
 var restMapper meta.RESTMapper
 
+const driverNamespace = "dra-example-driver"
+const driverPodSelector = "app.kubernetes.io/component=kubeletplugin"
+
 func init() {
 	currentDir, _ = os.Getwd()
 	rootDir = filepath.Join(filepath.Dir(currentDir), "..")
@@ -62,6 +66,11 @@ const (
 	checkPodLogsInterval = "1s"
 )
 
+var (
+	gpuDeviceRegexp = regexp.MustCompile(`(?m)^declare -x GPU_DEVICE_[0-9]+="(.+)"$`)
+	gpuIDRegexp     = regexp.MustCompile(`^gpu-([0-9]+)$`)
+)
+
 func TestE2e(t *testing.T) {
 	flag.Parse()
 	RegisterFailHandler(Fail)
@@ -69,11 +78,6 @@ func TestE2e(t *testing.T) {
 }
 
 var _ = BeforeSuite(func(ctx SpecContext) {
-	suiteConfig, _ := GinkgoConfiguration()
-	if suiteConfig.ParallelTotal > 1 {
-		Fail("tests cannot be run in parallel")
-	}
-
 	// Create a Kubernetes clientset
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	configOverrides := &clientcmd.ConfigOverrides{}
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_setup_test.go
```diff
@@ -149,6 +130,16 @@ func verifyWebhook(ctx context.Context) {
 	}, "30s", "1s").WithContext(ctx).Should(Succeed())
 }
 
+// deployManifest creates resources from a manifest file and returns a cleanup function.
+func deployManifest(ctx context.Context, manifestFile string) func() {
+	GinkgoHelper()
+	absPath := filepath.Join(demoManifestsDir, manifestFile)
+	createManifest(ctx, dynamicClient, absPath)
+	return func() {
+		deleteManifest(context.Background(), dynamicClient, absPath)
```
nojnhuh: Could we make this returned function take a `context.Context` as a parameter instead of using `context.Background()`? Then the caller can supply a more relevant context.

test/e2e/e2e_setup_test.go
nojnhuh: Are we able to remove this and run the tests in parallel now?

test/e2e/e2e_setup_test.go
```diff
@@ -92,6 +96,10 @@ var _ = BeforeSuite(func(ctx SpecContext) {
 	verifyWebhook(ctx)
 })
 
+// ---------------------------------------------------------------------------
+// Webhook verification
+// ---------------------------------------------------------------------------
```
nojnhuh: Can we remove these comments? IMO they don't help.

test/e2e/e2e_setup_test.go
```diff
@@ -149,6 +138,47 @@ func verifyWebhook(ctx context.Context) {
...
+		return
+	}
+	runKubectl("describe", "pods", "-n", namespace)
+	runKubectl("get", "events", "-n", namespace, "--sort-by=.lastTimestamp")
+	runKubectl("logs", "-n", driverNamespace, "-l", "app.kubernetes.io/component=kubeletplugin",
+		"--all-containers", "--tail=20")
+}
+
+func runKubectl(args ...string) {
+	cmd := exec.Command("kubectl", args...)
```
nojnhuh: We should always prefer client-go to exec-ing kubectl.

test/e2e/e2e_setup_test.go
nojnhuh, line 299: Can we update this to wait for all the objects to be gone?

test/e2e/e2e_setup_test.go
```diff
@@ -149,6 +138,47 @@ func verifyWebhook(ctx context.Context) {
...
+	GinkgoHelper()
+	absPath := filepath.Join(demoManifestsDir, manifestFile)
+	createManifest(ctx, dynamicClient, absPath)
+	// DeferCleanup is LIFO: register cleanup first, then diagnostics second.
+	// On teardown, diagnostics run first (while pods exist), then cleanup deletes them.
+	DeferCleanup(func() {
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		deleteManifest(cleanupCtx, dynamicClient, absPath)
+	})
```
nojnhuh: Can we make this a bit more Ginkgo-native?

```suggestion
	DeferCleanup(func(ctx context.Context) {
		deleteManifest(ctx, dynamicClient, absPath)
	}, NodeTimeout(30*time.Second))
```

test/e2e/e2e_setup_test.go
```diff
@@ -286,48 +316,55 @@ func createManifestWithDryRun(ctx context.Context, dynamicClient dynamic.Interfa
...
+// ---------------------------------------------------------------------------
+
+// getGPUsFromPodLogs retrieves pod logs and extracts GPU device information.
+// Returns errors via g so callers inside Eventually can retry on transient failures.
+func getGPUsFromPodLogs(g Gomega, namespace, pod, container string) ([]string, string) {
 	GinkgoHelper()
 	req := clientset.CoreV1().Pods(namespace).GetLogs(pod, &v1.PodLogOptions{
 		Container: container,
 	})
 	podLogs, err := req.Stream(context.TODO())
```
nojnhuh: As you're going through things, feel free to clean up the context handling, even if just to bubble up the `context.TODO` up a level.

test/e2e/e2e_setup_test.go
```diff
@@ -447,14 +474,47 @@ func verifyGPUProperties(g Gomega, logs, namespace, podName, containerName strin
...
+	container string
+}
+
+// sharingGroup describes a set of pod/containers that should all share the same GPU,
+// along with the expected sharing properties.
+type sharingGroup struct {
+	members           []podContainer
+	expectedStrategy  string
+	expectedProperty  string
+	expectedPropValue string
```
nojnhuh: Can you add comments describing what these fields mean?

test/e2e/e2e_setup_test.go
willie-yao: I think so, but the parallel tests would compete for the same GPU pool. That might be a problem when we continue to add more tests. We can maybe bump the numDevices relative to the number of tests?

nojnhuh: I tried hacking around this and the tests still seem to work. Some Pods just end up waiting for devices allocated to other tests. Too much contention will start to hit our "pods running and ready" timeout but I don't see us running into that soon. Longer-term I'd like to see the driver itself also get deployed per-test where we wouldn't have to worry about that.

test/e2e/e2e_setup_test.go
willie-yao: Diff of changes [here](https://github.com/kubernetes-sigs/dra-example-driver/pull/175/changes/9a8c65f4569a4e28a3f2f98728ce395dbc69f87f#diff-cc839ec4e1a1c750e2f610cae70c6cb30ee66eafc9dd751179551ed16c31cad4R173-R218). It did result in a lot more code so let me know if this is what you're looking for. I used client-go to get the pod status, events, and driver logs.

test/e2e/e2e_setup_test.go
```diff
@@ -158,33 +155,68 @@ func deployManifest(ctx context.Context, namespace string, manifestFile string)
...
+		return
+	}
+	for _, pod := range driverPods.Items {
+		for _, c := range pod.Spec.Containers {
+			stream, err := clientset.CoreV1().Pods(driverNamespace).GetLogs(pod.Name, &v1.PodLogOptions{
+				Container: c.Name,
+				TailLines: &tailLines,
+			}).Stream(ctx)
+			if err != nil {
+				fmt.Fprintf(GinkgoWriter, "Driver %s/%s: failed to get logs: %v\n", pod.Name, c.Name, err)
```
nojnhuh: Can we format this to be a bit more explicit about exactly what `Driver %s/%s"`refers to? Something like "Driver Pod %s, container %s"? Here and below.

test/e2e/e2e_setup_test.go
nojnhuh: You could argue it's less code if you count all the lines in kubectl :)

test/e2e/e2e_setup_test.go
nojnhuh: Can we try removing this and adding `-p` here? https://github.com/kubernetes-sigs/dra-example-driver/blob/9a8c65f4569a4e28a3f2f98728ce395dbc69f87f/Makefile#L120

nojnhuh:
@willie-yao Could you please rebase again to make sure we test with the most recent changes?

nojnhuh (commented):
/lgtm

@willie-yao Could you please clean up the git history to squash the follow-up commits, keeping the other logical sets of changes in separate commits? I'll approve once you've pushed those changes.

willie-yao:
@nojnhuh Done! Squashed into the original two commits

nojnhuh:
/approve

Thanks!

willie-yao:
`error: unable to contact snap store`
I don't think this is related to my changes, maybe an outage of some sort?

nojnhuh:
> `error: unable to contact snap store` I don't think this is related to my changes, maybe an outage of some sort?

Indeed

https://status.canonical.com/
> Canonical’s web infrastructure is under a sustained, cross-border attack and we are working to address it. We will provide more information in our official channels as soon as we are able to.

PR #177, opened 2026-04-29
refactor: move kind cluster scripts under demo/clusters/kind

### LINKED ISSUE #94: Restructure the README file to allow separate instructions for creating clusters and installing driver

vishalanarase:
## Summary
- move kind cluster lifecycle scripts to `demo/clusters/kind/`
- keep existing entrypoints `demo/create-cluster.sh` and `demo/delete-cluster.sh` as thin wrappers for backward compatibility
- add `demo/clusters/README.md` as the platform registry entrypoint

Fixes #94 

## Why
This is a refactor-only change to establish `demo/clusters` as the canonical location for platform-specific cluster lifecycle scripts, while preserving existing user-facing commands in the top-level demo flow.

## Scope
This PR only includes the kind script refactor.

## Test plan
- [x] `bash -n demo/create-cluster.sh demo/delete-cluster.sh`
- [x] `bash -n demo/clusters/kind/create-cluster.sh demo/clusters/kind/delete-cluster.sh`
- [x] verified top-level wrappers delegate to `demo/clusters/kind/*`


demo/create-cluster.sh
nojnhuh, line 1: Can we move these files altogether and update references in the demo to refer to the scripts in `demo/clusters/kind` directly?

demo/clusters/README.md
```diff
@@ -0,0 +1,11 @@
+# Cluster setup guides
+
+The demo uses kind by default.
+
+- kind scripts: `demo/clusters/kind/`
```
nojnhuh: We don't need to list each "flavor" here individually IMO. All I think this README needs to say is that this directory contains scripts/docs to bring up DRA-enabled clusters on various platforms. The names of the subdirectories should make clear what those platforms are.

If we think each one will have the same general structure (create-cluster.sh, delete-cluster.sh, etc.), then we should document that here too.

vishalanarase: Done

nojnhuh (approved):
/lgtm
/approve

Thanks!

PR #180, opened 2026-04-29
Add multi-arch container image support (amd64/arm64)

Basavaraju-G:
## Summary

Add multi-architecture container image build support for `linux/amd64` and `linux/arm64` using Docker buildx. This is needed for downstream consumers like [Kueue CI](https://github.com/openshift/release/pull/78156) that run on multi-arch clusters.

### Approach

The Dockerfile's build stage uses `--platform=$BUILDPLATFORM` so Go always compiles **natively** on the host architecture (no QEMU emulation needed). Cross-compilation is handled by passing `GOARCH=$TARGETARCH` to the Go build. Since the final stage has no `RUN` commands (only `COPY` and `LABEL`), no emulation is required for any target platform.

### Changes

- **`deployments/container/Dockerfile`**: Add `--platform=$BUILDPLATFORM` to the build stage and `TARGETARCH`/`TARGETOS` args for cross-compilation
- **`Makefile`**: Add `GOARCH` variable and pass it to all `go build` commands
- **`deployments/container/Makefile`**: Add `PLATFORMS` variable and `push-multiarch` targets using `docker buildx build --push --platform`
- **`demo/scripts/build-driver-image.sh`**: Skip local build when `MULTI_ARCH=true` (buildx combines build+push)
- **`demo/scripts/push-driver-image.sh`**: Call `*-push-multiarch` target when `MULTI_ARCH=true`
- **`cloudbuild.yaml`**: Set up buildx builder and enable `MULTI_ARCH=true` for release artifacts
- **`.github/workflows/multiarch.yaml`**: New CI workflow to verify multi-arch builds on every push/PR

### Backward compatibility

- Existing single-arch build targets are **unchanged**; `make -f deployments/container/Makefile ubuntu22.04` still works as before
- Multi-arch is opt-in via `MULTI_ARCH=true` env var (set automatically in Cloud Build)
- `PLATFORMS` defaults to `linux/amd64,linux/arm64` and can be overridden for additional architectures (e.g., `ppc64le`, `s390x`)

### Testing

- Verified Go cross-compilation produces correct ELF binaries for both architectures:
  - `x86-64` (amd64): `ELF 64-bit LSB executable, x86-64, statically linked`
  - `aarch64` (arm64): `ELF 64-bit LSB executable, ARM aarch64, statically linked`
- Verified Docker build with updated Dockerfile succeeds and correctly applies `TARGETARCH`/`TARGETOS`
- CI workflow added to validate multi-arch builds on PRs

/cc @klueska @pohly @byako

### DIFF
--- a/.github/workflows/multiarch.yaml
+++ b/.github/workflows/multiarch.yaml
@@ -1,0 +1,24 @@
+name: Build multi-arch images
+
+on: [ push, pull_request ]
+
+permissions:
+  contents: read
+
+jobs:
+  build-multiarch:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+    - name: Build multi-arch images
+      run: |
+        docker buildx build --pull \
+          --platform linux/amd64,linux/arm64 \
+          --build-arg GOLANG_VERSION="1.25.5" \
+          --build-arg BASE_IMAGE="docker.io/ubuntu:22.04" \
+          --build-arg VERSION="ci-test" \
+          -f deployments/container/Dockerfile \
+          .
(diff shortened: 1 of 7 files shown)


nojnhuh:
This is already in progress in #167. Please offer any feedback there. Thanks!

/close

PR #188, opened 2026-04-29
Bump Kubernetes dependencies and test environment to 1.36

nojnhuh:

PR #189, opened 2026-04-30
Fix Go version variable collision with Docker build

nojnhuh:
#188 broke the postsubmit image push job when it bumped Go to 1.26: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-dra-example-driver-image/2049630160541454336

Upstream Go Docker images (and images based on them like the gcb-docker-gcloud image in cloudbuild.yaml) set the `GOLANG_VERSION` variable to reflect the version of the local toolchain installed in the image. Separately (AFAICT), the scripts here use the `GOLANG_VERSION` variable as input to select a build image. Since cloudbuild.yaml defines an image using Go 1.25 (and `GOLANG_VERSION=1.25.x`), `golang:1.25.x` base images which defined `GOTOOLCHAIN=local` are used which fail to operate against our module which specifies Go 1.26.0.

Renaming the variable our scripts use to `GO_VERSION` avoids this conflict and allows the version defined in common.mk to take precedence over the value defined by the build image without any potential side-effects from the image's `GOLANG_VERSION` drifting from its actual toolchain version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ export CHART_GIT_TAG ?= $(shell git describe --tags --always --dirty --match 'ch
 
 include $(CURDIR)/common.mk
 
-BUILDIMAGE_TAG ?= golang$(GOLANG_VERSION)
+BUILDIMAGE_TAG ?= golang$(GO_VERSION)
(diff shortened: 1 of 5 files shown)


willie-yao (approved):
/lgtm

PR #193, opened 2026-05-02
Add DRAWorkloadResourceClaims example

### LINKED ISSUE #183: Add example for ResourceClaim support for workloads (KEP-5729)
KEP-5729 (https://github.com/kubernetes/enhancements/issues/5729) adds fields to the Workload and PodGroup APIs to reference ResourceClaims and ResourceClaimTemplates. A ResourceClaim referenced by a Workload is reserved for the Workload rather than individual Pods, removing the 256-entry limit on `status.reservedFor`. This reached Alpha in 1.36.

The example driver should include a demo showing how a Workload object can define claims for DRA resources that are shared among its component Pods.

nojnhuh:
This PR adds a new example and e2e test utilizing the DRAWorkloadResourceClaims feature.

Fixes #183

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -19,11 +19,15 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"path/filepath"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	gpuv1alpha1 "sigs.k8s.io/dra-example-driver/api/example.com/resource/gpu/v1alpha1"
 )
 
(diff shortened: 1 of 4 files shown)


test/e2e/setup-e2e.sh
```diff
@@ -41,6 +41,6 @@ helm upgrade -i \
   --create-namespace \
   --namespace dra-example-driver \
   --set webhook.enabled=true \
-  --set kubeletPlugin.numDevices=10 \
+  --set kubeletPlugin.numDevices=12 \
```
willie-yao, line 44: Is this being increased due to parallelization and will this have to be done for each additional test case?

test/e2e/e2e_test.go
willie-yao, line 1: Right now it seems like the only assertions the test is making are that pods in a group get the same GPU and each group gets a different GPU. Is that enough to verify DRAWorkloadResourceClaims is doing what it's supposed to? Since the driver applies Timeslicing/Default by default, I feel like this could be missing some extra assertions like checking if there is exactly one ResourceClaim per PodGroup, but I could be misunderstanding.

test/e2e/setup-e2e.sh
nojnhuh: As we test more scenarios, we need to bump the number of devices (either like this or by adding nodes), increase timeouts to account for waiting for devices used by another test running in parallel, or run fewer tests in parallel. Probably not strictly necessary right now, but I'd rather get into a habit of bumping this to be safe than letting it go long enough for the tests to start flaking.

As part of #171, I want the tests to install a separate driver instance for each individual test to ensure that each test uses dedicated devices.

test/e2e/e2e_test.go
nojnhuh: IMO I think this is sufficient to make sure we don't accidentally make any meaningful changes to the example. This doesn't check that every pod in each group only gets one GPU, but all the pods from each group are built from the same template. If a pod did end up with more than one GPU, we would at least see this test flake. I've run this enough locally though that I'm not worried about that being an issue.

nojnhuh: Let's keep looking for ways to make assertions in tests more flexible to accommodate different kinds of scenarios like this though.

willie-yao (approved):
/lgtm

PR #196, opened 2026-05-04
DRAExtendedResource: add example for KEP-5004 extended resource requests

### LINKED ISSUE #181: Add example for extended resource requests via DRA driver (KEP-5004)
KEP-5004 (https://github.com/kubernetes/enhancements/issues/5004) allows DRA drivers to honor requests made via the extended resource API (e.g. `nvidia.com/gpu: 2`) rather than requiring a standard device plugin. This reached Beta in 1.36.

The example driver should demonstrate how a DRA driver can be configured to handle extended resource requests, showing how pods can request devices using the traditional `resources.limits` syntax and have those requests fulfilled by the DRA driver.

This would help driver authors understand how to implement extended resource support and give users a working example to validate the feature.

willie-yao:
This PR adds a new example and e2e test utilizing the DRAExtendedResource feature.

Fixes #181

### DIFF
--- a/demo/extended-resource-request.yaml
+++ b/demo/extended-resource-request.yaml
@@ -1,0 +1,72 @@
+# Example: Extended Resource Requests via DRA (KEP-5004)
+#
+# Two pods, one container each.
+# Each container asks for one GPU using the classic `resources.limits` API
+# instead of via a ResourceClaim. The kube-scheduler dynamically allocates a
+# device from the DRA driver to satisfy each request.
+#
+# pod0 uses the implicit name `deviceclass.resource.kubernetes.io/<className>`
+# and works against the default chart install.
+# pod1 uses an explicit name (`example.com/gpu`) and requires the chart to be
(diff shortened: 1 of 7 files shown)


nojnhuh:
GitHub is failing to let me add comments inline. Two things I saw so far:

Is the demo/test-extended-resource.sh script doing anything substantially different from the test or the comment in the example? If we can avoid having to keep an extra script in sync with the example and its e2e test, I think we can remove the script. I know we have a separate script like this for admin access, but now the comments in each example have enough that a user should be able to follow that. That should be the goal at least, IMO.

The example lists Kubernetes 1.36+ as a requirement. This feature was added as alpha in 1.34. Is the example relying on significant changes made since then that require 1.36? Can you test with Kubernetes 1.34 and see if that works?

willie-yao:
@nojnhuh Thanks for the review! I removed the script and verified it works on 1.34

nojnhuh (approved):
/lgtm
/approve

PR #198, opened 2026-05-12
build: support proxy build args for image builds

ttsuuubasa:
## Summary
Support proxy build args for Docker image builds, enabling builds behind corporate/network proxies.

## What changed
- **Makefile**:
  Conditionally assemble `--build-arg` flags for `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` when the corresponding environment variables are set, and pass them to `docker build`.
- **deployments/container/Makefile**:
   Add the same conditional proxy build-arg logic for distribution image builds.
- **deployments/container/Dockerfile**:
  Declare `ARG` and `ENV` for `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` in the build stage.
- **docker/Dockerfile.devel**:
  Add the same `ARG`/`ENV` proxy handling for the development image.

## Why
Image builds fail in environments that require an proxy to reach external registries or package repositories. This change propagates proxy settings into the Docker build context so that package downloads and dependency fetches succeed behind a proxy.

## Behavior
- **Proxy env vars set**: proxy settings are forwarded into the build via `--build-arg` and applied as `ENV` inside the build stage.
- **Proxy env vars unset**: no additional build args are passed; behavior is unchanged from before.

## Impact
- Build-time only. No changes to runtime logic, APIs, or deployed artifacts.

### DIFF
--- a/deployments/container/Dockerfile
+++ b/deployments/container/Dockerfile
@@ -16,6 +16,16 @@ ARG GOLANG_VERSION=undefined
 ARG BASE_IMAGE=undefined
 FROM golang:${GOLANG_VERSION} AS build
 
+# Proxy arguments (optional)
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+ARG NO_PROXY
+
+# Set proxy environment variables only if they are provided as build args
+ENV HTTP_PROXY=${HTTP_PROXY}
+ENV HTTPS_PROXY=${HTTPS_PROXY}
+ENV NO_PROXY=${NO_PROXY}
+
 WORKDIR /build
 COPY . .
 
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,18 @@ include $(CURDIR)/common.mk
 BUILDIMAGE_TAG ?= golang$(GOLANG_VERSION)
 BUILDIMAGE ?= $(IMAGE_NAME)-build:$(BUILDIMAGE_TAG)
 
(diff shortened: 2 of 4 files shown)


deployments/container/Dockerfile
```diff
@@ -16,6 +16,16 @@ ARG GOLANG_VERSION=undefined
...
 
+# Proxy arguments (optional)
+ARG HTTP_PROXY
+ARG HTTPS_PROXY
+ARG NO_PROXY
+
+# Set proxy environment variables only if they are provided as build args
+ENV HTTP_PROXY=${HTTP_PROXY}
+ENV HTTPS_PROXY=${HTTPS_PROXY}
+ENV NO_PROXY=${NO_PROXY}
```
nojnhuh, line 27: IIUC docker already sets these automatically for builds: https://docs.docker.com/engine/cli/proxy/#build-with-a-proxy-configuration

What error are you running into without these changes?

ttsuuubasa: Thank you for your suggestion.
As you mentioned, configuring the proxy via `~/.docker/config.json` worked successfully, and I was able to build images without any issues.
Therefore, it would be fine to close the PR related to the proxy configuration.

nojnhuh: Thanks for checking!

/close

Issue #71, opened 2024-12-10
Advertise devices shared between nodes


nojnhuh:
While the driver currently advertises Node-local "GPU" devices, it would be great also to show what devices shared between Nodes look like and how to claim those. e.g. NVIDIA IMEX channels: https://github.com/NVIDIA/k8s-dra-driver/blob/32805fec62d6b3269e75ddb0eddeaa5c4d214564/demo/specs/quickstart/imex-test1.yaml

PR #199, opened 2026-05-12
feat: Add DRA Device Binding Conditions (KEP-5007)

ttsuuubasa:
## Summary

Add configurable Binding Conditions support to the DRA example driver, allowing the scheduler to defer final device binding decisions until the driver confirms device readiness on the node.

## Motivation

The DRA example driver serves as a reference implementation for developers building their own DRA drivers. By demonstrating how to implement Binding Conditions — including the required ResourceSlice fields and end-to-end demo scripts — this PR helps driver developers understand the expected behavior and integration patterns for this feature introduced in Kubernetes 1.34.

### Kubelet Plugin
- **Feature flag**: Add `--binding-conditions` CLI flag (env: `BINDING_CONDITIONS`, default: `false`).
- **GPU Profile**: When enabled, `EnumerateDevices()` sets `BindingConditions` and `BindingFailureConditions` on each device in the ResourceSlice.
- **Helm chart**: Add `kubeletPlugin.bindingConditions: false` to _values.yaml_ and wire it to the kubelet plugin container env in _kubeletplugin.yaml_.
- **Demo manifest**: _demo/binding-conditions/binding-conditions.yaml_ — sample manifest demonstrating the workflow.
- **Kind cluster config**: Enable `DRADeviceBindingConditions` feature gate.

### Controller & Tests
- **dra-example-controller**: New controller binary using controller-runtime with a plugin architecture.
  - _plugins/bindingconditions.go_: Watches allocated ResourceClaim objects and automatically satisfies binding conditions by setting the device condition to True.
  - _controller.go_: Generic `ClaimReconciler` that dispatches to registered plugins. Plugins run in parallel using goroutines.
  - _main.go_: Plugin registry, flag parsing (`--enable-plugin`, `--driver-name`), manager setup.
  - _controller_test.go_: Unit tests for `ClaimReconciler` covering successful plugin execution, plugin failure aggregation, and irrelevant claim filtering.
- **Helm templates**: _controller-deployment.yaml_ (only rendered when `controller.plugins` is non-empty), updated _clusterrole.yaml_ with RBAC for ResourceClaims, new `controller.plugins: []` field in _values.yaml_.
- **Dockerfile**: Add `dra-example-controller` binary to the container image.
- **E2E tests**: The `BindingConditions` test context installs the driver with `kubeletPlugin.bindingConditions=true` and `controller.plugins={BindingConditions}`; verifies ResourceSlice fields and pod readiness lifecycle.
- **Demo documentation**: _demo/binding-conditions/README.md_ with step-by-step walkthrough.

## Testing

```bash
# Run all E2E tests (includes BindingConditions context)
make setup-e2e
make test-e2e

# Or run only binding conditions tests
go run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -p --focus="BindingConditions" ./test/e2e/...

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -391,6 +391,120 @@ To run this demo:
...
+            driver: gpu.example.com
+            pool: dra-example-driver-cluster-worker
+            request: gpu
+      reservedFor:
+      - name: pod0
+        resource: pods
+    ...
+    ```
+
+4. **Set the binding condition**: In a real driver, an external controller or other component would update the `ResourceClaim` status to signal that the device is ready. In this demo, simulate that by editing the status directly. First get the current timestamp in UTC:
```
nojnhuh: The example driver should be applying the condition itself instead of relying on users to do this. This better reflects how the feature is implemented and used with real drivers. The example driver doesn't have to wait for anything interesting before applying the condition, but we should include a comment saying where a real driver _would_ wait for some real signal before adding the condition.

README.md
```diff
@@ -391,6 +391,120 @@ To run this demo:
 
 This demonstration shows the end-to-end flow of the DRA AdminAccess feature. In a production environment, drivers could use this admin access indication to provide additional privileged capabilities or information to authorized workloads.
 
+### Demo DRA Device Binding Conditions Feature
```
nojnhuh: Can we start putting feature-specific docs in their own directory? I think this would fit somewhere like `demo/binding-conditions/README.md`.

demo/test-binding-conditions.sh
nojnhuh, line 1: Instead of a script like this, could we add validation to the e2e tests? If we update the driver to add the condition, I think all an e2e test would need to check is that the binding conditions are added to the devices in the ResourceSlices and that the Pods still become Ready and Running. We can assume that Kubernetes is behaving correctly otherwise.

README.md
ttsuuubasa: Given the nature of the BindingConditions feature, I believe that we need to implement a controller on the example driver side to handle this. Is my understanding correct?

README.md
ttsuuubasa: I’m planning to create a new directory and move the content currently documented in the README there.

demo/test-binding-conditions.sh
ttsuuubasa: I’m trying to implement this in the e2e tests instead of using the current script.

README.md
nojnhuh: Yes, the example driver doesn't run any controller currently. Ideally we set it up to be fairly generic to accommodate other features too, e.g. #71.

ttsuuubasa: @nojnhuh 
I have added an implementation of a controller for setting BindingConditions. To ensure generality, each piece of logic is designed to be implemented as a plugin, allowing the corresponding processing to be executed by implementing its own plugin.

README.md
ttsuuubasa: @nojnhuh 
The feature-specific docs for the Binding Conditions demo have been moved under `demo/binding-conditions`.

demo/test-binding-conditions.sh
ttsuuubasa: I have added an e2e test that verifies whether binding conditions are set on the ResourceSlice and that the Pod reaches the Running state after the controller sets the binding conditions.

cmd/dra-example-controller/controller.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+func (r *ClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&resourceapi.ResourceClaim{}).
+		Complete(r)
+}
+
+func (r *ClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
nojnhuh, line 62: Can we add some basic unit tests to this package?

demo/binding-conditions/binding-conditions.yaml
```diff
@@ -0,0 +1,83 @@
...
+#   bindingFailureConditions:
+#   - BindingFailureConditions
+#
+# The scheduler uses these fields to track the binding lifecycle of allocated
+# devices, deferring final binding decisions until the driver confirms device
+# readiness on the node.
+#
+# Key requirements:
+#   - The driver must be installed with bindingConditions=true:
+#       helm upgrade -i ... --set kubeletPlugin.bindingConditions=true
```
nojnhuh: Does `controller.plugins={BindingConditions}` also need to be set?

demo/binding-conditions/README.md
```diff
@@ -0,0 +1,116 @@
...
+    Each device should contain:
+    ```yaml
+    bindsToNode: true
+    bindingConditions:
+    - BindingConditions
+    bindingFailureConditions:
+    - BindingFailureConditions
+    ```
+
+3. **Create the demo pod**: Apply the example manifest and confirm that the pod stays `Pending` because the binding conditions have not yet been satisfied:
```
nojnhuh: Could you update this to account for the controller automatically adding the binding condition?

deployments/helm/dra-example-driver/templates/clusterrole.yaml
```diff
@@ -7,7 +7,14 @@ metadata:
...
   resources: ["resourceclaims"]
-  verbs: ["get"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["resource.k8s.io"]
+  resources: ["resourceclaims/status"]
+  verbs: ["get", "update", "patch"]
+- apiGroups: ["resource.k8s.io"]
+  resources: ["resourceclaims/driver"]
+  resourceNames: [{{ include "dra-example-driver.driverName" . | quote }}]
+  verbs: ["associated-node:update", "arbitrary-node:update"]
```
nojnhuh: Do we need `arbitrary-node:update` if binding conditions only operate on node-local ResourceSlices?

deployments/helm/dra-example-driver/values.yaml
```diff
@@ -67,6 +67,8 @@ kubeletPlugin:
   affinity: {}
   kubeletRegistrarDirectoryPath: /var/lib/kubelet/plugins_registry
   kubeletPluginsDirectoryPath: /var/lib/kubelet/plugins
+  # bindingConditions enables or disables binding conditions processing in the DRA driver
+  bindingConditions: false
```
nojnhuh, line 75: Could we use a single value to control both the kubelet plugin and the controller?
```yaml
bindingConditions:
  enabled: false
```

internal/profiles/profiles.go
```diff
@@ -25,6 +25,11 @@ import (
 	cdiapi "tags.cncf.io/container-device-interface/pkg/cdi"
 )
 
+const (
+	BindingConditions        = "BindingConditions"
+	BindingFailureConditions = "BindingFailureConditions"
+)
```
nojnhuh: Since these are only added for the GPU profile, can we define them in that specific package?

test/e2e/setup-e2e.sh
```diff
@@ -43,5 +50,6 @@ helm upgrade -i \
   --set webhook.enabled=true \
   --set kubeletPlugin.numDevices=14 \
   --set deviceClass.extendedResourceName=example.com/gpu \
+  ${BINDING_CONDITIONS_OPTS} \
```
nojnhuh: FYI: #201 will merge soon and will conflict with this change. Hopefully that makes this unnecessary by letting the e2e test itself tweak the driver parameters.

README.md
```diff
@@ -401,7 +401,8 @@ kubectl delete --wait=false --filename=demo/basic-resourceclaimtemplate.yaml \
   --filename=demo/basic-shared-claim-across-containers.yaml \
   --filename=demo/basic-shared-claim-across-pods.yaml \
   --filename=demo/basic-resourceclaim-opaque-config.yaml \
-  --filename=demo/admin-access.yaml
+  --filename=demo/admin-access.yaml \
+  --filename=demo/binding-conditions/binding-conditions.yaml
```
nojnhuh: Now that binding conditions have their own dedicated demo, we can remove it from the main README.

deployments/helm/dra-example-driver/values.yaml
ttsuuubasa: I initially considered keeping this as a single value, but decided to split it because I thought that would be more flexible.
One reason is to support use cases where, for example, users only want to expose BindingConditions on the ResourceSlice side, while implementing their own controller. If we keep it as a single value, the example driver’s controller would also be started automatically, and BindingConditions would be set automatically as well. In that case, it would be difficult to verify the behavior of the custom controller.
To accommodate such scenarios, I chose to separate the values—what do you think?

demo/binding-conditions/binding-conditions.yaml
ttsuuubasa: That’s a good point. I’ll add it.

demo/binding-conditions/README.md
ttsuuubasa: I’ll update it.

deployments/helm/dra-example-driver/templates/clusterrole.yaml
ttsuuubasa: That’s true - `arbitrary-node:update` wasn’t necessary, and it worked without it. I’ll remove it.

internal/profiles/profiles.go
ttsuuubasa: Right, I’ll move them to `gpu.go`.

demo/binding-conditions/binding-conditions.yaml
ttsuuubasa: DONE

demo/binding-conditions/README.md
ttsuuubasa: DONE

deployments/helm/dra-example-driver/templates/clusterrole.yaml
ttsuuubasa: The latest commit added fine‑grained RBAC, so I kept it as is and did not add `arbitrary-node:update`.

internal/profiles/profiles.go
ttsuuubasa: DONE

test/e2e/setup-e2e.sh
ttsuuubasa: With this change, I switched to using the `installDriver()` function and moved the BindingConditions-related configuration into the e2e test, removing the need for environment variables like `${BINDING_CONDITIONS_OPTS}` as a prerequisite.

README.md
ttsuuubasa: DONE

cmd/dra-example-controller/controller.go
ttsuuubasa: As a small start, I’ve added a few simple unit tests.
Let me know if there are other unit tests you think would be useful.

ttsuuubasa:
@nojnhuh 
Thanks for the review. I’ve addressed your comments—could you please take another look?

nojnhuh (commented):
LGTM aside from the minor issues pointed out by the linter and the unit test failure.

ttsuuubasa:
@nojnhuh 
Thanks for the review. I’ve addressed the lint issues and the unit test failure.

nojnhuh (approved):
Thanks!

/lgtm
/approve

ttsuuubasa:
@nojnhuh 
Thank you for the approval! However, it looks like the merge hasn’t progressed as expected.

nojnhuh:
Let's try again

/lgtm
/approve

PR #201, opened 2026-05-13
Install a separate driver instance for each test

### LINKED ISSUE #171: Improve e2e test isolation, error reporting, and extensibility

willie-yao:
Each `It` block now installs its own DRA driver release with a unique driver name, and uninstalls it on teardown.

Fixes #171 

Changes:
 
 - New `test/e2e/driver_test.go` with `installDriver(ctx, DriverConfig)` helper. Shells out to `helm` (already required by `setup-e2e.sh`); honors `HELM_CHART_PATH` env var.
 - `deployManifest` accepts a `DriverConfig` and substitutes `gpu.example.com` -> per-test driver name (and `example.com/gpu` -> per-test extended resource name) in demo manifests before applying.
 - Beyond `helm --wait`, polls explicitly for `DeviceClass` + `ResourceSlices` + webhook readiness before returning from install.
 - Per-test `DeferCleanup` LIFO: driver-log diagnostics -> helm uninstall -> namespace delete with foreground propagation + termination wait.
 - `setup-e2e.sh` no longer installs the driver. Cert-manager + kind cluster + image build stay (cluster-wide infra).
 - Webhook specs share one driver pinned to `gpu.example.com` and run `Ordered, Serial` so their static testdata stays valid.


test/e2e/driver_test.go
```diff
@@ -0,0 +1,316 @@
...
+// runHelm executes a helm subcommand and returns its combined output.
+func runHelm(ctx context.Context, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, "helm", args...)
+	out, err := cmd.CombinedOutput()
+	return string(out), err
+}
+
+// verifyWebhook waits until the validating webhook is serving for the given
+// DeviceClass by creating a dry-run ResourceClaim until it succeeds.
+func verifyWebhook(ctx context.Context, deviceClassName string) {
```
willie-yao, line 316: This code is mostly moved from e2e_setup_test.go at line 107: https://github.com/willie-yao/dra-example-driver/blob/main/test/e2e/e2e_setup_test.go#L107

Changes: now takes a deviceClassNAme parameter, the test claim's Name is suffixed with it due to parallel webhook installs and added DeviceClass to error messages

test/e2e/driver_test.go
```diff
@@ -0,0 +1,316 @@
...
+				continue
+			}
+			buf := new(bytes.Buffer)
+			_, _ = io.Copy(buf, stream)
+			stream.Close()
+			fmt.Fprintf(GinkgoWriter,
+				"Driver pod %s, container %s (last %d lines):\n%s\n",
+				pod.Name, c.Name, tailLines, buf.String())
+		}
+	}
```
willie-yao: This is moved from e2e_setup_test.go on line 192: https://github.com/willie-yao/dra-example-driver/blob/main/test/e2e/e2e_setup_test.go#L192

test/e2e/driver_test.go
```diff
@@ -0,0 +1,316 @@
...
+func resolveChartPath() string {
+	if p := os.Getenv(chartPathEnvVar); p != "" {
+		return p
+	}
+	return filepath.Join(rootDir, "deployments", "helm", "dra-example-driver")
+}
+
+// runHelm executes a helm subcommand and returns its combined output.
+func runHelm(ctx context.Context, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, "helm", args...)
```
nojnhuh: Can we use Helm's Go library instead of invoking the binary? To keep dependencies separate, we should turn the `test` directory into a separate module like CAPZ.

willie-yao: Done in https://github.com/kubernetes-sigs/dra-example-driver/pull/201/changes/784f6e91bcc1f67b9d0f370bde092b28f47e9426. CAPZ doesn't actually have an entirely different submodule but uses a `.golangci.yml`, but I think a sub-module would work better here anyways.

.github/workflows/lint.yaml
```diff
@@ -10,9 +10,9 @@ jobs:
...
-      uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
+    - name: Install Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
       with:
-        version: latest
-        args: -v --timeout 5m
-        skip-cache: true
+        go-version: 1.26.2
+    - name: Lint
+      run: make lint
```
willie-yao: These changes were needed to accommodate the new test submodule

test/e2e/driver_test.go
nojnhuh: Ah, CAPI does this but I forgot CAPZ doesn't.

test/go.mod
```diff
@@ -0,0 +1,126 @@
+module sigs.k8s.io/dra-example-driver/test
+
+go 1.26.0
+
+replace sigs.k8s.io/dra-example-driver => ../
```
nojnhuh: If we set up a [workspace](https://go.dev/doc/tutorial/workspaces) can we drop the `replace`?

I'd also be interested to know if that makes things like `golangci-lint run ./...` operate across both modules or if we still need to call that twice.

willie-yao: I just tried this and it looks like workspace does let us drop replace, but it doesn't work with golangci-lint. It seems like golangci-lint doesn't honor go.work, which means the root lint fails with unknown revision v0.0.0. Turning the workspace off breaks the test module's lint as well (missing go.sum entry). I feel like sticking with replace is the best course of action here.

nojnhuh: With go.work, if you remove the v0.0.0 dependency on the main module then `go mod tidy` the test module, then I think golangci-lint should work.

Makefile
```diff
@@ -80,6 +80,7 @@ ineffassign:
 
 lint:
 	golangci-lint run ./...
+	cd test && golangci-lint run --build-tags=e2e ./...
```
nojnhuh: This hides lint failures in the test module if any issues are found in the main module. Can we combine into one call?


```suggestion
golangci-lint run ./...
	golangci-lint run --build-tags=e2e ./... ./test/...
```

.github/workflows/lint.yaml
```diff
@@ -10,9 +10,20 @@ jobs:
...
-    - name: Lint
+    - name: Install Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      with:
+        go-version: 1.26.2
+    - name: Lint root module
       uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
       with:
         version: latest
         args: -v --timeout 5m
```
nojnhuh: If we can call golangci-lint once to lint both modules on the command line, can we similarly consolidate these? I suppose something like this should work:

```suggestion
        args: -v --timeout 5m --build-tags=e2e ./... ./test/...
```

test/go.mod
```diff
@@ -0,0 +1,126 @@
...
+
+require (
+	github.com/onsi/ginkgo/v2 v2.28.3
+	github.com/onsi/gomega v1.40.0
+	helm.sh/helm/v3 v3.20.2
+	k8s.io/api v0.36.0
+	k8s.io/apimachinery v0.36.0
+	k8s.io/client-go v0.36.0
+	sigs.k8s.io/dra-example-driver v0.0.0
```
nojnhuh: `go mod tidy` seems to want to put v0.2.1 here if this doesn't already exist. That should still ultimately resolve to the local version though and not that actual tagged version.

test/go.mod
nojnhuh, line 1: Let's make sure dependabot is wired up to this new module.

Makefile
```diff
@@ -126,7 +127,7 @@ setup-e2e:
 	test/e2e/setup-e2e.sh
 
 test-e2e:
-	go run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -p ./test/e2e/...
+	cd test && go run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -p ./e2e/...
```
nojnhuh: nit: I find it annoying to track the scope of any given `cd`, so I try to avoid that if I can. This does the same thing and it's more clear to me that neither another command right below this nor my interactive shell will stay in the new directory:
```suggestion
	go -C test run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -p ./e2e/...
```

nojnhuh (commented):
Sorry for the annoying comments on the workspace/lint periphery here. I promise I'm starting to look at the main point of this PR.

Makefile
willie-yao: Tried combining it with `golangci-lint run --build-tags=e2e ./... ./test/...` but it silently doesn't lint the test module. It logs pattern ./test/...: directory prefix test does not contain main module or its selected dependencies and exits 0 with 0 issues. I'm thinking to do it by just creating another make target for test

```
.PHONY: lint lint-root lint-test
lint:
	@$(MAKE) -k lint-root lint-test

lint-root:
	golangci-lint run ./...

lint-test:
	cd test && golangci-lint run --build-tags=e2e ./...
```

.github/workflows/lint.yaml
willie-yao: I'm not sure if this will work for the same reason as above. I did add an if: always() so that the lint also runs for the test module even if it fails for the main module.

test/go.mod
willie-yao: I confirmed locally that with replace => ../ present, v0.2.1 or any other version resolves to the local checkout. Should I change this to v0.2.1?

Makefile
nojnhuh: With a workspace set up, it seems to work for me: https://github.com/willie-yao/dra-example-driver/compare/separate-driver-install...nojnhuh:dra-example-driver:workspace-lint

```console
% make lint
golangci-lint run --build-tags=e2e ./... ./test/...
cmd/dra-example-webhook/main.go:183:1: Comment should end in a period (godot)
// function
^
test/e2e/e2e_setup_test.go:54:1: Comment should end in a period (godot)
// driverPodSelector finds kubelet plugin Pods within an installed driver's release namespace
^
2 issues:
* godot: 2
make: *** [Makefile:83: lint] Error 1
```

willie-yao: Ah that was completely my bad. I had v0.0.0 set before when testing something. Setting it to v0.2.1 fixed it! Thanks for checking this for me

test/e2e/driver_test.go
```diff
@@ -0,0 +1,386 @@
...
+	runUpgradeOrInstall(ctx, actionCfg, install, chrt, buildHelmValues(cfg))
+
+	waitForDriverReady(ctx, cfg)
+	return cfg
+}
+
+// runUpgradeOrInstall emulates `helm upgrade --install`: if a release with
+// the configured name already exists it is upgraded in place, otherwise a
+// fresh install is performed.
+func runUpgradeOrInstall(ctx context.Context, actionCfg *action.Configuration, install *action.Install, chrt *chart.Chart, values map[string]any) {
```
nojnhuh: In practice, will this ever actually perform an upgrade of an existing Helm release? If we expect each test to have a separate instance of the driver, then I think we'd want to error if that driver already exists instead of upgrade and risk affecting a different test.

The script used `helm upgrade --install` before mostly to be copy-pasteable, but that was always expected to first be run on a fresh kind cluster without the driver installed where a plain `helm install` should have worked.

test/e2e/driver_test.go
```diff
@@ -0,0 +1,386 @@
...
+	// LIFO so driver-log diagnostics run first.
+	DeferCleanup(uninstallDriver, cfg, NodeTimeout(driverUninstallTimeout))
+	DeferCleanup(dumpDriverDiagnostics, cfg, NodeTimeout(30*time.Second))
+
+	registryClient, err := registry.NewClient()
+	Expect(err).NotTo(HaveOccurred(), "Failed to create OCI registry client")
+	actionCfg := newHelmActionConfig(cfg.Namespace, registryClient)
+
+	install := action.NewInstall(actionCfg)
+	install.ReleaseName = cfg.ReleaseName
```
nojnhuh: Could we automatically generate a release name instead? That will make sure there aren't conflicts between tests, tests don't have to decide for themselves what the driver should be called, and we don't have to validate them up front.

test/e2e/driver_test.go
```diff
@@ -0,0 +1,386 @@
...
+		"Failed to upgrade helm release %s/%s", install.Namespace, install.ReleaseName)
+}
+
+// withDriverDefaults validates the release name and fills in unset fields.
+func withDriverDefaults(cfg DriverConfig) DriverConfig {
+	GinkgoHelper()
+	Expect(validation.IsDNS1123Label(cfg.ReleaseName)).To(BeEmpty(),
+		"DriverConfig.ReleaseName %q must be a valid DNS-1123 label", cfg.ReleaseName)
+	if cfg.Namespace == "" {
+		cfg.Namespace = cfg.ReleaseName
```
nojnhuh: Like the release name, can we generate a new, separate namespace for each instance of the driver?

test/e2e/driver_test.go
```diff
@@ -0,0 +1,386 @@
...
+	driverUninstallTimeout = 60 * time.Second
+
+	// defaultDriverNumDevices is the smallest count that satisfies every demo
+	// manifest. Tests that need more can override via DriverConfig.NumDevices.
+	defaultDriverNumDevices = 2
+
+	// maxDriverNameLen is the longest driver name that fits within Linux's
+	// 108-byte UNIX_PATH_MAX after the kubelet appends its registrar socket
+	// prefix and per-pod UID suffix.
+	maxDriverNameLen = 28
```
nojnhuh: Can you expand on how this value was calculated?

test/e2e/e2e_test.go
nojnhuh: Does it simplify things if we construct these ResourceClaims inline vs. reading them from a file?

test/e2e/e2e_setup_test.go
```diff
@@ -97,69 +103,52 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
 	DeferCleanup(dumpDiagnosticsOnFailure, namespace, NodeTimeout(15*time.Second))
 }
 
-// dumpDiagnosticsOnFailure collects pod status, events, and driver logs
-// when a test has failed. Intended for use as a DeferCleanup callback.
+// substituteDriverIdentifiers replaces the demo manifest driver identifiers
+// with the per-test driver name and (when set) extended resource name.
+func substituteDriverIdentifiers(raw string, drv DriverConfig) string {
+	GinkgoHelper()
+	out := strings.ReplaceAll(raw, defaultDeviceClassName, drv.DriverName)
```
nojnhuh, line 138: Eventually, the manifests will be referencing other drivers in addition to `gpu.example.com`. Some manifests may even rely on multiple separate instances of the driver running different profiles. Not a blocker for this PR, but we should find a better way to handle dynamic driver names.

test/e2e/driver_test.go
willie-yao: Ah you're right, this actually neds up hiding the errors and was leftover from trying to replicate `helm upgrade --install`. I'll remove it and use `install.RunWithContext` directly.

test/e2e/driver_test.go
willie-yao: TBH copilot came up with this and I don't think it's necessary for the E2E test since we auto-generate default driver names that won't go over the limit anyways. I'll go ahead and remove it

nojnhuh:
FYI @willie-yao A couple of dependabot PRs merged which conflict with this PR.

test/e2e/driver_test.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// withDriverDefaults populates auto-generated fields and fills in unset
+// optional fields with package defaults.
+func withDriverDefaults(cfg DriverConfig) DriverConfig {
+	GinkgoHelper()
+	Expect(cfg.ReleaseName).To(BeEmpty(),
+		"DriverConfig.ReleaseName is populated by installDriver; callers must not set it")
+	Expect(cfg.Namespace).To(BeEmpty(),
+		"DriverConfig.Namespace is populated by installDriver; callers must not set it")
```
nojnhuh: Making these available to tweak and then checking to make sure they weren't tweaked isn't ideal. Could we set this in a separate local variable in `installDriver` and pass that around separately from the `DriverConfig`?

nojnhuh (commented):
One more nit, otherwise LGTM.

Could you also check that all the mod/work/sum files are up to date?
```
go mod tidy && go -C test mod tidy
go work sync
go mod tidy && go -C test mod tidy # again
```

nojnhuh:
> Could you also check that all the mod/work/sum files are up to date?
> ```
> go mod tidy && go -C test mod tidy
> go work sync
> go mod tidy && go -C test mod tidy # again
> ```

Having written this out, this sounds kind of insane and I'm open to the possibility that separate modules is going to be more pain going forward than it's worth. We could try sticking with one module for now and see if anybody runs into any issues. Splitting things up later will probably be easier than putting them back together.

willie-yao:
> Having written this out, this sounds kind of insane and I'm open to the possibility that separate modules is going to be more pain going forward than it's worth. We could try sticking with one module for now and see if anybody runs into any issues. Splitting things up later will probably be easier than putting them back together.

Yeah you're right. I noticed the go.work and go.sum files kept going out of date even when doing go mod tidy... I can go ahead and refactor into one module if you think that's a better way to go about things.

nojnhuh:
> > Having written this out, this sounds kind of insane and I'm open to the possibility that separate modules is going to be more pain going forward than it's worth. We could try sticking with one module for now and see if anybody runs into any issues. Splitting things up later will probably be easier than putting them back together.
> 
> Yeah you're right. I noticed the go.work and go.sum files kept going out of date even when doing go mod tidy... I can go ahead and refactor into one module if you think that's a better way to go about things.

Yeah, I'm thinking let's go back to one module for now. Trying it here was still a worthwhile experiment nonetheless.

go.mod
```diff
@@ -4,88 +4,153 @@ go 1.26.0
...
 	github.com/google/uuid v1.6.0
-	github.com/onsi/ginkgo/v2 v2.29.0
-	github.com/onsi/gomega v1.41.0
+	github.com/onsi/ginkgo/v2 v2.28.1
+	github.com/onsi/gomega v1.39.1
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.11.1
 	github.com/urfave/cli/v2 v2.27.7
 	google.golang.org/grpc v1.81.1
+	helm.sh/helm/v3 v3.21.0
```
nojnhuh: There's a v4 for Helm which I don't think has too many changes from v3. Can we use that here? If it turns out to be more work, let's file an issue to update this since I'm not sure how long v3 will get security updates.

nojnhuh (approved):
/lgtm
/approve

/hold in case we want to try to use Helm v4

nojnhuh:
@willie-yao Could you also please squash?

willie-yao:
Added helm v4
/hold cancel

PR #139, opened 2026-01-19
Add net profile (with consumable capacity)

sunya-ch:
based on 
- https://github.com/kubernetes-sigs/dra-example-driver/pull/138

Please review only last commit.

Closes https://github.com/kubernetes-sigs/dra-example-driver/issues/137

PR #203, opened 2026-05-15
Add example for native resource requests

### LINKED ISSUE #184: Add example for native resource requests (KEP-5517)
KEP-5517 (https://github.com/kubernetes/enhancements/issues/5517) handles resource accounting for core resources (CPU, Memory) managed through DRA. When a DRA driver manages core resources (like https://github.com/kubernetes-sigs/dra-driver-cpu), the standard kube-scheduler resource accounting model is unaware of those allocations, leading to potential oversubscription. This KEP introduces a unified accounting model. It reached Alpha in 1.36.

The example driver should demonstrate how a DRA driver can advertise and manage native resources while maintaining correct resource accounting with the scheduler.

willie-yao:
Fixes #184 

Adds a new `cpu` device profile that demonstrates the DRA native (node-allocatable) resource accounting feature ([KEP-5517](https://github.com/kubernetes/enhancements/issues/5517), Alpha in 1.36 behind `DRANodeAllocatableResources`), modeled after how a real DRA CPU driver ([dra-driver-cpu](https://github.com/kubernetes-sigs/dra-driver-cpu)) represents devices.

The driver publishes one DRA device per fake NUMA node, each one exposing its CPUs as consumable capacity. Every device carries `NodeAllocatableResourceMappings` with `CapacityKey` pointing at that capacity pool, so kube-scheduler debits the consumed CPU count from the node's standard allocatable budget every time a claim is satisfied. This prevents oversubscription when CPUs are managed through DRA alongside the classic pod-spec request API.

Because `AllowMultipleAllocations: true` is set on each NUMA device, multiple pods can share a NUMA node by claiming different amounts of capacity, mirroring real-world CPU pooling.

The `cpu` profile does not support opaque per-claim configuration in this iteration (any `parameters` block on a CPU `ResourceClaim` is rejected by the webhook). There's no `CpuConfig` API type defined yet because no configurable knob has surfaced as needed for the demo. The obvious candidates (exclusive vs shared, SMT-sibling preference, etc.) belong to follow-up examples that exercise specific behaviors. Adding opaque config later is a straightforward additive change: define the type under `api/example.com/resource/cpu/v1alpha1`, register it in `SchemeBuilder()`, and handle the typed config in `Validate` / `ApplyConfig`.

### Changes

- `internal/profiles/cpu/cpu.go` (new): publishes per-NUMA `Device`s with `AllowMultipleAllocations: true`, `Capacity[<driverName>/cpu]`, and `NodeAllocatableResourceMappings[cpu].CapacityKey`. The capacity key is derived from the driver name at runtime so multiple driver releases can coexist without colliding.
- `cmd/dra-example-kubeletplugin/main.go` and `cmd/dra-example-webhook/main.go`: register the `cpu` profile, plumb a new `--cpus-per-numa-node` flag.
- Helm chart: `cpu` added to the `deviceProfile` enum; new `kubeletPlugin.cpu.cpusPerNUMANode` value wired through.
- `demo/scripts/kind-cluster-config.yaml`: enables both `DRANodeAllocatableResources` and `DRAConsumableCapacity` feature gates (the latter is needed for `AllowMultipleAllocations`, `Device.Capacity`, and claim-side `capacity.requests`).
- `demo/native-resource-request.yaml`: the example. Pod claims `2` units of `<driver>/cpu` capacity via a `ResourceClaimTemplate`.
- `test/e2e/e2e_setup_test.go`: extend `substituteDriverIdentifiers` to also rewrite `cpu.example.com`; add `verifyNodeAllocatableResourceClaimStatus` and `waitForResourceSlicesWithNodeAllocatableMappings` (the latter accepts an `expectedMapping` that handles both `AllocationMultiplier` and `CapacityKey` shapes).
- `test/e2e/e2e_test.go`: new spec that installs the cpu profile, verifies the published mappings, deploys the demo, and asserts `pod.status.nodeAllocatableResourceClaimStatuses`.
- Unit tests for the cpu profile covering device naming, NUMA attributes, `AllowMultipleAllocations`, capacity, `CapacityKey` mapping shape, env-var injection, and config validation.

### DIFF
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/dra-example-driver/internal/profiles"
+	"sigs.k8s.io/dra-example-driver/internal/profiles/cpu"
 	"sigs.k8s.io/dra-example-driver/internal/profiles/gpu"
 	"sigs.k8s.io/dra-example-driver/pkg/flags"
 )
@@ -54,6 +55,7 @@ type Flags struct {
 	driverName                    string
 	podUID                        string
 	gpuPartitions                 int
+	cpusPerNUMANode               int
 }
 
 type Config struct {
@@ -67,6 +69,9 @@ type Config struct {
 var validProfiles = map[string]func(flags Flags) profiles.Profile{
 	gpu.ProfileName: func(flags Flags) profiles.Profile {
 		return gpu.NewProfile(flags.nodeName, flags.numDevices, flags.gpuPartitions)
+	},
+	cpu.ProfileName: func(flags Flags) profiles.Profile {
+		return cpu.NewProfile(flags.nodeName, flags.driverName, flags.numDevices, flags.cpusPerNUMANode)
 	},
 }
 
@@ -110,7 +115,7 @@ func newApp() *cli.App {
 		},
 		&cli.IntFlag{
 			Name:        "num-devices",
-			Usage:       "The number of devices to be generated. Only relevant for the " + gpu.ProfileName + " profile.",
+			Usage:       "Number of top-level devices to advertise: GPUs for the " + gpu.ProfileName + " profile, NUMA nodes for the " + cpu.ProfileName + " profile.",
 			Value:       8,
 			Destination: &flags.numDevices,
 			EnvVars:     []string{"NUM_DEVICES"},
@@ -161,6 +166,13 @@ func newApp() *cli.App {
 			Value:       0,
 			Destination: &flags.gpuPartitions,
 			EnvVars:     []string{"GPU_PARTITIONS"},
+		},
+		&cli.IntFlag{
+			Name:        "cpus-per-numa-node",
+			Usage:       "Number of CPUs each fake NUMA-node device advertises as consumable capacity. Only relevant for the " + cpu.ProfileName + " profile.",
+			Value:       4,
+			Destination: &flags.cpusPerNUMANode,
+			EnvVars:     []string{"CPUS_PER_NUMA_NODE"},
 		},
 	}
 	cliFlags = append(cliFlags, flags.kubeClientConfig.Flags()...)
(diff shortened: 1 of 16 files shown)


nojnhuh:
Could we formulate this as a new `cpu` profile instead of adding on to the `gpu` profile? I think that CPU use case is a bit more broadly applicable than the "GPU that requires CPU" use case. That also opens up the possibility for more examples demonstrating how to request CPUs and GPUs together. We can skip memory in this first iteration IMO.

willie-yao:
@nojnhuh I added a proper cpu profile, lmk what you think!

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -111,7 +116,7 @@ func newApp() *cli.App {
 		},
 		&cli.IntFlag{
 			Name:        "num-devices",
-			Usage:       "The number of devices to be generated. Only relevant for the " + gpu.ProfileName + " profile.",
+			Usage:       "Number of top-level devices to advertise: GPUs for the " + gpu.ProfileName + " profile, NUMA nodes for the " + cpu.ProfileName + " profile.",
```
nojnhuh: Can we make this a separate flag, like `--cpu-numa-nodes`? "How many devices" is going to be ambiguous if we start using it for everything.

internal/profiles/cpu/cpu.go
```diff
@@ -0,0 +1,131 @@
...
+	return resourceapi.QualifiedName(p.driverName + "/" + CPUCapacitySuffix)
+}
+
+func (p Profile) EnumerateDevices() (resourceslice.DriverResources, error) {
+	capacityKey := p.CapacityKey()
+	devices := make([]resourceapi.Device, p.numNUMANodes)
+	for i := 0; i < p.numNUMANodes; i++ {
+		devices[i] = resourceapi.Device{
+			Name:                     fmt.Sprintf("numa-%d", i),
+			AllowMultipleAllocations: ptr.To(true),
```
nojnhuh, line 73: I think to support consumable capacity we'll need to add the `ShareID` to the gRPC response like #139 does: https://github.com/kubernetes-sigs/dra-example-driver/pull/139/changes#diff-37a4b723138c8e7d62ccb364c381c4965f48c5792835f756c1ad1b43b8f58116R116-R120
https://github.com/kubernetes-sigs/dra-example-driver/blob/5bbf7f7f620963cbf898b137a6ccde89b772a970/cmd/dra-example-kubeletplugin/driver.go#L117-L119

@sunya-ch Is that right?

nojnhuh (approved):
/lgtm
/approve

PR #206, opened 2026-05-17
Add demo for DRA Resource Availability Visibility (KEP-5677)

### LINKED ISSUE #187: Add example for resource availability visibility (KEP-5677)
KEP-5677 (https://github.com/kubernetes/enhancements/issues/5677) adds a `ResourcePoolStatusRequest` API that allows users to query DRA resource pool availability, and enhances `kubectl describe resourceslice` and `kubectl describe node` to show DRA resource consumption information. This reached Alpha in 1.36.

The example driver should include a demo or documentation showing how to use the new APIs and kubectl commands to inspect resource pool availability with the example driver's devices.

nmn3m:
## Summary

[KEP-5677](https://github.com/kubernetes/enhancements/issues/5677) introduced a new `ResourcePoolStatusRequest` API in Kubernetes 1.36 (alpha) that lets users query how many devices in each DRA pool are allocated vs. available, without needing read access to every namespace's `ResourceClaim`s. The aggregation runs data, so the example driver itself needs no code changes — only a demo.

  Fixes #187

### DIFF
--- a/demo/resource-pool-status.yaml
+++ b/demo/resource-pool-status.yaml
@@ -1,0 +1,95 @@
+# Example: DRA Resource Availability Visibility (ResourcePoolStatusRequest)
+#
+# Demonstrates the
+# [DRA resource pool status](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#resource-pool-status)
+# feature.
+# For more information see
+# [KEP-5677](https://github.com/kubernetes/enhancements/issues/5677).
+#
+# This manifest is self-contained: it creates one Pod that consumes 1 GPU and a
+# cluster-scoped `ResourcePoolStatusRequest` that asks
+# `kube-controller-manager` to publish a one-shot snapshot of every pool served
+# by the example driver. The snapshot lists total / allocated / available /
+# unavailable device counts per pool.
+#
+# `ResourcePoolStatusRequest` is particularly useful for non-admin users:
+# `ResourceClaim`s are namespaced, so a user cannot ordinarily inspect claims
+# in other namespaces. A cluster-scoped `ResourcePoolStatusRequest` lets them
+# see aggregate consumption without that visibility. The driver itself needs
+# no code changes; the aggregation is computed entirely by
+# `kube-controller-manager` from existing `ResourceSlice`s and
+# `ResourceClaim`s.
+#
+# Expected: once the controller reconciles the request, `.status` reports
+# `allocatedDevices: 1` for the pool backing the consumer Pod. Inspect with:
+#   kubectl wait --for=condition=Complete \
+#     resourcepoolstatusrequest/gpu-pool-status --timeout=30s
+#   kubectl get resourcepoolstatusrequest/gpu-pool-status -o yaml
+#
+# `ResourcePoolStatusRequest` is one-shot. To refresh the snapshot after
+# launching or tearing down workloads, delete and re-create the request:
+#   kubectl delete resourcepoolstatusrequest/gpu-pool-status
+#   kubectl apply --filename=demo/resource-pool-status.yaml
+#
+# Driver requirements:
+#   Profile: gpu
+#   GPUs: 1
+#
+# Cluster requirements:
+#   Kubernetes 1.36+
+#   Feature gate: DRAResourcePoolStatus (on kube-apiserver and
+#     kube-controller-manager)
+#   API enabled: resource.k8s.io/v1alpha3
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -391,6 +391,56 @@ To run this demo:
 
 This demonstration shows the end-to-end flow of the DRA AdminAccess feature. In a production environment, drivers could use this admin access indication to provide additional privileged capabilities or information to authorized workloads.
 
+### Demo DRA Resource Availability Visibility Feature
```
nojnhuh: Overall it seems like there's a lot of overlap between the content here and the comment in the example manifest. Do you think we could consolidate to only the example manifest? I actually don't see anything critical here that isn't already mentioned in the example manifest, so maybe we can simply delete this? 

If separate docs still make sense, could we instead make a directory under `demo/` and add this to a new README.md there? That directory would contain the YAML manifests then too.

README.md
```diff
@@ -391,6 +391,56 @@ To run this demo:
...
+This example driver works with the [DRA Resource Availability Visibility feature](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5677-dra-resource-availability-visibility) ([KEP-5677](https://github.com/kubernetes/enhancements/issues/5677)), which adds a `ResourcePoolStatusRequest` API for querying how many devices in each pool are allocated vs. available. The aggregation is computed by `kube-controller-manager` from the existing ResourceSlices and ResourceClaims — the driver itself needs no code changes. The feature reached Alpha in Kubernetes 1.36; the demo `kind` cluster created above already enables the `DRAResourcePoolStatus` feature gate and serves the `resource.k8s.io/v1alpha3` API.
+
+This is particularly useful for non-admin users: `ResourceClaim`s are namespaced, so a user cannot ordinarily inspect claims in other namespaces. A cluster-scoped `ResourcePoolStatusRequest` lets them see aggregate consumption without that visibility.
+
+#### Usage Example
```
nojnhuh: Could we add some lightweight verification to the e2e tests? We don't need to reimplement the k/k tests, just enough to make sure that the example isn't outright broken, like if the driver name is misspelled or anything.

README.md
```diff
@@ -391,6 +391,56 @@ To run this demo:
 
 This demonstration shows the end-to-end flow of the DRA AdminAccess feature. In a production environment, drivers could use this admin access indication to provide additional privileged capabilities or information to authorized workloads.
 
+### Demo DRA Resource Availability Visibility Feature
+This example driver works with the [DRA Resource Availability Visibility feature](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5677-dra-resource-availability-visibility) ([KEP-5677](https://github.com/kubernetes/enhancements/issues/5677)), which adds a `ResourcePoolStatusRequest` API for querying how many devices in each pool are allocated vs. available. The aggregation is computed by `kube-controller-manager` from the existing ResourceSlices and ResourceClaims — the driver itself needs no code changes. The feature reached Alpha in Kubernetes 1.36; the demo `kind` cluster created above already enables the `DRAResourcePoolStatus` feature gate and serves the `resource.k8s.io/v1alpha3` API.
```
nojnhuh: Could we split this raw markdown into separate lines? https://kubernetes.io/docs/contribute/style/style-guide/#line-breaks

nojnhuh: > This example driver works with the [DRA Resource Availability Visibility feature](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5677-dra-resource-availability-visibility) ([KEP-5677](https://github.com/kubernetes/enhancements/issues/5677))

Could we lead with a link to the feature docs instead of the KEP? https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#resource-pool-status

A "for more information" link to the KEP is probably still useful, but the user-facing docs are probably a better starting point.

README.md
```diff
@@ -391,6 +391,56 @@ To run this demo:
...
+
+This is particularly useful for non-admin users: `ResourceClaim`s are namespaced, so a user cannot ordinarily inspect claims in other namespaces. A cluster-scoped `ResourcePoolStatusRequest` lets them see aggregate consumption without that visibility.
+
+#### Usage Example
+
+Start by deploying one of the `basic-*` demos from earlier in this guide so that there is something to consume GPUs. For example, `demo/basic-resourceclaimtemplate.yaml` allocates 2 GPUs from the worker's pool of 8.
```
nojnhuh: Ideally this shouldn't depend on any specific combination of other examples already deployed in the cluster, so if a Pod+ResourceClaim consuming some resources is useful to illustrate the feature then let's add one dedicated to this example.

PR #207, opened 2026-05-18
Name the driver with a *.k8s.io domain instead of *.example.com

### LINKED ISSUE #174: Name the driver with a *.k8s.io domain instead of *.example.com
In https://github.com/kubernetes-sigs/dra-example-driver/pull/148#discussion_r3123043705, @pohly suggests `dra-example-driver.sigs.k8s.io`, but other suggestions are welcome.

cc @klueska @johnbelamaric

willie-yao:
Fixes #174 

This PR renames the driver with a k8s.io domain instead of example.com. Most instances of example.com are replaced with `dra-example-driver.sigs.k8s.io`.


common.mk
willie-yao: I left this `example.com` as-is since it's a placeholder for registry unrelated to the driver identity, but I can also update this if we want to remove all mentions of `example.com`.

test/e2e/e2e_test.go
```diff
@@ -158,7 +158,7 @@ var _ = Describe("Test GPU allocation", func() {
 		containerName := "ctr0"
 		expectedGPUCount := 1
 		expectedResourceNames := map[string]string{
-			"pod0": "deviceclass.resource.kubernetes.io/gpu.example.com",
+			"pod0": "deviceclass.resource.kubernetes.io/gpu.dra-example-driver.sigs.k8s.io",
 			"pod1": "example.com/gpu",
```
willie-yao: I also left `example.com/gpu` for the extended resource names since that was what KEP-5004 used for their examples, but I'm also happy to change this if you think it'll be better.

nojnhuh (changes requested):
I still see some `example.com`s, mostly limited to the extended resources example and test. I think we can move away from "example.com" everywhere.

willie-yao:
@nojnhuh Removed all mentions!

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -186,7 +186,7 @@ func newApp() *cli.App {
 			}
 
 			if flags.driverName == "" {
-				flags.driverName = flags.profile + ".example.com"
+				flags.driverName = flags.profile + ".dra-example-driver.sigs.k8s.io"
```
nojnhuh, line 189: @pohly The e2e tests here are failing because the kubelet plugin registration socket name is too long with a driver name formatted this way. e.g. `/var/lib/kubelet/plugins_registry/gpu.dra-example-driver.sigs.k8s.io-f358c92b-e974-4f29-a84f-6c64bec0949d-reg.sock`. That is 114 bytes long and the [limit is 108 bytes](https://manpages.ubuntu.com/manpages/jammy/man7/unix.7.html#Address_format.).

Assuming we can't shorten the kubelet plugin registry directory path and that we want to enable seamless upgrades, we have 28 characters left for the driver name, (or [maybe 27](https://manpages.ubuntu.com/manpages/jammy/man7/unix.7.html#bugs)?), and `.sigs.k8s.io` takes up 12.

Overall I'm slightly more inclined to stick with `example.com` than clipping `dra-example-driver.sigs.k8s.io` into something that might be ambiguous and make it challenging to name new device profiles.

cc @johnbelamaric @klueska In case either of you have suggestions for a concise driver name.

pohly: I think it would be worthwhile to support DRA driver names longer than ~28 characters. That's pretty tight. The socket name includes a full UUID. Maybe we can use something shorter there?

nojnhuh: Is there anything we can do in this driver if the 28-character limit ultimately comes from the helper? The UUID portion is added for rolling updates and the helper doesn't allow setting the socket name when rolling updates are enabled. If there's a better way to format socket names, that would have to change in the helper. I think the driver name is all that we can tweak here right now unless we disable rolling updates by default.

pohly: I'm afraid you are right: we would first have to improve the helper, then rebase onto it and use the longer driver name. That implies waiting one release cycle.

I'm okay with staying with `example.com` for now, but would like to see the driver make the change after 1.37.

nojnhuh: Opened https://github.com/kubernetes/kubernetes/issues/139166 to track this for the helper.

Thanks @willie-yao for bringing this to light. I'll close this PR for now and we can revisit this later.

/close

PR #215, opened 2026-06-01
demo: add helper for installing KubeVirt daily build

nestoracunablanco:
- fetch latest KubeVirt nightly version dymamically
- apply operator and CR manifests from nightly release
- wait for KubeVirt availability
- enable DRA-related KubeVirt feature gates

### DIFF
--- a/demo/scripts/install-kubevirt-daily.sh
+++ b/demo/scripts/install-kubevirt-daily.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Copyright The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


nestoracunablanco:
Alexander collaborated on this PR, but I removed him as an official co-author to unblock the CLA check. Once he's authorized to contribute, I'll add him back as co-author.

nestoracunablanco:
@byako @pohly what do you think about this change?

demo/scripts/install-kubevirt-daily.sh
nojnhuh, line 1: Why do we need this script here if it's not called or documented anywhere?

nestoracunablanco: Hi @nojnhuh, the enf goal of this PR is to enable KubeVirt testing with the dra-example-driver. This is an initial step toward that functionality.
I prefer to keep PRs focused and incremental to make review easier. However, if you'd prefer, I can consolidate these changes into a larger PR.
I'll document the new functionality in the README.

pohly: > if you'd prefer, I can consolidate these changes into a larger PR.

We don't want to merge A in one PR if we then don't follow up with merging B, or the review of B affects A. Separate commits in a single PR work better in such a situation: a reviewer can still see the individual steps if they want.

nestoracunablanco: Perfect, will drop the PR and prepare a new one with different commits for this use-case. Will take some time, due to another priorities right now.

PR #217, opened 2026-06-03
Test on Kubernetes 1.36.1

nojnhuh:
A 1.36.1 pre-built kind node image is available which should save our CI from building a 1.36.0 image itself. A pre-built image is also required for systems with a container runtime other than Docker. The new kind images require the latest [kind v0.32.0 release](https://github.com/kubernetes-sigs/kind/releases/tag/v0.32.0).


willie-yao (approved):
/lgtm

Issue #91, opened 2025-04-03
Add Feature Gates to Manage dra-example-driver Compatibility


bg-chun:
In this issue, we discuss whether we should introduce feature gates for the dra-example-driver.

The core DRA API (Structured Parameters) will become GA in version 1.34 and will be enabled by default. Additionally, DRA features in beta will also start being enabled by default.

The Beta features in version 1.34 related to the dra-example-device are:
- DRA: Partitionable Devices
- DRA: Prioritized Alternatives in Device Requests

Even if the dra-example-driver implements these two features, they may not function properly in clusters where these features are not supported or enabled. For instance:
- Version 1.32 clusters only support core DRA as a beta feature.
- Version 1.33 clusters have core DRA in beta, while Partitionable Devices and Prioritized Alternatives are alpha features and may not enabled.

Furthermore, more DRA-related features will continue to be added upstream. Users installing the latest dra-example-driver's helm charts and images may encounter errors due to unsupported features in their cluster environments.

Should we introduce a kind of feature gates in the dra-example-driver to manage compatibility across different Kubernetes releases?

Here is a simple example illustrating this concept:
| DRA Feature               | dra-example-driver v0.1.0 | dra-example-driver v0.2.0 | dra-example-driver v0.3.0 |
|---------------------------|---------------------------|---------------------------|---------------------------|
| **Compatible K8s version**| v1.32                     | v1.33                     | v1.34                     |
| Partitionable Devices API | Disabled                  | Disabled                  | Enabled                   |
| Prioritized Alternatives  | Disabled                  | Disabled                  | Enabled                   |

bg-chun:
cc @nojnhuh

pohly:
Prioritized Alternatives has no impact on the driver, or no impact that would need a feature gate (it might have to handle sub-request names, but I forgot the details - we might handle it for the driver in the helper).

Usage of partititionable devices would be covered by what hardware the driver is being asked to simulate, which is better handled by a separate flag than a feature gate.

The bigger question is: which API version should the driver and helper package use? We kept using v1beta1 in 1.33.

**Pro:** DRA drivers using that code, like the example driver, work with 1.32.

**Con:** they are less compatible with future Kubernetes releases, because v1beta1 gets removed one release earlier than v1beta2.

If we keep doing this right until v1beta1 really gets removed, then all DRA drivers will have to be updated before users can upgrade to that new release (no transition period!).

[...]

PR #220, opened 2026-06-11
Add support for featureGates

Sreeja1725:
Adds a shared feature gate support and uses VFIOGPUEnabled to select the active driver profile at runtime.

### DIFF
--- a/cmd/dra-example-kubeletplugin/main.go
+++ b/cmd/dra-example-kubeletplugin/main.go
@@ -33,6 +33,7 @@ import (
 
 	"sigs.k8s.io/dra-example-driver/internal/profiles"
 	"sigs.k8s.io/dra-example-driver/internal/profiles/gpu"
+	"sigs.k8s.io/dra-example-driver/pkg/featuregates"
 	"sigs.k8s.io/dra-example-driver/pkg/flags"
 )
 
@@ -50,7 +51,6 @@ type Flags struct {
 	kubeletRegistrarDirectoryPath string
 	kubeletPluginsDirectoryPath   string
 	healthcheckPort               int
-	profile                       string
 	driverName                    string
 	podUID                        string
 	gpuPartitions                 int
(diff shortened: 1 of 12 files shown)


pkg/featuregates/featuregates.go
```diff
@@ -0,0 +1,93 @@
...
+// featureGateEmulationVersion must use Kubernetes-style versions (major.minor
+// matching the vendored k8s.io/* release), not driver SemVer. Driver-local
+// gates use 0.x Version fields and remain visible because 0.x < 1.y.
+var featureGateEmulationVersion = version.MajorMinor(1, 36)
+
+const (
+	// VFIOGPUEnabled enables PCI passthrough device discovery and preparation via
+	// the vfio-gpu profile. When enabled, the driver selects that profile
+	// instead of the default gpu profile.
+	VFIOGPUEnabled featuregate.Feature = "VFIOGPUEnabled"
```
alaypatel07, line 37: This is bleeding in from the other PR? Should this be left out?

cmd/dra-example-kubeletplugin/main.go
```diff
@@ -138,16 +138,9 @@ func newApp() *cli.App {
 			Destination: &flags.healthcheckPort,
 			EnvVars:     []string{"HEALTHCHECK_PORT"},
 		},
-		&cli.StringFlag{
```
alaypatel07, line 141: we can keep this flag right?

we should document with vfio that the profile will be changed to vfio automatically

alaypatel07 (changes requested):
Adding feature gate is a good feature for the example driver, it will help everyone follow the good practices.

Please have a very tight scoped change of introducing the feature gate package. You can leave an example feature gate in comment for documentation and remove vfio traces from this PR

alaypatel07:
cc @nojnhuh @pohly

nojnhuh:
I think we've mostly decided against the "feature gate" form factor in #91. Device profiles, command line flags, and Helm values should be enough to enable/disable certain features.

alaypatel07:
@nojnhuh @pohly we have a community dra driver that uses feature gates: https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/tree/main/pkg/featuregates

The example dra driver is supposed to be an example of how to implement the dra driver. For features like vfio driver and enabling discoverable metadata, KEP-5304, the dra-driver-nvidia-gpu uses feature gates.

Is there any value in aligning both of them either to profile or feature-gate?

My main reason for having feature gates was because inspired from the DRA Driver for NVIDIA GPUs.

nojnhuh:
The mechanism to toggle features isn't an integral part of how DRA drivers _should_ be implemented IMO. It's ok if different drivers do that in different ways. If anything, the example driver and the NVIDIA driver doing different things can help reinforce that and empower other drivers to do what works for them.

pohly:
I agree with @nojnhuh. The example driver has to support different "personas". I think profiles work well for that. A production driver may only have a single persona. Then boolean flags to toggle something can make sense. Those don't have to be feature gates, a plain boolean flag or config option can be better choice because those can be associated with a dedicated help text.

PR #222, opened 2026-06-12
Manage tools with Go modules

### LINKED ISSUE #192: Improve tool dependency management
This repo uses several third-party tools like helm, kubectl, golangci-lint, and others which are mostly installed ad hoc. Currently a mix of inline `go run`/`go install`, system package managers, and a custom Docker image are used. All(?) of our tools are Go-based, so a more unified approach should be possible.
https://github.com/kubernetes-sigs/dra-example-driver/blob/a39c14cc6e976e3c550a209266184665c029b802/.github/workflows/e2e.yaml#L22-L23
https://github.com/kubernetes-sigs/dra-example-driver/blob/a39c14cc6e976e3c550a209266184665c029b802/Makefile#L19
https://github.com/kubernetes-sigs/dra-example-driver/blob/a39c14cc6e976e3c550a209266184665c029b802/Makefile#L94
https://github.com/kubernetes-sigs/dra-example-driver/blob/a39c14cc6e976e3c550a209266184665c029b802/docker/Dockerfile.devel#L17-L22

Ideally, the pattern we establish here will:
- install tools on-demand at pinned versions
- enable dependabot to automate updates

Go 1.24 introduced the `tool` directive in go.mod to help with this. _We should not use it here._ At least not in [the way it's meant to be used](https://go.dev/doc/modules/managing-dependencies#tools). Tools in go.mod share dependencies with each other and the rest of the module, so a single version of each module dependency must be compatible with every module that depends on it. This is already the case with the existing set of go.mod dependencies, but the problem is exacerbated when other, independent executables with their own dependency graphs are in the mix. We should be able to build and run those tools with the exact versions of the dependencies that they already compile and test against without affecting our own dependencies compiled into the driver.

One approach I've thought about is to continue to leverage Go modules, but define a separate module for each tool. https://github.com/kubernetes-sigs/dra-driver-topology/pull/5 is one implementation of that idea.

nojnhuh:
This change implements the approach described in #192 to manage tool dependencies like `golangci-lint`, `helm`, and `conversion-gen` among others:
- `hack/tools/go.{mod,sum}` record tool dependencies
  - Add a new dependency
    -  `go get -modfile hack/tools/go.mod -tool <module path>`
      - e.g. `go get -modfile hack/tools/go.mod -tool helm.sh/helm/v4/cmd/helm@v4.2.0`
  - Run a tool
    - `go tool -modfile hack/tools/go.mod <tool>`
      - e.g. `go tool -modfile hack/tools/go.mod kind version`
- `hack/tools/golangci-lint/go.{mod,sum}` is a module only for `golangci-lint` since they claim to be particularly sensitive to dependency versions.
  - `make lint` invokes the tool version.

Notably, building and running the driver locally may not have depended on a local Go installation before, but now it does. Running all of the scripts in a dev container with Go installed probably still works, but I'm open to fixing issues.

Fixes #192


hack/tools/golangci-lint/go.mod
```diff
@@ -0,0 +1,220 @@
...
+	github.com/go-xmlfmt/xmlfmt v1.1.3 // indirect
+	github.com/gobwas/glob v0.2.3 // indirect
+	github.com/godoc-lint/godoc-lint v0.11.2 // indirect
+	github.com/gofrs/flock v0.13.0 // indirect
+	github.com/golang/protobuf v1.5.3 // indirect
+	github.com/golangci/asciicheck v0.5.0 // indirect
+	github.com/golangci/dupl v0.0.0-20260401084720-c99c5cf5c202 // indirect
+	github.com/golangci/go-printf-func-name v0.1.1 // indirect
+	github.com/golangci/gofmt v0.0.0-20250106114630-d62b90e6713d // indirect
+	github.com/golangci/golangci-lint/v2 v2.12.1 // indirect
```
nojnhuh, line 90: This is intentionally out of date to make sure dependabot is wired up correctly for this new module.

demo/scripts/build-driver-image.sh
```diff
@@ -45,6 +45,6 @@ export VERSION="${DRIVER_IMAGE_TAG}"
 export CONTAINER_TOOL="${CONTAINER_TOOL}"
 
 # Regenerate the CRDs and build the container image locally (demo / kind workflow).
-make docker-generate
+make generate
```
harche, line 48: The sibling release script still calls the `docker-generate` target this PR removes, so `make push-release-artifacts` now fails with `No rule to make target 'docker-generate'` (reproduced locally; it aborts before pushing anything):

https://github.com/nojnhuh/dra-example-driver/blob/446b301e8c53e4c5ed445ead96b570e22a8e3e99/demo/scripts/push-driver-image.sh#L37-L38

It probably just needs the same change as here, plus a comment update.

test/e2e/setup-e2e.sh
```diff
@@ -30,7 +30,7 @@ if [[ "${HELM_CHART_PATH}" != oci://* ]]; then
 fi
 bash demo/clusters/kind/create-cluster.sh
 
-helm upgrade -i \
+go tool -modfile hack/tools/go.mod helm upgrade -i \
```
harche, line 33: Nit: the GKE scripts still use a system `helm`, while the kind/e2e path now uses the pinned one. Was that intentional for GKE? If not, the same `go tool` invocation would keep them consistent:

https://github.com/nojnhuh/dra-example-driver/blob/446b301e8c53e4c5ed445ead96b570e22a8e3e99/demo/clusters/gke/install-driver.sh#L24

https://github.com/nojnhuh/dra-example-driver/blob/446b301e8c53e4c5ed445ead96b570e22a8e3e99/demo/clusters/gke/uninstall-driver.sh#L22

README.md
```diff
@@ -22,6 +22,7 @@ The procedure below has been tested and verified on both Linux and Mac.
 
 ### Prerequisites
 
+* [Go 1.26+](https://go.dev)
```
harche, line 25: Nit: kind and helm now come from `hack/tools/go.mod`, so they're no longer prerequisites for the demo scripts. Could drop these two lines (kubectl is still needed):

https://github.com/nojnhuh/dra-example-driver/blob/446b301e8c53e4c5ed445ead96b570e22a8e3e99/README.md#L29-L30

hack/tools/golangci-lint/go.mod
```diff
@@ -0,0 +1,219 @@
+module sigs.k8s.io/dra-example-driver/tools/golangci-lint
+
+go 1.26.4
```
harche, line 3: Nit: this module says `go 1.26.4` while the root and `hack/tools` modules say `1.26.0` and CI pins `1.26.2`. With `GOTOOLCHAIN=auto`, anyone on 1.26.2 running `make lint` triggers a toolchain download. I confirmed with `GOTOOLCHAIN=go1.26.2+auto go -C hack/tools/golangci-lint version` -> `go1.26.4`. Worth aligning to one version?

harche (changes requested):
Thanks

demo/scripts/build-driver-image.sh
nojnhuh: Thanks, double-checked that there aren't any other `docker-*` targets mentioned anymore.

test/e2e/setup-e2e.sh
nojnhuh: I didn't add it there at first because we don't run those scripts in CI. Since we provide `helm` though, we might as well use it.

README.md
nojnhuh: Updated.

hack/tools/golangci-lint/go.mod
nojnhuh: The `go` directive is mainly to communicate compatibility. IMO it doesn't make sense to bump `go` in the module for the driver components only because the linter requires a different version.

Using a newer toolchain in CI than go.mod is deliberate to take advantage of any fixes or improvements to Go. Ideally we'd test with a variety of versions, but we only publish images built with a single version of Go so supporting custom-built images with other versions of Go is going to be limited and best-effort.

1.26.2 isn't the latest patch and should be updated, but that should be done separately. Not a release blocker IMO.

test/e2e/setup-e2e.sh
harche: Thanks for updating the GKE scripts! Could we double-check the source path and Helm invocation? I ran a couple of checks against `8223db87fb85` in a Linux container with Bash **5.2.37**, using scripts downloaded directly from the latest PR head:

- After confirming the PR’s Git tree has no `demo/clusters/gke/common.sh`, I ran the unmodified install/uninstall scripts in a temporary directory. Both failed at line 22:
  ```text
  /repro/install-driver.sh: line 22: /repro/common.sh: No such file or directory
  exit status: 1

  /repro/uninstall-driver.sh: line 22: /repro/common.sh: No such file or directory
  exit status: 1
  ```
  The shared file appears to be `demo/scripts/common.sh`.

- Separately, I checked the quoting behavior using the same command-string form as the shared script:
  ```bash
  HELM="go tool -modfile /example/hack/tools/go.mod helm"
  "${HELM}" version
  ```
  That exited **127** because Bash treats the whole string as one executable name. Could we use a direct `go tool` invocation or a Bash array instead?

These were shell-level checks with networking disabled, not a full GKE deployment. Since these scripts aren’t exercised in CI, would a small Bats smoke test with mocked external commands be useful to cover both paths?

harche: One more thing that surfaces once the `source` path is fixed. The GKE scripts run with `set -u`:

https://github.com/nojnhuh/dra-example-driver/blob/8223db87fb85/demo/clusters/gke/install-driver.sh#L20

but `demo/scripts/common.sh` reads `CONTAINER_TOOL` without a default, so sourcing it aborts before reaching helm:

https://github.com/nojnhuh/dra-example-driver/blob/8223db87fb85/demo/scripts/common.sh#L66

```text
demo/scripts/common.sh: line 66: CONTAINER_TOOL: unbound variable
```

The kind scripts don't use `-u`, which is why they're unaffected:

https://github.com/nojnhuh/dra-example-driver/blob/8223db87fb85/demo/clusters/kind/create-cluster.sh#L22-L23

Either `${CONTAINER_TOOL:-}` in `common.sh` or dropping `-u` from the GKE scripts would fix it. Reproduced locally by fixing only the source path and running `uninstall-driver.sh`; with `CONTAINER_TOOL=docker` preset it gets past this and hits the `"${HELM}"` issue above.

nojnhuh: Thanks, should be fixed now.

PR #225, opened 2026-06-16
Debug cert-manager slow install

nojnhuh:


nojnhuh:
Mystery solved: regression between Helm v4.2.0 and v4.2.1: https://github.com/helm/helm/issues/32214

#222 should make this kind of thing easier to pin down and control going forward.

/close

PR #231, opened 2026-06-24
Add claimName parameter to ApplyConfig interface

nirdothan:
This change adds the ResourceClaim name as a parameter to the ApplyConfig method in the Profile interface. This enables profiles to generate claim-specific CDI configurations, such as unique mount paths or claim-specific identifiers.

Use case: Profiles that need to create per-claim resources (directories, device configurations, etc.) can use the claim name to ensure uniqueness and avoid conflicts when multiple claims use the same device type.

Example: A network profile that creates per-claim directories:
  /var/run/devices/<claim-name>/device-0/

Changes:
- Add claimName parameter to ConfigHandler.ApplyConfig interface
- Update GPU profile to accept (but not use) claimName parameter
- Update CPU profile to accept (but not use) claimName parameter
- Update state.go to pass claim.Name to ApplyConfig
- Update CPU test to pass dummy claimName

All existing profiles continue to work unchanged - they simply ignore the claimName parameter if not needed.


vishalanarase (commented):
@nirdothan  needs cncf-cla signed

PR #245, opened 2026-07-14
fix: bypass kind load to fix containerd config version 4 error

nestoracunablanco:
kind v0.31+ internally generated a containerd config version 4 when running `kid load image-archive`, but containerd v2.x inside kindest/node:v1.36.1 only supports versions 2 and 3, causing the following error:

unknown containerd config version: 4 (supported versions: 2 and 3)

The call is replaced with a loop that iterates over each cluster node and pipes the image image tar directly into the node's container.

Assisted-by: IBM Bob


nojnhuh (commented):
/lgtm
/approve

Issue #240, opened 2026-07-09
Unprepare panics with a nil pointer dereference when the checkpoint is unreadable or corrupt


thc1006:
### What happened?

`DeviceState.Unprepare` panics with a nil pointer dereference when the on-disk checkpoint exists but cannot be read or decoded.

`readCheckpoint` returns `(nil, err)` on any `ReadFile` error other than not-exist, and on any decode error (`checkpoint.go`). In `Unprepare` the error branch writes a fresh checkpoint to disk but does not assign it to the local `checkpoint` variable, which is still `nil`:

```go
checkpoint, err := readCheckpoint(s.checkpointPath, s.checkpointDecoder)
if err != nil {
    if err := writeCheckpoint(s.checkpointPath, s.checkpointEncoder, new(checkpointapi.Checkpoint)); err != nil {
        return fmt.Errorf("unable to create new checkpoint: %v", err)
    }
}
// checkpoint is still nil here
if err := s.unprepareDevices(claimUID, checkpoint); err != nil { ... }
s.removeClaimFromCheckpoint(checkpoint, claimUID)
```

`removeClaimFromCheckpoint` then dereferences `checkpoint.PreparedClaims`, so the kubelet plugin's `NodeUnprepareResources` handler panics and the DaemonSet pod crashes. The panic is on the exact path meant to recover from a bad checkpoint, so the intended self-heal never runs; it also happens before `DeleteClaimSpecFile`, so the claim's CDI spec is left orphaned in the CDI directory.

### What did you expect to happen?

On an unreadable or corrupt checkpoint, `Unprepare` should recover (reset to an empty checkpoint) and continue, not panic.

### How can we reproduce it?

Write invalid bytes to the plugin's `checkpoint.json`, then call `Unprepare` (or trigger `NodeUnprepareResources`). It panics with `runtime error: invalid memory address or nil pointer dereference` in `removeClaimFromCheckpoint`, from `Unprepare`.

### Suggested fix

Assign the fresh checkpoint to the local variable in the error branch:

```go
if err != nil {
    checkpoint = new(checkpointapi.Checkpoint)
    if err := writeCheckpoint(s.checkpointPath, s.checkpointEncoder, checkpoint); err != nil {
        return fmt.Errorf("unable to create new checkpoint: %v", err)
    }
}
```

[...]

PR #248, opened 2026-07-24
Fix panic in UnprepareResourceClaims when checkpoint file is unreadab…

vikasbolla:
Fixes - https://github.com/kubernetes-sigs/dra-example-driver/issues/240.
So this change assigns the newly created checkpoint file to a local variable, which prevents the panic when this file is used in later part of the `unprepare()`.  And I have also added a test case to verify this particular case.

### DIFF
--- a/cmd/dra-example-kubeletplugin/state.go
+++ b/cmd/dra-example-kubeletplugin/state.go
@@ -195,9 +195,12 @@ func (s *DeviceState) Unprepare(claimUID types.UID) error {
 
 	checkpoint, err := readCheckpoint(s.checkpointPath, s.checkpointDecoder)
 	if err != nil {
-		if err := writeCheckpoint(s.checkpointPath, s.checkpointEncoder, new(checkpointapi.Checkpoint)); err != nil {
-			return fmt.Errorf("unable to create new checkpoint: %v", err)
-		}
+		// The checkpoint file exists but cannot be read or decoded. Return the
+		// error and let the kubelet retry rather than attempting any recovery.
+		// Writing a new checkpoint here would silently discard records of all
+		// other currently-prepared claims, and we cannot safely act on an
+		// unknown state.
+		return fmt.Errorf("unable to read checkpoint: %w", err)
 	}
 
 	if err := s.unprepareDevices(claimUID, checkpoint); err != nil {
--- a/cmd/dra-example-kubeletplugin/state_test.go
+++ b/cmd/dra-example-kubeletplugin/state_test.go
@@ -17,6 +17,8 @@
 package main
 
 import (
+	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 	"testing"
@@ -237,4 +239,72 @@ func TestComputeDeviceConfigSharedDeviceContainerEdits(t *testing.T) {
 	}
 	assert.Equal(t, "1", consumedByShare["share-0"], "share-0 should keep its own consumed CPU edit")
 	assert.Equal(t, "3", consumedByShare["share-1"], "share-1 should keep its own consumed CPU edit")
+}
+
+// TestUnprepareReturnsErrorOnUnreadableCheckpoint verifies that Unprepare returns an error
+// (rather than panicking) when checkpoint.json exists but cannot be decoded.
+// The fix returns the error immediately without touching the checkpoint, letting
+// the kubelet retry once the operator has corrected the file.
+func TestUnprepareReturnsErrorOnUnreadableCheckpoint(t *testing.T) {
+	const (
+		nodeName   = "test-node"
+		driverName = "cpu.example.com"
+		claimUID   = "some-claim-uid"
+	)
+
+	// t.TempDir() is used for both CDI files and the checkpoint directory so
+	// the test stays fully self-contained and cleans up automatically.
+	tmpDir := t.TempDir()
+
+	flags := &Flags{
+		cdiRoot:                     tmpDir,
+		driverName:                  driverName,
+		profile:                     "cpu",
+		nodeName:                    nodeName,
+		cpuNUMANodes:                1,
+		cpusPerNUMANode:             4,
+		kubeletPluginsDirectoryPath: tmpDir,
+	}
+
+	state, err := NewDeviceState(&Config{
+		flags:   flags,
+		profile: cpu.NewProfile(nodeName, driverName, flags.cpuNUMANodes, flags.cpusPerNUMANode),
+	})
+	require.NoError(t, err)
+
+	// The plugin directory must exist before writing files (NewDeviceState does
+	// not create it; that is done by RunPlugin at startup).
+	pluginDir := filepath.Join(tmpDir, driverName)
+	require.NoError(t, os.MkdirAll(pluginDir, 0750))
+
+	// create a CDI spec file for the claim, simulating a claim that was fully
+	// prepared before the checkpoint was corrupted. After Unprepare returns an
+	// error the file must still be present — nothing should have been released.
+	cdiSpecPath := filepath.Join(tmpDir, "k8s.cpu.example.com-cpu-"+claimUID+"-0.yaml")
+	require.NoError(t, os.WriteFile(cdiSpecPath, []byte("placeholder"), 0600))
+
+	// Write garbage bytes to the checkpoint file so that readCheckpoint returns
+	// (nil, err) — the exact precondition that triggered the nil dereference.
+	corruptData := []byte("THIS IS NOT VALID CHECKPOINT JSON")
+	require.NoError(t, os.WriteFile(
+		filepath.Join(pluginDir, DriverPluginCheckpointFile),
+		corruptData,
+		0600,
+	))
+
+	// 1. Unprepare must return an error, not panic.
+	err = state.Unprepare(types.UID(claimUID))
+	require.Error(t, err, "Unprepare must return an error on an unreadable checkpoint")
+	assert.Contains(t, err.Error(), "unable to read checkpoint")
+
+	// 2. The corrupt file must be left exactly as it was — the driver must not
+	// overwrite it, preserving evidence and avoiding silent data loss.
+	data, readErr := os.ReadFile(state.checkpointPath)
+	require.NoError(t, readErr)
+	assert.Equal(t, corruptData, data, "corrupt checkpoint file must not be modified by Unprepare")
+
+	// 3. The CDI spec file for the claim must still be present — Unprepare
+	// must not release anything when it cannot read the checkpoint.
+	_, statErr := os.Stat(cdiSpecPath)
+	assert.NoError(t, statErr, "CDI spec file must still be present after Unprepare returns an error")
 }


vikasbolla:
@thc1006 Could you please take a look.

thc1006 (commented):
Thanks for taking this on, @vikasbolla, and for the comment spelling out the fail-open choice. The assignment to `checkpoint` fixes the original nil dereference and CI is green, so the core of the fix looks right to me. I left a few inline notes, mostly about letting the test prove the recovery actually finishes. I am not an approver on this repo, so please treat these as suggestions and the maintainers will have the final word.

cmd/dra-example-kubeletplugin/state_test.go
```diff
@@ -238,3 +240,55 @@ func TestComputeDeviceConfigSharedDeviceContainerEdits(t *testing.T) {
...
+	require.NoError(t, os.MkdirAll(pluginDir, 0750))
+	require.NoError(t, os.WriteFile(
+		filepath.Join(pluginDir, DriverPluginCheckpointFile),
+		[]byte("THIS IS NOT VALID CHECKPOINT JSON"),
+		0600,
+	))
+
+	require.NotPanics(t, func() {
+		_ = state.Unprepare(types.UID("some-claim-uid"))
+	})
```
thc1006: The test discards the return value and only checks that it did not panic. A version of `Unprepare` that rewrote the checkpoint and then returned an error, or returned before deleting the claim's CDI spec, would still pass this. Could it assert the success path directly? A panic would still fail the test, so this is strictly stronger:

```suggestion
	require.NoError(t, state.Unprepare(types.UID("some-claim-uid")))
```

One more step would make it airtight: create the CDI spec for the claim before you corrupt the checkpoint, then check that it is gone at the end. That proves the cleanup reached `DeleteClaimSpecFile`, rather than only that the checkpoint ended up empty.

cmd/dra-example-kubeletplugin/state_test.go
```diff
@@ -238,3 +240,55 @@ func TestComputeDeviceConfigSharedDeviceContainerEdits(t *testing.T) {
 	assert.Equal(t, "1", consumedByShare["share-0"], "share-0 should keep its own consumed CPU edit")
 	assert.Equal(t, "3", consumedByShare["share-1"], "share-1 should keep its own consumed CPU edit")
 }
+
+// TestUnprepareCorruptCheckpointPanics test calls Unprepare with a corrupt checkpoint file and expects it to
+// return an error (or succeed) without panicking.
+func TestUnprepareCorruptCheckpointPanics(t *testing.T) {
```
thc1006: Minor: the name reads as if a panic is expected, but the test now checks the opposite. `TestUnprepareRecoversFromCorruptCheckpoint` would line up with what it asserts. The `return an error (or succeed)` phrasing in the doc comment is also a bit loose. The issue asks for Unprepare to recover and finish, so `succeeds` is the behavior worth pinning down.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -195,7 +195,18 @@ func (s *DeviceState) Unprepare(claimUID types.UID) error {
...
+		// continue and release the requested claim.
+		//
+		// Deliberate fail-open: resetting drops tracking for every other
+		// currently-prepared claim on this node. That is acceptable here because
+		// unprepareDevices is a no-op — no real hardware is torn down. A
+		// production fork that performs real device teardown should instead
+		// fail-closed (return the error and let the kubelet retry) to avoid
+		// silently leaking other claims' hardware resources.
+		checkpoint = new(checkpointapi.Checkpoint)
+		if err := writeCheckpoint(s.checkpointPath, s.checkpointEncoder, checkpoint); err != nil {
```
thc1006: Optional, and happy to defer to the maintainers here: logging the original read error together with the checkpoint path and claim UID before the reset would help whoever hits this later understand why the checkpoint was cleared, rather than seeing an apparently normal unprepare. `klog` is already imported in this file. Separately, this empty checkpoint is written here and again at the end of the function; assigning it and letting the final `writeCheckpoint` commit it would drop the duplicate and narrow the window where the on-disk checkpoint is empty but cleanup has not run yet.

cmd/dra-example-kubeletplugin/state_test.go
vikasbolla: @thc1006 Thanks for the review. Yes it is good point.  I have modified the test case a bit according to the suggestions

cmd/dra-example-kubeletplugin/state_test.go
vikasbolla: Yes makes sense, changed it now. Thanks

cmd/dra-example-kubeletplugin/state.go
vikasbolla: I have added the log. And regarding the writing to new checkpoint, I am thinking that it better to keep both the writes as of now. Since the first write ensures the on-disk file is immediately valid. so when the process crashes between here and the end of the function, subsequent reads won't hit the corrupt data again.

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -195,7 +195,20 @@ func (s *DeviceState) Unprepare(claimUID types.UID) error {
...
+		// corrupted on disk). Reset to an empty checkpoint so Unprepare can
+		// continue and release the requested claim.
+		//
+		// Deliberate fail-open: resetting drops tracking for every other
+		// currently-prepared claim on this node. That is acceptable here because
+		// unprepareDevices is a no-op — no real hardware is torn down. A
+		// production fork that performs real device teardown should instead
+		// fail-closed (return the error and let the kubelet retry) to avoid
+		// silently leaking other claims' hardware resources.
+		klog.ErrorS(err, "Corrupt checkpoint reset to empty; all prepared-claim tracking on this node is lost",
```
thc1006: Optional: `readCheckpoint` returns an error for a file it can't read at all (any `os.ReadFile` failure other than not-exist), not only for a decode failure, so this recovery also runs for the "unreadable" case in the PR title. `Unreadable or corrupt checkpoint reset to empty` would describe what it actually handles. And if you move this line just below a successful `writeCheckpoint`, it won't report a reset that a write failure would otherwise have stopped.

thc1006 (commented):
The panic fix reads correctly, and the test now covers the whole recovery path (reset the checkpoint, delete the CDI spec, then re-read the empty checkpoint) rather than only asserting that nothing panics. That pins the behavior down well. One small optional note inline.

cmd/dra-example-kubeletplugin/state.go
vikasbolla: Done, I have modified it accordingly. Thank you

vikasbolla:
@nojnhuh Could you PTAL at this PR, when you have some time

thc1006:
Thanks for working through all of this, and sorry for the slow turnaround on my side.

I went back over the current state and everything I raised earlier is addressed: the test asserts the success path and proves the CDI spec file is gone, the name matches what it checks, and the log now names the unreadable case and sits after the write so a failed write cannot report a reset that did not happen.

I also traced the failure again to be sure the test pins the real thing rather than just the symptom. The panic comes from `removeClaimFromCheckpoint`, which does `checkpoint.PreparedClaims = slices.DeleteFunc(...)` on the pointer that `readCheckpoint` left nil. Assigning the empty checkpoint before that line is what closes it, and since your test corrupts the file so `readCheckpoint` returns an error, it fails on the old code and passes on the new one. That is the property worth having.

Your reasoning for keeping both writes stands up, and I am glad you pushed back rather than just taking the suggestion. Writing immediately means a crash between the reset and the end of `Unprepare` leaves a valid file behind instead of the corrupt one, which is the safer of the two failure modes for an example driver.

The comment explaining the fail-open choice is the part I would most want a reader to find, particularly the note that a fork doing real teardown should fail closed. That is the kind of thing someone copying this driver needs to see, and it is easy to leave out.

Splitting the deeper problem into #249 rather than growing this PR was the right call too. Rebuilding the checkpoint from the API and the CDI files on disk is a different design question and deserves its own discussion.

CI is green on e2e, lint, and both test matrix entries.

/lgtm

This still needs an approver from OWNERS for the `approved` label. @nojnhuh, would you mind taking a look when you have a moment?

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -195,9 +195,22 @@ func (s *DeviceState) Unprepare(claimUID types.UID) error {
...
+		// continue and release the requested claim.
+		//
+		// Deliberate fail-open: resetting drops tracking for every other
+		// currently-prepared claim on this node. That is acceptable here because
+		// unprepareDevices is a no-op — no real hardware is torn down. A
+		// production fork that performs real device teardown should instead
+		// fail-closed (return the error and let the kubelet retry) to avoid
+		// silently leaking other claims' hardware resources.
+		checkpoint = new(checkpointapi.Checkpoint)
+		if err := writeCheckpoint(s.checkpointPath, s.checkpointEncoder, checkpoint); err != nil {
```
nojnhuh: Since #148, I don't think we should write a new checkpoint in this case at all anymore. Could we simply return the error (wrapped if that's useful) from `readCheckpoint` here?

That won't fix the next time `Unprepare` gets called, but I think that's the behavior we want. The driver shouldn't do anything potentially harmful when it's already in an unknown state.

thc1006: Agreed, and this is a better fix than the one I signed off on. My `/lgtm` above is superseded.

The reasoning holds up when you look at what #148 changed. Checkpoint writes go through a temporary file and an atomic rename now, so the driver can no longer leave a half-written file of its own. A read that fails therefore means something outside the driver touched the file or the disk is unhappy, and resetting on that assumption throws away tracking for every other prepared claim on the node to recover from a cause we have not established.

@vikasbolla, the good news is that returning the error still closes #240. The panic came from `removeClaimFromCheckpoint`, which does `checkpoint.PreparedClaims = slices.DeleteFunc(...)` on the pointer `readCheckpoint` left nil. Returning before that line means it is never reached, so the fix gets smaller rather than larger:

```go
checkpoint, err := readCheckpoint(s.checkpointPath, s.checkpointDecoder)
if err != nil {
	return fmt.Errorf("unable to read checkpoint: %w", err)
}
```

The test needs to turn around with it. It currently asserts that `Unprepare` succeeds and that the CDI spec file is gone; under this behaviour neither happens, because the function returns before `DeleteClaimSpecFile`. What is worth pinning instead:

- `Unprepare` returns an error rather than panicking, which is still the regression guard for #240;
- the corrupt checkpoint file is left exactly as it was, since not overwriting it is the point;
- the CDI spec file for the claim is still present, so nothing was released on a state we could not read.

`TestUnprepareReturnsErrorOnUnreadableCheckpoint` would describe that better than the current name.

Worth keeping a short comment on the early return saying why it does not repair the file, since the reason lives in #148 rather than in this function and the next reader will wonder.

vikasbolla: @nojnhuh Thank you for the explanation — that makes the reasoning much clearer. since the driver can no longer produce a half-written file itself, a read failure genuinely means something external happened, and resetting on that assumption might not be right approach as you mentioned. I have updated the PR accordingly, to just return the error in this case.

@thc1006 Thanks for the detailed reiteration as well. I have modified the PR along with the test to match the new behaviour.

nojnhuh (approved):
/lgtm
/approve

Thanks!

Issue #249, opened 2026-07-25
Checkpoint file becomes permanently stale after plugin pod restart when workload pods are already Running


vikasbolla:
During the fix for this issue https://github.com/kubernetes-sigs/dra-example-driver/issues/240. I have found that there is another issue which might need a different fix upon the separate discussion. So created a separate issue with required information.

**Problem:** 
After the plugin pod restarts, the kubelet does not re-send NodePrepareResources for pods that are already Running — it only does so for newly scheduled pods. It already has the CDI device IDs cached from before the restart. This means the checkpoint is permanently stale after a restart: claims prepared before the restart are missing from the new checkpoint even though the pods are running fine and their CDI files are still on disk.

```
1. Clone the Dra-example-driver repository 
2.  Install example driver
	helm upgrade -i \
  	--create-namespace \
 	 --namespace dra-example-driver \
  	dra-example-driver \
  	deployments/helm/dra-example-driver

3. Once driver is installed you should be able to see resource slices and device class.
4. Kubectl get pods -n dra-example-driver to check the kubelet plugin pods
5. Create resourceclaimtemplate and pods using existing templates from dra-example-driver repo(kubectl apply -f demo/examples/basic-resourceclaimtemplate/basic-resourceclaimtemplate.yaml)
6. Kubectl get resourceclaims -n basic-resourceclaimtemplate —> to check the resource claims
7. Exec into kubelet plugin pod using kubectl exec -it -n dra-example-driver pod-name -- bash
8. Cat /var/lib/kubelet/plugins/gpu.example.com/checkpoint.json.
9. corrupt the  checkpoint.json file.
10. Then exit from plugin pod and delete pod0 of basic-resourceclaimtemplate namespace ( parallelly in the. Other terminal check the log using kubectl logs -n dra-example-driver  pod-name -f)
11. Panics and we see that the kubelet plugin restarts. If you go to plugin pod and see the checkpoint.json file again we see that prepared claims will be nil. we see that pod1-claim missing even though pod1 is Running
```
**Proposed Solution:**
At plugin startup in NewDeviceState(), adding a one-time sync that rebuilds the checkpoint from two sources of truth:

1. **The Kubernetes API** — list all ResourceClaims across all namespaces that are allocated to this node by this driver .
2. **The CDI spec files on disk** — seems like, a claim is only considered "truly prepared" if both the API says it's allocated here AND the CDI file exists under - /var/run/cdi( inside plugin pod).

[...]

PR #253, opened 2026-08-06
Add comment to explain why resetting a corrupt checkpoint is unsafe f…

vikasbolla:
**What this PR does**
Adds an explanatory comment in `Unprepare()` next to the fail-closed checkpoint error path.

**Why**
While investigating a checkpoint-related panic https://github.com/kubernetes-sigs/dra-example-driver/pull/248, it came up that automatically resetting a corrupt checkpoint to empty might seem like a safe recovery — but it isn't, and the reason is subtle enough to trip up anyone forking this driver.

**The short version:** the kubelet does not re-call Prepare for already-running claims, only for new ones or after a node reboot. So if you wipe the checkpoint, it stays incomplete until the next reboot. For this example driver that is fine because `computeDeviceConfig` derives the device assignment purely from the claim spec — same input, same output, every time.

more info - https://github.com/kubernetes-sigs/dra-example-driver/issues/249.

### DIFF
--- a/cmd/dra-example-kubeletplugin/state.go
+++ b/cmd/dra-example-kubeletplugin/state.go
@@ -197,9 +197,17 @@ func (s *DeviceState) Unprepare(claimUID types.UID) error {
 	if err != nil {
 		// The checkpoint file exists but cannot be read or decoded. Return the
 		// error and let the kubelet retry rather than attempting any recovery.
-		// Writing a new checkpoint here would silently discard records of all
-		// other currently-prepared claims, and we cannot safely act on an
-		// unknown state.
+		//
+		// Deleting the file is not a way out: readCheckpoint treats a missing file
+		// as an empty checkpoint, so removing it writes off the records of every
+		// other prepared claim just as writing a new one would.
+		//
+		// This driver keeps only a prepared marker, because computeDeviceConfig
+		// rebuilds the CDI data from ResourceClaim.status.allocation and the
+		// devices this driver currently has. A fork that produces state during
+		// Prepare that it cannot rebuild afterwards, such as a MIG instance or a
+		// virtual function it later has to release, needs that state in the
+		// checkpoint for Unprepare to act on.
 		return fmt.Errorf("unable to read checkpoint: %w", err)
 	}
 


vikasbolla:
@nojnhuh @thc1006 , Could please take a look at this. Thanks!!

cmd/dra-example-kubeletplugin/state.go
```diff
@@ -197,9 +197,29 @@ func (s *DeviceState) Unprepare(claimUID types.UID) error {
...
+		// always returns the same result, so when Prepare is eventually re-called
+		// (e.g. after a node reboot) the driver recomputes the identical
+		// assignment and rewrites the same CDI spec file.
+		//
+		// A fork whose assignment logic consults runtime state — for example,
+		// picking whichever GPU is currently free — must not copy this reasoning.
+		// That decision was made at a specific point in time and cannot be
+		// replayed from the claim alone. Resetting the checkpoint would cause the
+		// driver to forget that a device is already in use and risk assigning it
+		// to a second claim while the original container is still running.
```
thc1006: Two things here I would change before this lands.

`readCheckpoint` returns an empty checkpoint when the file is missing: it ignores `fs.ErrNotExist` and only decodes when it actually read bytes. So `remove the corrupt checkpoint file` is the same operation as the reset the next paragraph warns against, and it drops the records of every other prepared claim, which is what #248 settled on avoiding.

`computeDeviceConfig` reads `claim.Status.Allocation` and `s.allocatable` rather than the claim spec, and the device was already chosen by the scheduler, so it rebuilds CDI data rather than deriving an assignment. That makes `picking whichever GPU is currently free` an awkward example, because a kubelet-side Prepare should not be choosing devices at all. The hazard worth naming is state a fork creates during Prepare that it cannot rebuild afterwards, such as a MIG instance or a virtual function, which Unprepare still has to release.

Smaller one in the same paragraph: that function groups its results in a map and ranges over it, and it also depends on the driver's current device inventory, so `always returns the same result` promises more than it can hold. Equivalent output for unchanged code and inventory is a claim it can keep.

Here is one way to fold those together, if it helps:

```suggestion
		// The checkpoint file exists but cannot be read or decoded. Return the
		// error and let the kubelet retry rather than recovering in the driver.
		// Deleting the file is not a way out: readCheckpoint treats a missing file
		// as an empty checkpoint, so removing it writes off the records of every
		// other prepared claim just as writing a new one would.
		//
		// This driver keeps only a prepared marker, because computeDeviceConfig
		// rebuilds the CDI data from ResourceClaim.status.allocation and the
		// devices this driver currently has. A fork that produces state during
		// Prepare that it cannot rebuild afterwards, such as a MIG instance or a
		// virtual function it later has to release, needs that state in the
		// checkpoint for Unprepare to act on.
```

thc1006 (commented):
Thanks for picking this up, and for going back to #249 for the context. The last paragraph is the part a fork author most needs, so I would like to see this land.

I left one inline comment with the two things I think are worth changing first, and a suggestion that folds them together.

(Disclosure: I used AI assistance while reviewing this. The two code references come from `readCheckpoint` in checkpoint.go and `computeDeviceConfig` on this head, which I read.)

cmd/dra-example-kubeletplugin/state.go
vikasbolla: Thanks for reading the actual code before commenting — really appreciate that.

I have rechecked the points you mentioned and got more clarity on what the comment should say, so I've applied your suggested replacement. The MIG/VF framing is more accurate than the GPU example I used — the real hazard is physical state created during `Prepare` that only the checkpoint remembers and that `Unprepare` still has to release. I also missed that deleting the checkpoint file produces the same empty checkpoint as a reset, so that advice was contradicting the very paragraph after it.

nojnhuh (commented):
/lgtm
/approve

PR #254, opened 2026-08-10
Add test for pod resources API- kep 3695

### LINKED ISSUE #186: Add example for PodResources API with DRA resources (KEP-3695)
KEP-3695 (https://github.com/kubernetes/enhancements/issues/3695) extends the kubelet PodResources gRPC API to include resources allocated by DRA drivers. This was Beta in 1.34 and reached Stable in 1.36.

The example driver should include a demo script or documentation showing how to query the PodResources API to see DRA-allocated devices for a running pod. This is useful for monitoring tools and device plugins that need to discover allocated DRA resources.

Shilpa-Gokul:
Fixes #186 

Adds an end-to-end test and accompanying demo that verifies the kubelet PodResources gRPC API correctly exposes DRA-allocated devices via the dynamicResources field introduced by [KEP-3695](https://github.com/kubernetes/enhancements/issues/3695).

Verification is done through a short-lived debug pod created on the same node as the workload. The debug pod uses two containers:

- curlimages/curl:latest (init container) — downloads api.proto from the Kubernetes v1.36.0 source into a shared EmptyDir volume. This is required because the kubelet gRPC server does not support server reflection.
- fullstorydev/grpcurl:latest (query container) — mounts the kubelet socket via a HostPath volume and calls v1.PodResourcesLister/List using the downloaded proto file via -import-path/-proto flags, running privileged as root to access the socket.

The response is captured from pod logs and asserted to contain the expected dynamicResources fields: pod name, container name, driver name, and claim name.

Assisted by: IBM Bob

### DIFF
--- a/test/e2e/e2e_setup_test.go
+++ b/test/e2e/e2e_setup_test.go
@@ -46,14 +46,19 @@ import (
 	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/utils/ptr"
 )
 
-var rootDir, demoManifestsDir string
-var clientset *kubernetes.Clientset
-var dynamicClient dynamic.Interface
-var restMapper meta.RESTMapper
+var (
+	config                    *rest.Config
+	rootDir, demoManifestsDir string
+	clientset                 *kubernetes.Clientset
+	dynamicClient             dynamic.Interface
+	restMapper                meta.RESTMapper
+)
 
 // driverPodSelector finds kubelet plugin Pods within an installed driver's release namespace.
 const driverPodSelector = "app.kubernetes.io/component=kubeletplugin"
@@ -113,7 +118,8 @@ var _ = BeforeSuite(func(ctx SpecContext) {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	configOverrides := &clientcmd.ConfigOverrides{}
 	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
-	config, err := kubeConfig.ClientConfig()
+	var err error
+	config, err = kubeConfig.ClientConfig()
 	Expect(err).NotTo(HaveOccurred())
 
 	clientset, err = kubernetes.NewForConfig(config)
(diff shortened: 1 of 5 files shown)


test/e2e/e2e_setup_test.go
```diff
@@ -994,3 +1000,131 @@ func verifyGPUConsumedCapacity(ctx context.Context, namespace, driverName string
...
+			HostNetwork:   true,
+			HostPID:       true,
+			RestartPolicy: v1.RestartPolicyNever,
+			InitContainers: []v1.Container{
+				{
+					Name:  "download-proto",
+					Image: "curlimages/curl:latest",
+					Command: []string{
+						"sh", "-c",
+						"mkdir -p /tmp/podresources && curl -sSL https://raw.githubusercontent.com/kubernetes/kubernetes/v1.36.0/staging/src/k8s.io/kubelet/pkg/apis/podresources/v1/api.proto -o /tmp/podresources/api.proto",
```
nojnhuh: Can we derive the version here from the running cluster instead of hardcoding `v1.36.0`?

Shilpa-Gokul: Thanks @nojnhuh. I have addressed it.

test/e2e/e2e_setup_test.go
```diff
@@ -994,3 +1000,135 @@ func verifyGPUConsumedCapacity(ctx context.Context, namespace, driverName string
...
+	Expect(err).NotTo(HaveOccurred(), "Failed to get pod %s/%s", namespace, podName)
+	nodeName := pod.Spec.NodeName
+	Expect(nodeName).NotTo(BeEmpty(), "Pod %s/%s has no node assigned", namespace, podName)
+
+	fmt.Fprintf(GinkgoWriter, "Querying PodResources API on node %s for pod %s/%s\n",
+		nodeName, namespace, podName)
+
+	serverVersion, err := clientset.Discovery().ServerVersion()
+	Expect(err).NotTo(HaveOccurred(), "Failed to get server version")
+	k8sVersion := fmt.Sprintf("v%s.%s.0", serverVersion.Major, serverVersion.Minor)
```
nojnhuh: Can you check to see if this works with #255 which uses v1.37.0-rc.0? I suspect it won't since the `v1.37.0` tag hasn't been created yet.

Shilpa-Gokul: @nojnhuh You are right, it didn't work for v1.37.0-rc.0. 
To fix it, I have changed the code to get the serverVersion.GitVersion instead which provides the full version.
Have tested with https://github.com/kubernetes-sigs/dra-example-driver/pull/255 changes as well and it works fine now.

nojnhuh (commented):
/lgtm
/approve

PR #255, opened 2026-08-10
Bump Kubernetes dependencies and test environment to 1.37.0-rc.0

harche:
This PR updates the example driver's dependencies and test environment to correspond with the Kubernetes 1.37.0-rc.0 release, and handles the collateral damage associated with that:
- Adapt to the `resource/v1` API change replacing `NodeAllocatableResourceMappings` with `NodeAllocatableResources`
- Implement the new `WatchHealthStatus` method required by the `kubeletplugin.DRAPlugin` interface, returning `ErrHealthNotSupported` for now (KEP-4680 device health reporting will be added in a follow-up)
- Fix a rest client metrics test to account for metrics now being registered lazily on first client construction

Since no `kindest/node` image is published for the rc yet, e2e clusters build the node image from the release artifacts. I plan to bump everything to v1.37.0 once it is available.


harche:
Keeping this in draft until Kubernetes v1.37.0 final is released. At that point I'll switch the go.mod dependencies, the kind node image tag, and the code-generator tool versions from v1.37.0-rc.0 to the final release and mark this ready for review.

harche:
@pohly should we wait for final 1.37 to get released or should we consider bumping up to 1.37 rc0 for now (and then bump again for 1.37)?

nojnhuh:
> @pohly should we wait for final 1.37 to get released or should we consider bumping up to 1.37 rc0 for now (and then bump again for 1.37)?

Let's wait for the final release.

harche:
> > @pohly should we wait for final 1.37 to get released or should we consider bumping up to 1.37 rc0 for now (and then bump again for 1.37)?
> 
> Let's wait for the final release.

thanks, I will close this one then. We can directly bump to 1.37 when it is cut. 

/close

PR #259, opened 2026-08-29
Recreate claim CDI specs when restoring prepared DRA claims

Gaurav598:
## What this PR does

When `DeviceState.Prepare()` restores a prepared claim from checkpoint, recreate
the transient claim-specific CDI spec before returning the restored CDI device IDs.

After a node/runtime restart, the checkpoint can survive while transient CDI
specs are lost. Previously, the restored prepare path could return the
checkpointed CDI IDs without recreating the corresponding CDI spec. This allowed
the prepare call to succeed while the container runtime could no longer resolve
the returned CDI devices.

This change makes restored `Prepare()` re-establish the claim CDI spec before
reporting success.

## Testing

Added regression coverage for:

- restored prepare when the claim CDI spec is missing
- idempotent restored prepare when the CDI spec already exists
- failure when the CDI spec cannot be recreated

Tests:

- `go test ./cmd/dra-example-kubeletplugin -count=1`
- `go test ./...`

## Related

Related to kubernetes/kubernetes#141433
Related to kubernetes/kubernetes#141619

### DIFF
--- a/cmd/dra-example-kubeletplugin/state.go
+++ b/cmd/dra-example-kubeletplugin/state.go
@@ -169,6 +169,9 @@ func (s *DeviceState) Prepare(ctx context.Context, claim *resourceapi.ResourceCl
 		return nil, fmt.Errorf("unable to restore from checkpoint: %v", err)
 	}
 	if restoredDevices != nil {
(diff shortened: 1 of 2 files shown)


nojnhuh:
/lgtm

based on @harche's [review](https://github.com/kubernetes-sigs/dra-example-driver/pull/259#issuecomment-5479821093).

PR #267, opened 2026-09-07
Merge device status into ResourceClaim status.devices instead of replacing it

### LINKED ISSUE #266: Kubelet plugin overwrites status.devices[].conditions set by the BindingConditions controller
### What happened?

When `gpuDeviceStatus` (KEP-4817 device status, added in #128) and binding conditions (KEP-5007, added in #199) are both enabled, the kubelet plugin's `NodePrepareResources` overwrites the `BindingConditions=True` condition that the `BindingConditions` controller plugin wrote into `ResourceClaim.status.devices[]`.

The controller watches ResourceClaims, notices the condition is gone, and re-adds it. The end state is correct, but every prepare causes a wipe-and-restore round trip and a transient revision where the condition is absent.

Observed claim revisions for a single pod (summary of `status.devices` per watch event):

```
rv=1527  allocated  devices=(none)
rv=1529  allocated  gpu-0{conditions=BindingConditions=True  data=false}   # controller sets condition
rv=1540  allocated  gpu-0{conditions=(none)                  data=true}    # kubelet plugin prepare wipes it
rv=1541  allocated  gpu-0{conditions=BindingConditions=True  data=true}    # controller re-adds it
```

<details>
<summary>Where this lives in the code</summary>

`updateDeviceStatus` in `cmd/dra-example-kubeletplugin/state.go` replaces the whole `status.devices` slice:

```go
claim = claim.DeepCopy()
claim.Status.Devices = devices
_, err = rc.UpdateStatus(ctx, claim, metav1.UpdateOptions{})
```

The entries it writes come from `BuildDeviceStatus` in `internal/profiles/gpu/gpu.go`, which only populates `Device`, `Driver`, `Pool`, and `Data`. `Conditions` is always nil, so anything the controller set is dropped.

</details>

<details>
<summary>Controller and kubelet plugin logs</summary>

Controller logs `Set binding condition` twice for the same claim. The kubelet plugin's `Publishing device status to ResourceClaim` lands between them.

```
12:49:03Z  controller     Set binding condition   claim=pod0-gpu-wqc25 device=gpu-0 condition=BindingConditions
12:49:03Z  controller     Updating ResourceClaim status
12:49:09Z  kubeletplugin  Publishing device status to ResourceClaim  method=NodePrepareResources claim=pod0-gpu-wqc25 devices=1
12:49:09Z  controller     Set binding condition   claim=pod0-gpu-wqc25 device=gpu-0 condition=BindingConditions
12:49:09Z  controller     Updating ResourceClaim status
```

</details>

### What did you expect to happen?

The kubelet plugin should merge its `data` into the existing per-device entry (matched on `driver`, `pool`, `device`) and leave `conditions` and any other fields untouched. The controller side already does this correctly in `setDeviceCondition` in `cmd/dra-example-controller/plugins/bindingconditions.go`.

### How can we reproduce it (as minimally and precisely as possible)?

Install the driver with both features on, then run the existing binding-conditions example while watching the claim:

```
helm upgrade -i --create-namespace --namespace dra-example-driver \
  --set gpuDeviceStatus=true \
  --set kubeletPlugin.bindingConditions=true \
  --set 'controller.plugins={BindingConditions}' \
  dra-example-driver deployments/helm/dra-example-driver

kubectl apply -f demo/examples/binding-conditions/binding-conditions.yaml
```

Watch for the revision where `conditions=` is empty and `data=true`, followed by one where the condition returns.

<details>
<summary>Full steps, including cluster creation and the watch command</summary>

kind cluster from `demo/scripts/kind-cluster-config.yaml` (feature gate `DRADeviceBindingConditions` is already on there), driver built from `3d2e975`:

```
bash demo/build-driver.sh
bash demo/clusters/kind/create-cluster.sh

helm upgrade -i --create-namespace --namespace dra-example-driver \
  --set gpuDeviceStatus=true \
  --set kubeletPlugin.bindingConditions=true \
  --set 'controller.plugins={BindingConditions}' \
  dra-example-driver deployments/helm/dra-example-driver

kubectl create ns binding-conditions
kubectl get resourceclaim -n binding-conditions -w -o json --output-watch-events \
  | jq -rc --unbuffered '.object as $o | "rv=\($o.metadata.resourceVersion) " + (($o.status.devices // []) | map("\(.device){conditions=\(.conditions // [] | map(.type + "=" + .status) | join(",")) data=\(.data != null)}") | join(" "))' &

kubectl apply -f demo/examples/binding-conditions/binding-conditions.yaml
```

</details>

### Anything else we need to know?

- Benign today: the scheduler has already consumed the condition before prepare runs, and the controller restores it within one reconcile.
- The transient window is where a second pod sharing the same claim could observe the condition missing.
- Any other writer of `status.devices` (a future `networkData` or a health condition from the kubelet plugin) would be clobbered the same way.
- Neither feature's e2e test enables the other, so the combination is not exercised in CI. A follow-up e2e case with both flags on could assert the condition is never absent once set.

### Kubernetes version

```
Client Version: v1.37.0
Server Version: v1.37.0 (kind, kindest/node:v1.37.0)
```

### Driver version

Built from `main` at `3d2e975` (chart appVersion v0.4.0).

harche:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When the kubelet plugin publishes per-device status (`--gpu-device-status`), `updateDeviceStatus` replaced the whole `ResourceClaim.status.devices` slice with entries that carry only `data`. That dropped the `BindingConditions=True` condition written earlier by the `BindingConditions` controller plugin, which then had to reconcile again to restore it.

This change merges the kubelet plugin's status into the existing per-device entries instead. Entries are matched on `driver`, `pool`, `device` and `shareID`; only `data` is replaced on a match, and unmatched entries are appended. Fields written by other actors, such as `conditions`, are left untouched.

Before, with both features enabled, one pod produced these claim revisions:

```
rv=1529  gpu-0{conditions=BindingConditions=True  data=false}   # controller sets condition
rv=1540  gpu-0{conditions=(none)                  data=true}    # kubelet plugin prepare wipes it
rv=1541  gpu-0{conditions=BindingConditions=True  data=true}    # controller re-adds it
```

After:

```
rv=2790  gpu-0{conditions=BindingConditions=True  data=false}   # controller sets condition
rv=2804  gpu-0{conditions=BindingConditions=True  data=true}    # kubelet plugin adds data, condition kept
```

#### Which issue(s) this PR fixes:

Fixes #266

#### Special notes for your reviewer:

Added unit tests for the merge helper:

- existing conditions are preserved when `data` is merged in, and unmatched devices are appended
- entries that differ only by `shareID` are treated as distinct

Tests run:

- `go test ./cmd/dra-example-kubeletplugin -count=1`
- `go test ./...`

Verified on a kind v1.37.0 cluster with `gpuDeviceStatus=true`, `kubeletPlugin.bindingConditions=true` and `controller.plugins={BindingConditions}`, running `demo/examples/binding-conditions`: the controller now logs `Set binding condition` once per claim instead of twice, and the final `status.devices` entry carries both the condition and the data.

AI usage disclosure: AI assistance (Claude Code) was used for testing and verifying the results, namely reproducing the issue on a kind cluster, capturing the ResourceClaim revisions before and after the change, and running the test suite.

#### Does this PR introduce a user-facing change?

```release-note
Fixed the kubelet plugin overwriting `ResourceClaim.status.devices[].conditions` (for example the BindingConditions condition set by the controller) when publishing device status data.
```

### DIFF
--- a/cmd/dra-example-kubeletplugin/state.go
+++ b/cmd/dra-example-kubeletplugin/state.go
@@ -37,6 +37,7 @@ import (
 
 	"k8s.io/klog/v2"
 	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1"
+	"k8s.io/utils/ptr"
 	cdiapi "tags.cncf.io/container-device-interface/pkg/cdi"
 
 	checkpointapi "sigs.k8s.io/dra-example-driver/internal/api/checkpoint"
@@ -506,9 +507,29 @@ func (s *DeviceState) updateDeviceStatus(ctx context.Context, ns, name string, d
 
 		// copy the object and update only status.devices
 		claim = claim.DeepCopy()
-		claim.Status.Devices = devices
+		claim.Status.Devices = mergeDeviceStatus(claim.Status.Devices, devices)
 
 		_, err = rc.UpdateStatus(ctx, claim, metav1.UpdateOptions{})
 		return err
 	})
+}
+
+// mergeDeviceStatus merges updates into existing per-device status entries,
+// matched on driver, pool, device and shareID. Only the Data field of a
+// matching entry is replaced, so fields written by other actors (for example
+// Conditions set by the binding conditions controller) are preserved. Entries
+// with no existing match are appended.
+func mergeDeviceStatus(existing, updates []resourceapi.AllocatedDeviceStatus) []resourceapi.AllocatedDeviceStatus {
+	merged := slices.Clone(existing)
+	for _, u := range updates {
+		idx := slices.IndexFunc(merged, func(d resourceapi.AllocatedDeviceStatus) bool {
+			return d.Driver == u.Driver && d.Pool == u.Pool && d.Device == u.Device && ptr.Equal(d.ShareID, u.ShareID)
+		})
+		if idx < 0 {
+			merged = append(merged, u)
+			continue
+		}
+		merged[idx].Data = u.Data
+	}
+	return merged
 }
--- a/cmd/dra-example-kubeletplugin/state_test.go
+++ b/cmd/dra-example-kubeletplugin/state_test.go
@@ -30,6 +30,7 @@ import (
 	resourceapi "k8s.io/api/resource/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	drapbv1 "k8s.io/kubelet/pkg/apis/dra/v1"
 	"k8s.io/utils/ptr"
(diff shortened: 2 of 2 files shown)


nojnhuh (commented):
/lgtm
/approve

PR #268, opened 2026-09-08
GPU profile: Publish standard pcieRoot device attribute

wizhaoredhat:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

The DRA standard device attribute `resource.kubernetes.io/pcieRoot` lets schedulers and users express co-location constraints via `matchAttributes` (for example, allocate a GPU and a network device under the same PCIe root complex).

This PR adds optional, opt-in support for publishing that attribute on mock GPU and net devices in the example driver. Publishing is disabled by default and does not probe host topology; instead values are assigned deterministically in round-robin order from a configurable (or default) root list.

**Configuration**

| Flag / env var | Helm value | Default |
|---|---|---|
| `--gpu-publish-pcie-root` / `GPU_PUBLISH_PCIE_ROOT` | `gpuPublishPCIeRoot` | `false` |
| `--net-publish-pcie-root` / `NET_PUBLISH_PCIE_ROOT` | `netPublishPCIeRoot` | `false` |
| `--pcie-roots` / `PCIE_ROOTS` | `pcieRoots` | `""` |

When publishing is enabled and `PCIE_ROOTS` is empty, the driver uses built-in defaults (`pci0000:00`, `pci0000:80`). When set, roots are assigned in round-robin order (`gpu-0` → first root, `gpu-1` → second root, etc.).

**Motivation**

- Demonstrate cross-profile `matchAttributes` using only the example driver (GPU + net), without depending on other drivers.
- Allow external CI (for example, dra-driver-sriov) to override `PCIE_ROOTS` so mock GPU devices align with real SR-IOV VF topology on a given testbed.

Example: an SR-IOV VF publishes `resource.kubernetes.io/pcieRoot: pci0000:14`. With `gpuPublishPCIeRoot=true` and `pcieRoots="pci0000:14,..."`, a mock GPU can publish the same root so a `ResourceClaim` can match across drivers.

#### Which issue(s) this PR fixes:

N/A

#### Special notes for your reviewer:

- Publishing is opt-in per profile; existing deployments are unchanged.
- Assignment is configurable or defaults using round-robin.
- Shared helper logic lives in `internal/profiles/helpers/pcie_topology.go`.
- Unit tests cover configured roots, default roots, round-robin assignment, and the disabled-by-default behavior.

Tests run:

- `go test ./internal/profiles/helpers/... -count=1`
- `go test ./internal/profiles/gpu/... -count=1`
- `go test ./internal/profiles/net/... -count=1`
- `go test ./...`

#### Does this PR introduce a user-facing change?

```release-note
Added optional support for publishing the standard `resource.kubernetes.io/pcieRoot` device attribute on mock GPU and net devices. Enable via `GPU_PUBLISH_PCIE_ROOT` / `NET_PUBLISH_PCIE_ROOT`; override root values with `PCIE_ROOTS`. Disabled by default.

### DIFF
--- a/internal/profiles/helpers/pcie_topology.go
+++ b/internal/profiles/helpers/pcie_topology.go
@@ -1,0 +1,66 @@
+/*
+ * Copyright The Kubernetes Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package helpers
+
+import (
+	"strings"
+
+	resourceapi "k8s.io/api/resource/v1"
+	"k8s.io/dynamic-resource-allocation/deviceattribute"
+)
+
+// DefaultPCIeRoots are used when pcieRoot publishing is enabled but no roots
+// are configured explicitly.
+var DefaultPCIeRoots = []string{"pci0000:00", "pci0000:80"}
+
+// ParsePCIeRoots splits a comma-separated list of PCIe root values.
+func ParsePCIeRoots(s string) []string {
+	if s == "" {
+		return nil
+	}
+	var roots []string
+	for _, part := range strings.Split(s, ",") {
+		root := strings.TrimSpace(part)
+		if root != "" {
+			roots = append(roots, root)
+		}
+	}
+	return roots
+}
+
+// ResolvePCIeRoots returns configured roots when set, otherwise DefaultPCIeRoots.
+func ResolvePCIeRoots(configured []string) []string {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: internal/profiles/helpers/pcie_topology.go [hunk 1]
```diff
@@ -0,0 +1,106 @@
...
+	return discoverPCIeAnchorsFromRoot(root)
+}
+
+func discoverPCIeAnchorsFromRoot(root string) ([]PCIeAnchor, error) {
+	var (
+		devicesDir string
+		mods       []deviceattribute.MachineModifier
+	)
+	if root == "" {
+		devicesDir = "/sys/bus/pci/devices"
```
nojnhuh: I'm not opposed to publishing the standard attribute, but I'd rather keep it entirely deterministic no matter the underlying host. Adding the same attribute to other relevant device profiles would enable examples showing how the attribute can be used without relying on other drivers.

SchSeba: +1 for examples!
but if possible to have an env variable we can control the pcieroot device for the expose devices that will be great so we can use this example dra driver in our (the dra-driver-sriov) ci system to test contains between the devices

wizhaoredhat: @nojnhuh and @SchSeba I made the changes for determinitic assignment of PCIe Roots done using ENV variables and defaults. PTAL. Thanks!

PR #273, opened 2026-09-15
Clean up CI Go version

nojnhuh:
This PR cleans up how the Go version used to run tests and build release artifacts is specified.

This is stacked on top of #222 so I'll update this after that merges.
/hold


nojnhuh:
> This is stacked on top of #222 so I'll update this after that merges.
> /hold

Rebased on top of that PR.
/hold cancel

/retitle Clean up CI Go version

harche:
/lgtm

Go version selection looks consistent across tests and container builds, and CI passes.

Optional follow-up: derive `GO_VERSION` in `common.mk` from the root `go.mod`, or add a consistency check, to prevent the versions drifting apart.

PR #274, opened 2026-09-16
Fix dependabot config syntax

nojnhuh:
GitHub only surfaced this error _after_ #222 merged 🙃 

https://github.com/kubernetes-sigs/dra-example-driver/runs/104764206280

> The property '#/updates/0/schedule/time' of type integer did not match the following type: string
> The property '#/updates/1/schedule/time' of type integer did not match the following type: string
> The property '#/updates/2/schedule/time' of type integer did not match the following type: string


harche:
Non-blocking follow-up: it would be useful to add Dependabot schema validation to CI so errors like this are caught before merging. The existing lint and test jobs don’t cover this configuration. This can be handled separately.

/lgtm

PR #275, opened 2026-09-16
Add Dependabot schema validation to CI

harche:
Adds CI validation for `.github/dependabot.yml` using `marocchino/validate-dependabot` pinned to v3.0.0.

This catches schema errors before Dependabot runs, including unquoted `schedule.time` values such as `08:00` that GitHub parses as integers.

@nojnhuh Could you review this follow-up to PR #274?


.github/workflows/dependabot.yaml
```diff
@@ -0,0 +1,23 @@
...
+  contents: read
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+    - name: Validate Dependabot configuration
+      uses: marocchino/validate-dependabot@d8ae5c0d03dd75fbd0ad5f8ab4ba8101ebbd4b37 # v3.0.0
```
nojnhuh, line 23: Given the issues that have come up in the past with GitHub Actions being compromised, I want to be particularly careful about adding new third-party dependencies here. Specifically, this tagged release is over 2 years old and the latest commit is already 6 months old. It also doesn't seem widely adopted or have a wide contributor base. Doing due diligence and auditing that code up front will probably take more time that it would to fix these issues as they come up.

I'm willing to accept a little bit of friction in only running this validation after merge if GitHub is defining that itself.

/close

